PR #35, opened 2024-06-04
Rename pinwheel, add modal, add CSP, create new service and connect

allthesignals:
Swaps in Pinwheel's Web SDK. 

Instead of re-using the short-lived user token, Pinwheel's API needs to generate a new one for every chosen employer.

Implements Pinwheel search and link.


app/app/controllers/api/pinwheel_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def provider
+    PinwheelService.new
+  end
+
+  def token_params
+    params.require(:pinwheel).permit(:response_type, :id)
+  end
+end
```
tdooner, line 22: Tests would be swell

app/app/javascript/controllers/cbv_flows_controller.js
```diff
@@ -12,56 +11,49 @@ export default class extends Controller {
...
-      .then(Argyle => initializeArgyle(Argyle, this.argyleUserToken, {
-        items: [itemId],
-        onAccountConnected: this.onSignInSuccess.bind(this),
-        onAccountError: this.onAccountError.bind(this),
-        // Unsure what these are for!
-        onDDSSuccess: () => { console.log('onDDSSuccess') },
-        onDDSError: () => { console.log('onDDSSuccess') },
-        onTokenExpired: updateToken,
+  submit(token) {
+    loadPinwheel()
```
tdooner, line 52: Maybe we should load pinwheel earlier, so it's ready by the time the user submits the search form?

app/app/javascript/utilities/pinwheel.js
```diff
@@ -0,0 +1,38 @@
...
+        resolve(Pinwheel);
+      }
+    });
+  });
+}
+
+export function initializePinwheel(Pinwheel, linkToken, callbacks) {
+  Pinwheel.open({
+    linkToken,
+    sandbox: metaContent('pinwheel_sandbox'), // Set to false for production environment.
```
tdooner, line 22: Pinwheel does not seem to have a `sandbox` option here - https://docs.pinwheelapi.com/public/v2022-09-09/docs/link-1#initializationparams

I assume it probably infers the environment either via the Link URL? Anyway, probably should remove this stuff so it's not confusing.

tdooner (approved):
Looks good; plenty here to follow up on as you keep working on the branch.

PR #43, opened 2024-06-07
Bugfix/ffs 861

GeorgeCodes19:
## Ticket

Resolves [FFS-861](https://jiraent.cms.gov/browse/FFS-861)

## Changes

- Fixes inconsistencies with local Docker development and application instances running in ECS by updating the docker compose configuration. This ensures that the postgres database used for local development matches the postgres database used in ECS.
- Removes the `ruby-debug-ide` gem. (A follow-up PR may be needed to include this gem for local debugging, but removes it during the Docker build phase)
- A widely adopted and supported chromium binary is not available for arm64 based machines. A conditional to check system architecture has been added to skip chromium installation when installing puppeteer
- Adds a `DOCKERIZED` environment variable for Docker containers so that the application can have varying runtime configuration (`wkhtmltopdf` binary location) based on the host architecture.

## Context for reviewers

This PR addresses an issue where `wkhtmltopdf` does not have access to all system-level dependencies and configures `wickedpdf` to use the proper temporary directory. This is resolved by setting the `TMPDIR` environment variable. 

## Testing
Screenshot 2024-06-07 at 4 57 18 PM

### DIFF
--- a/app/bin/dev
+++ b/app/bin/dev
@@ -1,5 +1,7 @@
 #!/usr/bin/env sh
 
+rm /rails/tmp/pids/server.pid
+
 if gem list --no-installed --exact --silent foreman; then
   echo "Installing foreman..."
   gem install foreman
--- a/app/bin/dev
+++ b/app/bin/dev
@@ -5,10 +5,12 @@ if gem list --no-installed --exact --silent foreman; then
   gem install foreman
 fi
 
-if ! which ngrok >/dev/null; then
-  echo "Install ngrok to continue!"
-  echo "  brew install ngrok"
-  exit 1
+if [ "$DOCKERIZED" != "true" ]; then
+  if ! which ngrok >/dev/null; then
+    echo "Install ngrok to continue!"
+    echo "  brew install ngrok"
+    exit 1
+  fi
 
   if ! ngrok config check; then
     echo "Set up ngrok to continue!"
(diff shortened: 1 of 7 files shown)


app/bin/dev
```diff
@@ -1,14 +1,18 @@
 #!/usr/bin/env sh
 
+rm /rails/tmp/pids/server.pid
```
tdooner, line 3: This seems wrong, especially outside of Docker - the point of the pidfile is to prevent multiple servers from simultaneously running, so it shouldn't be removed without first checking that the server has in fact crashed.

```suggestion
PIDFILE_PATH=$(cd $(dirname $0)/../tmp/pids; pwd)/server.pid
if [ -f $PIDFILE_PATH ] && ! ps -p "$(cat $PIDFILE_PATH)" >/dev/null; then
  echo "Found old pidfile at $PIDFILE_PATH... removing"
  rm $PIDFILE_PATH
fi
```

app/config/initializers/wicked_pdf.rb
```diff
@@ -11,6 +11,12 @@
 WickedPdf.configure do |config|
   # Path to the wkhtmltopdf executable: This usually isn't needed if using
   # one of the wkhtmltopdf-binary family of gems.
+
+  # set the wkhtmltopdf path for docker
+  if ENV["DOCKERIZED"] == "true"
+    config.exe_path = "/rails/bin/wkhtmltopdf"
+  end
```
tdooner, line 18: I don't think this should be necessary, if we don't do the moving-it-to-the-/rails/bin-directory thing, right? I'd expect `apt` to install it into somewhere on the $PATH automatically.

tdooner (approved):
Looks good with feedback. Since I've got a few more hours in the day I'll go ahead and merge and fix these things up so I can build on top of it.

PR #53, opened 2024-06-10
822 webhook

allthesignals:
## Ticket

Resolves https://jiraent.cms.gov/browse/FFS-822

## Changes
 - Adds the webhook verification logic
 - Listens for the correct Pinwheel events
 - Attempts to search for CbvFlows by given account ID.

### DIFF
--- a/app/app/javascript/controllers/cbv_flows_controller.js
+++ b/app/app/javascript/controllers/cbv_flows_controller.js
@@ -16,7 +16,6 @@ export default class extends Controller {
   cable = ActionCable.createConsumer();
 
   connect() {
-    // TODO: Set up ActionCable to listen for paystub sync events
     this.cable.subscriptions.create({ channel: 'PaystubsChannel' }, {
       connected: () => {
         console.log("Connected to the channel:", this);


app/app/controllers/webhooks/pinwheel/events_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+        PinwheelPaystubsChannel.broadcast_to(@cbv_flow, params)
+      end
+    end
+
+    if params["event"] == "accounts.added"
+      rep = ConnectedArgyleAccount.create!(
+        user_id: params["payload"]["end_user_id"],
+        account_id: params["payload"]["link_token_id"]
+      )
+      Rails.logger.info "ConnectedArgyleAccount created: #{rep}"
```
GeorgeCodes19, line 28: This would log incorrect data since we're not connecting an Argyle account.

GeorgeCodes19 (changes requested):
Overall this looks good. I do see a few concerning line items that could lead to false positives or incorrect information being logged.

app/spec/support/pinwheel_api_helper.rb
```diff
@@ -51,11 +51,14 @@ def stub_refresh_user_token_response
...
       File.dirname(__FILE__),
-      "fixtures/argyle/#{filename}"
-    )
-    JSON.parse(File.read(path))
+      "fixtures/pinwheel/#{filename}"
+    ))
+  end
+
+  def load_relative_json_file(filename)
+    JSON.parse(load_relative_file(filename))
```
tdooner, line 62: Is this used anywhere?

app/app/services/pinwheel_webhook_manager.rb
```diff
@@ -0,0 +1,31 @@
...
+    response = @pinwheel.create_webhook_subscription([
+      "account.added",
+      "paystubs.added"
+    ], URI.join(tunnel_url, "/webhooks/pinwheel/events", format_identifier_hash(name)))
+    new_webhook_subscription_id = response["data"]["id"]
+    puts "  ✅ Set up Pinwheel webhook: #{new_webhook_subscription_id}"
+  end
+
+  def format_identifier_hash(identifier)
+    "#subscription_name=#{identifier}"
```
tdooner, line 29: Oh neat. I'm kind of surprised this actually works. Very neat.

app/spec/services/pinwheel_service_spec.rb
```diff
@@ -27,6 +27,33 @@
...
+      '860860860'
+    }
+
+    let(:signature_digest) {
+      'v2=42fb9eba200e821d4de63667f5a30f7e1b83609b135e148e26ce01eef2aa6ba8'
+    }
+
+    let(:key) {
+      'TEST_KEY'
+    }
```
tdooner, line 45: nit: Generally use `do`/`end` for multiline blocks

app/app/services/pinwheel_webhook_manager.rb
```diff
@@ -0,0 +1,31 @@
+# This class manages pinwheel webhook subscriptions, and is intended for development environment setup only
+class PinwheelWebhookManager
+  def initialize
+    @pinwheel = PinwheelService.new
+  end
+
+  ## may need to track the id
```
tdooner, line 7: ^ stale comment

app/app/services/pinwheel_service.rb
```diff
@@ -48,4 +49,36 @@ def create_link_token(end_user_id:, response_type:, id:)
...
+  def create_webhook_subscription(events, url)
+    @http.post(build_url(WEBHOOKS_ENDPOINT), {
+      enabled_events: events,
+      url: url,
+      status: "active",
+      version: PINWHEEL_VERSION
+    }.to_json).body
+  end
+
+  def generate_signature_digest(timestamp, raw_body, api_key = ENV["PINWHEEL_API_TOKEN"])
```
tdooner, line 71: We probably should store `api_key` as an instance variable so we don't need to pass it in again.

app/app/controllers/webhooks/pinwheel/events_controller.rb
```diff
@@ -0,0 +1,36 @@
+class Webhooks::Pinwheel::EventsController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  def create
```
GeorgeCodes19, line 4: In general this method seems to be doing a lot. It violates the single responsibility principal. It's not only handling the creation of two different webhooks, but first handles validating a request signature. This means in order to test that the webhooks are processed correctly we need mock a valid request signature.

I would suggest that the `create` method handle delegating the processing of `paystubs.added` and `accounts.added`. 

```suggestion
def create
  switch params["event"]
  when "paystubs.added"
    handle_paystub_added(params)
  when "accounts.added"
    handle_accounts_added(params)
  end
end
```

Additionally, we can move the logic to validate a request to its own method.

app/app/javascript/controllers/cbv_flows_controller.js
```diff
@@ -17,19 +17,19 @@ export default class extends Controller {
 
   connect() {
     // TODO: Set up ActionCable to listen for paystub sync events
```
tdooner, line 19: Stale comment?

tdooner (approved):

GeorgeCodes19 (changes requested):
I would suggest that the create method handle delegating the processing of paystubs.added and accounts.added to other methods in the `Webhooks::Pinwheel::EventsController`.

PR #54, opened 2024-06-12
Fix AWS ECS healthcheck by using `curl` instead of `wget`

tdooner:
We don't actually install wget inside the containers, so this
healthcheck was always failing and has been provisioning new containers
all the time. This may have been the source of some of our deployment
downtime.

It works:
image


tdooner:
CPU utilization is way down after deploying this, thankfully:
image

GeorgeCodes19 (approved):

PR #59, opened 2024-06-13
Query for accounts/then payroll

allthesignals:
## Ticket

[Resolves ](https://jiraent.cms.gov/browse/FFS-883)

## Changes

 - Uses the CbvFlow to track end_user_id within the Pinwheel API context
 - Supplies that end_user_id upon Link token creation
 - Queries for accounts associated with that end_user_id
 - Uses those accounts to query for associated paystubs

### DIFF
--- a/app/app/controllers/cbv_flows_controller.rb
+++ b/app/app/controllers/cbv_flows_controller.rb
@@ -17,12 +17,12 @@ class CbvFlowsController < ApplicationController
 
     @payments = fetch_payroll.map do |payment|
       {
-        employer: payment["employer"],
-        amount: payment["net_pay"].to_i,
-        start: payment["paystub_period"]["start_date"],
-        end: payment["paystub_period"]["end_date"],
-        hours: payment["hours"],
-        rate: payment["rate"]
+        employer: payment["employer_name"],
+        amount: payment["net_pay_amount"].to_i,
+        start: payment["pay_period_start"],
+        end: payment["pay_period_end"],
+        hours: payment["earnings"][0]["hours"],
+        rate: payment["earnings"][0]["rate"]
       }
     end
 
@@ -91,7 +91,11 @@ class CbvFlowsController < ApplicationController
   end
 
   def fetch_payroll
-    provider.fetch_paystubs(user: @cbv_flow.argyle_user_id)["results"]
+    account_ids = provider.fetch_accounts(end_user_id: @cbv_flow.id)["data"].map { |account| account["id"]}
+
+    account_ids.map do |account_id|
+      provider.fetch_paystubs(account_id: account_id)["data"]
+    end.flatten
   end
 
   def provider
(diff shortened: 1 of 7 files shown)


app/app/services/pinwheel_service.rb
```diff
@@ -29,6 +29,11 @@ def initialize(api_key = ENV["PINWHEEL_API_TOKEN"])
     @http = Faraday.new(client_options) do |conn|
       conn.response :raise_error
       conn.response :json, content_type: "application/json"
+      conn.response :logger,
+        Rails.logger,
+        headers: true,
+        bodies: true,
+        log_level: :debug
```
tdooner, line 36: Make sure to not commit this for reals

app/app/services/pinwheel_service.rb
```diff
@@ -63,16 +76,16 @@ def create_webhook_subscription(events, url)
...
     }.to_json).body
   end
 
   def generate_signature_digest(timestamp, raw_body, api_key = ENV["PINWHEEL_API_TOKEN"])
     msg = "v2:#{timestamp}:#{raw_body}"
     digest = OpenSSL::HMAC.hexdigest(
-      OpenSSL::Digest.new("sha256"),
-      api_key.encode("utf-8"),
+      OpenSSL::Digest.new('sha256'),
+      api_key.encode('utf-8'),
```
tdooner, line 88: Your rubocop is confused, these should be double-quotes per our current style rules

app/app/controllers/cbv_flows_controller.rb
```diff
@@ -89,7 +89,11 @@ def fetch_employers(query = "")
   end
 
   def fetch_payroll
-    provider.fetch_paystubs(user: @cbv_flow.argyle_user_id)["results"]
+    account_ids = provider.fetch_accounts(end_user_id: @cbv_flow.id)["data"].map { |account| account["id"]}
```
tdooner, line 92: The more I think about it the more I do think we should use a UUID for the `end_user_id`, since all of our machines' environments will have integers that are colliding here, and we'll see weird behavior. We can do that later. Also - the ID/primary key does not need to be a UUID... we could add a separate column called something like `pinwheel_end_user_id` which is a UUID and then use that value.

app/db/migrate/20240613154132_add_pgcrypto_extension.rb
```diff
@@ -0,0 +1,5 @@
+class AddPgcryptoExtension < ActiveRecord::Migration[7.1]
+  def change
+    enable_extension 'pgcrypto'
```
tdooner, line 3: This shouldn't be necessary if we're only using `gen_random_uuid()` on Postgres 13+. I'm going to remove it and see if it still works.

app/app/services/pinwheel_service.rb
tdooner: Ok - if it's not too noisy, let's just keep it then.

PR #60, opened 2024-06-13
George/ffs 777

GeorgeCodes19:
## Ticket

Resolves [FFS-777](https://jiraent.cms.gov/browse/FFS-777)

## Changes

Adds the ability for applicants to send a copy of their verified income data to caseworkers.

## Testing

- [x] - Test to verify email subject is accurate
- [x] - Test to verify that email body is accurate
- [x] - Test to verify that email contains a pdf attachment
- [x] - Controller test for the `/share` endpoint
- [x] - Add missing env vars for  [ApplicantMailer](https://github.com/DSACMS/iv-cbv-payroll/pull/60/files/8c52869c92ea4ebd6cfcb1b378350200dd842664#diff-e1806ac57a45f45c67fb5205e36e44ab5f4c5aa8e672d0529a5812a24047724c) tests

Screenshot 2024-06-14 at 9 40 59 PM


app/app/controllers/cbv_flows_controller.rb
```diff
@@ -36,7 +36,13 @@ def summary
   end
 
   def share
+    if params[:commit] == I18n.t("cbv_flows.share.share_with_caseworker")
+      # replace with actual caseworker_email e.g. to_email = params[:cbv_flow][:caseworker_email]
+      to_email = ENV["SLACK_TEST_EMAIL"]
+    pdf_path = "#{Rails.root}/tmp/#{params[:id]}.pdf"
+    ApplicantMailer.send_pdf_to_applicant(to_email, pdf_path).deliver_now
   end
+end
```
tdooner, line 45: This indenting is messed up

app/app/controllers/cbv_flows_controller.rb
```diff
@@ -36,7 +36,13 @@ def summary
   end
 
   def share
+    if params[:commit] == I18n.t("cbv_flows.share.share_with_caseworker")
+      # replace with actual caseworker_email e.g. to_email = params[:cbv_flow][:caseworker_email]
+      to_email = ENV["SLACK_TEST_EMAIL"]
+    pdf_path = "#{Rails.root}/tmp/#{params[:id]}.pdf"
+    ApplicantMailer.send_pdf_to_applicant(to_email, pdf_path).deliver_now
```
tdooner, line 43: If we are going to use a temporary file, let's use Ruby's [Tempfile#create](https://ruby-doc.org/stdlib-3.1.1/libdoc/tempfile/rdoc/Tempfile.html#method-c-create) so that this file gets cleaned up after we're done sending.

```suggestion
    Tempfile.create("cbv-summary-#{params[:id]}", Rails.root.join("tmp")) do |f|
      # <placeholder for rendering the PDF into f.path>
      ApplicantMailer.send_pdf_to_applicant(to_email, f.path).deliver_now
    end
```

But actually after writing this comment, I found the documentation for WickedPdf that shows a one-liner for attaching a rendered PDF to an email: https://github.com/mileszs/wicked_pdf?tab=readme-ov-file#include-in-an-email-as-an-attachment
^ Check that out. Maybe we actually should just do that in the ApplicantMailer and not even worry about passing temporary files around. The PDFs would be stored entirely in memory, which could be inefficient for big PDFs, but I think these ones should be pretty small.

app/config/routes.rb
```diff
@@ -16,10 +16,13 @@
...
       get "/employer_search" => "cbv_flows#employer_search"
       get "/argyle_link" => "cbv_flows#argyle_link"
+      # summary
       patch "/summary" => "cbv_flows#summary"
       get "/summary" => "cbv_flows#summary"
-      patch "/share" => "cbv_flows#share"
+      # share
       get "/share" => "cbv_flows#share"
+      post "/share" => "cbv_flows#share"
+      patch "/share" => "cbv_flows#share"
```
tdooner, line 25: Only one of these two should be necessary. I don't think we need the `patch` one.

app/app/mailers/applicant_mailer.rb
```diff
@@ -1,8 +1,23 @@
 class ApplicantMailer < ApplicationMailer
-  attr_reader :email_address
+  before_action { @cbv_flow = params[:cbv_flow] }
+  attr_reader :email_address, :cbv_flow
```
GeorgeCodes19, line 3: Might revert this. I figured at some point we'd want a getter for the `cbv_flow`, but this may be introducing features not yet utilized/needed.

app/app/mailers/applicant_mailer.rb
```diff
@@ -1,8 +1,23 @@
...
+  helper :cbv_flows
```
GeorgeCodes19, line 4: Maybe we can just use the `ApplicationHelper` here? I want to avoid tight coupling though. This does introduce duplicate code since our view in particular is interested in the `format_date` method which now exists in two places (`application_helper` and `cbv_flows_helper`).

app/spec/mailers/applicant_mailer_spec.rb
```diff
@@ -1,6 +1,19 @@
 require "rails_helper"
 
 RSpec.describe ApplicantMailer, type: :mailer do
+  let(:payments) do
```
GeorgeCodes19, line 4: I'd prefer that this be a stub that can be recalled whenever we need it. It's duplicated in the `ApplicantMailer::Preview` class. Where is an ideal place for this in Rails?

app/app/mailers/applicant_mailer.rb
tdooner: Ultimately these Helper modules are all arbitrary anyway so I wouldn't lose sleep over whether this is in `CbvFlowsHelper` or any other. I would say in general it's good to align the names of the Helper with the controller/mailer that is using them. I.e. `ApplicantMailerHelper`.

tdooner (commented):
Hopping into a meeting but here's some feedbakc

app/app/helpers/cbv_flows_helper.rb
```diff
@@ -0,0 +1,10 @@
+module CbvFlowsHelper
+  def format_date(timestamp_string)
+    begin
+      Time.parse(timestamp_string).strftime("%B %d, %Y")
+    rescue => e
+      "Invalid timestamp"
+      timestamp_string
```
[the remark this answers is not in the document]
GeorgeCodes19: Gotcha! This was [copied from the application_helper.rb file](https://github.com/DSACMS/iv-cbv-payroll/blob/george/ffs-777/app/app/helpers/application_helper.rb#L10). It's duplicate code and I wasn't sure how to make a general _helper_. I was attempting to follow the [Rails helper conventions ](https://api.rubyonrails.org/classes/ActionController/Helpers.html)which do some type of autoloading to resolve the helpers based on controller name. 

> In previous versions of Rails the controller will include a helper which matches the name of the controller, e.g., MyController will automatically include MyHelper.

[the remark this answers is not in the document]
GeorgeCodes19: Figured it out

app/app/mailers/applicant_mailer.rb
GeorgeCodes19: I had abandoned that pattern before reading your comment. There's really only two helper methods that get used by the views so I've created a single `ViewHelper` class. I figured we could break them up down the road if the project called for it.

tdooner (approved):
No additional comments beyond what Matt left!

app/app/controllers/cbv_flows_controller.rb
```diff
@@ -35,6 +25,15 @@ def summary
   end
 
   def share
+    if Rails.env.development?
+      email_address = ENV["SLACK_TEST_EMAIL"]
+    elsif Rails.env.test?
+      email_address = "test@example.com"
+    else
+      email_address = @cbv_flow.email
```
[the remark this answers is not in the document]
GeorgeCodes19: Example please?

PR #62, opened 2024-06-13
George/ffs 857

GeorgeCodes19:
## Ticket

Resolves [FFS-857](https://jiraent.cms.gov/browse/FFS-857)

## Changes

> Updated permissions for Github Actions to verify SES related configuration.

## Context for reviewers

> This PR will enable Github Actions to automatically deploy merges into the main branch to the AWS demo environment.
## Testing

Screenshot 2024-06-13 at 5 30 15 PM


infra/modules/auth-github-actions/main.tf
```diff
@@ -29,6 +29,17 @@ data "aws_iam_policy_document" "github_actions" {
...
     resources = ["*"]
   }
+   statement {
+    sid       = "SESPermissions"
+    effect    = "Allow"
+    actions   = [
+      "ses:GetIdentityVerificationAttributes",
+      "ses:VerifyDomainIdentity",
+      "ses:GetIdentityMailFromDomainAttributes",
+      "ses:GetIdentityDkimAttributes"
```
tdooner, line 39: It looks like these will be passed in automatically if you just add "ses" to the giant array in `infra/project-config/aws-services.tf`. (Well, actually, `ses:*` will be passed in, which is more permissions than we need, but that seems okay.)

infra/project-config/aws-services.tf
```diff
@@ -28,6 +28,7 @@ locals {
     "secretsmanager",
     "servicediscovery",
     "sns",
+    "ses",
```
tdooner, line 31: It looks like these are in alphabetical order... so.... uh.... could you move it up a line? :)

tdooner (approved):

infra/project-config/aws-services.tf
tdooner: (This may be because I am an order muppet)

PR #66, opened 2024-06-18
Added case number and timestamps to email and generated summary PDFs

GeorgeCodes19:
Resolves [FFS-776](https://jiraent.cms.gov/browse/FFS-776)

## Changes

This PR adds the case number and timestamp (Document origination date) to the generated PDF.

- Added timestamp to generated PDF
- Added case number to generated PDF
- Added `format_view_datetime` method to `ViewHelper`

## Testing

> The `applicant_mailer_spec` has been updated so that it asserts the presence of the case number and timestamp.

Screenshot 2024-06-18 at 6 52 31 PM
Screenshot 2024-06-18 at 5 07 20 PM


app/app/helpers/view_helper.rb
```diff
@@ -15,4 +15,14 @@ def format_date(timestamp_string)
       timestamp_string
     end
   end
+
+  def format_view_datetime(timestamp_string)
+    begin
+      formatted_time = Time.parse(timestamp_string).strftime("%B %d, %Y")
+      raw_timestamp = Time.parse(timestamp_string).strftime("%I:%M %p %Z")
+      "#{formatted_time} - #{raw_timestamp}"
```
tdooner, line 23: Why not just put it all in one `strftime` call?

app/spec/mailers/previews/applicant_mailer_preview.rb
```diff
@@ -1,13 +1,16 @@
 # Preview all emails at http://localhost:3000/rails/mailers/applicant_mailer
+require Rails.root.join('spec/support/test_helpers')
 class ApplicantMailerPreview < ActionMailer::Preview
+  include ViewHelper
+  include TestHelpers # in order to use stub_payments
```
tdooner, line 5: huh... how was it working before then?

app/app/views/cbv_flows/summary.pdf.erb
```diff
@@ -1,3 +1,8 @@
+<p>
+  <%= t('cbv_flows.summary.case_number', case_number: @cbv_flow.case_number) %>
+</p>
+<p>
+  <%= t('cbv_flows.summary.document_timestamp', timestamp: format_view_datetime(Time.now.to_s)) %></p>
```
tdooner, line 5: Why pass `Time.now.to_s` to this function as a string instead of the actual Time object? It's stringifying and then parsing it for no reason.

tdooner (approved):

app/app/helpers/view_helper.rb
tdooner: (e.g.

```ruby
Time.parse(timestamp_string).strftime("%B %d, %Y - %I:%M %p %Z")
```
)

app/app/views/cbv_flows/summary.pdf.erb
GeorgeCodes19: This same pattern can be found in the [summary.html.erb](https://github.com/DSACMS/iv-cbv-payroll/blob/george/ffs-777/app/app/views/cbv_flows/summary.html.erb#L25). I didn't want to create a new pattern whereas an object gets passed. So I opted to stick with what's already in the codebase. Two different approaches means that somewhere down the line- someone needs to make a decision to refactor them for uniformity. Not worth the mental energy IMO.

app/spec/mailers/previews/applicant_mailer_preview.rb
GeorgeCodes19: It wasn't working before because the underlying pdf summary view didn't call for the use of a helper method. After too much Googling and experimentation- this is the (a?) correct approach. 

`ActionMailer::Preview` instances do not automatically inherit helper methods the same way that controllers do. Since `stub_payments` is technically a helper method used in testing and the ActionMailer::Preview is used for testing and local development- it seemed okay to use a test helper here. 

TLDR; helpers should be explicitly imported in classes extending ActionMailer::Preview if you wish for those helper methods to be available in views used by the `ActionMailer::Preview`. Otherwise you'll see an error such as the one below if you do not include the helper module.

Screenshot 2024-06-20 at 11 49 48 AM

PR #76, opened 2024-06-28
Enabling new relic agent, adding newrelic_key env var to terraform co…

GeorgeCodes19:
Enabling the New Relic agent and adding the required environment variable to AWS System Manger in order to view metrics from the CBV demo environment.

- Enables browser monitoring (geographical region, browser agent, AJAX requests, page navigations, time to paint, etc.

## Ticket

Initial step to resolve [FFS-940](https://jiraent.cms.gov/browse/FFS-940)

## Context for reviewers

> This PR doesn't contain the full feature to close out FFS-940, but rather enables data to flow into New Relic so that we can customize the dashboard and setup alerts in accordance to the data being observed.

## Testing
Screenshot 2024-06-28 at 3 19 15 PM

### DIFF
--- a/app/config/newrelic.yml
+++ b/app/config/newrelic.yml
@@ -26,14 +26,14 @@ common: &default_settings
 
   browser_monitoring:
     # include js code via partial to comply with CSP settings
-    auto_instrument: false
+    auto_instrument: true
 
   # This line disables agent regardless of other settings.
   # To enable the New Relic agent:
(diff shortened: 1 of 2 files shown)


tdooner (approved):
I wonder if we'll have to change the collector host to be something other than the gov-collector.newrelic.com?

GeorgeCodes19:
> I wonder if we'll have to change the collector host to be something other than the gov-collector.newrelic.com?

Good point. We should make it dynamic in a following PR. Maybe include the env?

PR #78, opened 2024-06-28
FFS-986: Update copy on search page

JosephGasiorekUSDS:
## Ticket

Resolves [FFS-986](https://jiraent.cms.gov/browse/FFS-986)

## Changes

Updated copy on the `employer_search` screen

## Context for reviewers

I did not add the missing results button. That seems to be more part of [FFS-984](https://jiraent.cms.gov/browse/FFS-984).

### Screenshots
Screen Shot 2024-06-28 at 13 57 24-fullpage
Screen Shot 2024-06-28 at 13 57 30-fullpage

PR #79, opened 2024-06-28
FFS-984: Missing Results Page

JosephGasiorekUSDS:
## Ticket

Resolves [FFS-984](https://jiraent.cms.gov/browse/FFS-984)

## Changes

1. Adds a screen to say what to do when the user can not find their employer on the search page
2. Links to the screen from the search results page

## Context for reviewers

NOTE: This should be reviewed after #78.

There is a link to the "agency site". I used https://www.cms.gov but open to suggestions.

### Screenshots
Screen Shot 2024-06-28 at 14 34 59-fullpage
Screen Shot 2024-06-28 at 14 35 04-fullpage
Screen Shot 2024-06-28 at 14 35 10-fullpage
Screen Shot 2024-06-28 at 14 35 14-fullpage

### DIFF
--- a/app/app/views/cbv/employer_searches/_employer.html.erb
+++ b/app/app/views/cbv/employer_searches/_employer.html.erb
@@ -39,14 +39,15 @@
 
   <% if @query.present? %>
     <h3><%= t('cbv.employer_searches.show.employer_not_listed') %></h3>
-    <button
-      data-action="click->cbv-employer-search#select"
-      data-id=""
-      data-response-type=""
-      class="usa-button usa-button--outline"
-      type="button"
-      >
-        <%= t('cbv.employer_searches.show.search_by_payroll_provider') %>
-    </button>
+    <%= link_to 'missing_results', data: { turbo_frame: "_top" } do %>
+      <button
+        data-id=""
+        data-response-type=""
+        class="usa-button usa-button--outline"
+        type="button"
+        >
+          <%= t('cbv.employer_searches.show.can_not_find_employer') %>
+      </button>
+    <% end %>
   <% end %>
 <% end %>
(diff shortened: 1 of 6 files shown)


app/app/views/cbv/employer_searches/_employer.html.erb
```diff
@@ -39,14 +39,15 @@
...
-    <button
-      data-action="click->cbv-employer-search#select"
-      data-id=""
-      data-response-type=""
-      class="usa-button usa-button--outline"
-      type="button"
-      >
-        <%= t('cbv.employer_searches.show.search_by_payroll_provider') %>
-    </button>
+    <%= link_to 'missing_results', data: { turbo_frame: "_top" } do %>
```
tdooner, line 42: Does `link_to 'missing_results'` work? Never seen that syntax before (but Rails contains multitudes, so I legit don't know).

I would have expected:

`link_to cbv_flow_missing_result_path, [...]`

app/app/views/cbv/employer_searches/_employer.html.erb
```diff
@@ -39,14 +39,15 @@
...
+      <button
+        data-id=""
+        data-response-type=""
+        class="usa-button usa-button--outline"
```
tdooner, line 46: I think we should just apply these classes to the `<a>` tag rather than having a button in a link.

e.g. `link_to [...], class: "usa-button usa-button--outline"`

app/app/views/cbv/employer_searches/_employer.html.erb
JosephGasiorekUSDS: Thanks. Updated the PR to use this method

app/app/views/cbv/employer_searches/_employer.html.erb
JosephGasiorekUSDS: Switch it to be an anchor link

PR #85, opened 2024-06-28
Update task_executor IAM policy before running migrations

tdooner:
When adding an environment variable to AWS SSM, the task executor will
need access to read the environment variable in order to start up a
container. This is normally handled later in the deploy, when the
`bin/deploy-release.sh` script runs `make infra-update-app-service`. But
by then, the migrations container will already have failed.

This commit updates the targets for the pre-migration `terraform` run to
also update the `aws_iam_role_policy.task_executor` resource so that it
will have the permissions it needs to start up the migrations.

### DIFF
--- a/bin/run-database-migrations.sh
+++ b/bin/run-database-migrations.sh
@@ -45,7 +45,10 @@ MIGRATOR_ROLE_ARN=$(terraform -chdir="infra/$APP_NAME/service" output -raw migra
 echo
 echo "::group::Step 1. Update task definition without updating service"
 
-TF_CLI_ARGS_apply="-input=false -auto-approve -target=module.service.aws_ecs_task_definition.app -var=image_tag=$IMAGE_TAG" make infra-update-app-service APP_NAME="$APP_NAME" ENVIRONMENT="$ENVIRONMENT"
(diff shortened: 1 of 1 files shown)


bin/run-database-migrations.sh
```diff
@@ -45,7 +45,10 @@ MIGRATOR_ROLE_ARN=$(terraform -chdir="infra/$APP_NAME/service" output -raw migra
 echo
 echo "::group::Step 1. Update task definition without updating service"
 
-TF_CLI_ARGS_apply="-input=false -auto-approve -target=module.service.aws_ecs_task_definition.app -var=image_tag=$IMAGE_TAG" make infra-update-app-service APP_NAME="$APP_NAME" ENVIRONMENT="$ENVIRONMENT"
+TF_CLI_ARGS_apply="-input=false -auto-approve -var=image_tag=$IMAGE_TAG " \
+  "-target=module.service.aws_ecs_task_definition.app " \
+  "-target=module.service.aws_iam_role_policy.task_executor" \
```
tdooner, line 50: ^ adding this `-target` is the only actual diff other than breaking the command across multiple lines

GeorgeCodes19 (approved):

PR #89, opened 2024-07-01
Use NR Event Tracker

allthesignals:
## Ticket

Addresses [FFS-705](https://jiraent.cms.gov/browse/FFS-705)

## Changes

Makes use of the NewRelic event tracker module to record [one of the metrics listed](https://confluenceent.cms.gov/display/SFIV/Dashboards+and+Metrics+for+CBV+MVP).

### DIFF
--- a/app/app/services/cbv_invitation_service.rb
+++ b/app/app/services/cbv_invitation_service.rb
@@ -5,7 +5,6 @@ class CbvInvitationService
     send_invitation_email(email_address, invitation.to_url)
     NewRelicEventTracker.track("ApplicantInvitedToFlow", {
       timestamp: Time.now.to_i,
-      case_number: case_number,
       invitation_id: invitation.id
     })
   end


app/app/controllers/cbv/base_controller.rb
```diff
@@ -1,3 +1,5 @@
+require Rails.root.join('lib', 'new_relic_event_tracker.rb')
```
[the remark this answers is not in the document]
GeorgeCodes19: I think modules are autoloaded by default in Rails right? The shorthand way to do this:

```suggestion
require 'new_relic_event_tracker'
```

app/app/controllers/cbv/base_controller.rb
```diff
@@ -11,6 +13,7 @@ def set_cbv_flow
       end
 
       @cbv_flow = invitation.cbv_flow || CbvFlow.create_from_invitation(invitation)
+      NewRelicEventTracker.track('ClickedCBVInvitationLink', { timestamp: Time.now.to_i })
```
GeorgeCodes19, line 16: This is a good start. I'd imagine that we'd want to collect things such as the `session_id` and `invitation_id` as well. 

```suggestion
NewRelicEventTracker.track('ClickedCBVInvitationLink', { timestamp: Time.now.to_i, session_id: session_id, invitation_id: invitation.id })
```
Something like that.

app/app/controllers/cbv/shares_controller.rb
```diff
@@ -11,7 +11,8 @@ def update
       cbv_flow: @cbv_flow,
       payments: @payments
     ).caseworker_summary_email.deliver_now
-
+    
+    NewRelicEventTracker.track('IncomeSummarySharedWithCaseworker', { timestamp: Time.now.to_i })
```
GeorgeCodes19, line 15: Same as previous comment. 

It might make sense to create a decorator method that already collects the metadata pertaining to all events i.e. `session_id` and then pass in the custom event name and event specific metrics.

GeorgeCodes19 (approved):

tdooner (approved):

PR #94, opened 2024-07-07
Fix OWASP test

JosephGasiorekUSDS:
## Ticket

Resolves [FFS-1083](https://jiraent.cms.gov/browse/FFS-1083)

## Changes

* Bring OWASP tests back to working.

## Context for reviewers

The [OWASP test have not been running for some time](https://github.com/DSACMS/iv-cbv-payroll/actions/workflows/owasp-daily-scan.yml).

## Testing

### DIFF
--- a/.github/actions/run-server/action.yml
+++ b/.github/actions/run-server/action.yml
@@ -18,6 +18,7 @@ runs:
         DATABASE_URL: ${{ inputs.database_url }}
         SECRET_KEY_BASE: not-actually-secret
       run: bundle exec rails server &
+      working-directory: app
 
     - name: "Wait for startup"
       shell: bash
--- a/.github/workflows/owasp-daily-scan.yml
+++ b/.github/workflows/owasp-daily-scan.yml
@@ -46,5 +46,5 @@ jobs:
           target: 'http://localhost:3000/'
           fail_action: true
           rules_file_name: 'app/zap.conf'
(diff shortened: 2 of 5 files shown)


joeyg:
@tdooner Are we able to put Sandbox credentials for pinwheel up on github? This fails because it try's to make requests to pinwheel

tdooner:
@joeyg We could, I guess, as a secret environment variable. Is OWASP going through every page in the app or something? I'm just trying to think about what the right thing to do here is architecturally - do we want to somehow stub out Pinwheel in test environments so we're not actually making 3rd party calls. Certainly the Pinwheel webhooks will not arrive properly without some legwork. Can we just have it exclude the "Start Flow Manually" link on the homepage if it's just crawling for links?

joeyg:
@tdooner you can block URLs but there is a bug preventing that from working https://github.com/zaproxy/action-full-scan/issues/83

tdooner:
@joeyg Hm, that sucks. Can we pass in a `-configfile` with these fields rather than having to pass in these escaped regexes on the command line?

tdooner:
@joeyg What is your capacity going to be to work on this? Do you want to keep taking this PR on or should we take it back?

joeyg:
@tdooner thanks for asking - you all should take it on if you need it done soon. My availability changes a lot from day to day.

tdooner:
Closing in favor of #116

PR #104, opened 2024-07-12
FFS-1085: Show employment information

allthesignals:
## Ticket

Resolves [1085](https://jiraent.cms.gov/browse/FFS-1085). Depends on https://github.com/DSACMS/iv-cbv-payroll/pull/101

## Changes

Add employment information above the payment details.

## Context for reviewers

Adds 2 new queries to Pinwheel.

Note that we will need to add layout-wide "gutters" later, there's at ticket for that. So it won't look exactly like the design.

image

Screenshot 2024-07-12 at 5 35 06 PM

### DIFF
--- a/app/app/controllers/cbv/payment_details_controller.rb
+++ b/app/app/controllers/cbv/payment_details_controller.rb
@@ -31,11 +31,11 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
   end
 
   def employment_status
-    @employment["status"]
+    @employment["status"].humanize
   end
 
   def pay_period_frequency
-    @income_metadata["compensation_unit"]
+    @income_metadata["compensation_unit"].humanize
   end
 
   def compensation_amount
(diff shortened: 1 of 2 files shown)


app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -1,21 +1,45 @@
...
+    :gross_pay,
+    :employment_start_date,
+    :employment_end_date,
+    :employment_status,
+    :pay_period_frequency,
+    :compensation_amount
 
   def show
     account_id = params[:user][:account_id]
+    @employment = pinwheel.fetch_employment(account_id: account_id)["data"]
```
JosephGasiorekUSDS, line 14: Could we use Rails threads to tun the three API calls to pinwheel simultaneously?

JosephGasiorekUSDS (commented):
I think you will need to mock the API calls for the tests.

app/app/controllers/cbv/payment_details_controller.rb
JosephGasiorekUSDS: I don't think the Threads example is necessary today but certainly something to keep in mind. There api requests to render one screen seems like a lot.

GeorgeCodes19 (approved):

PR #105, opened 2024-07-15
1092: Summary View Updates

allthesignals:
## Ticket

Addresses, but does not fully resolve [1092](https://jiraent.cms.gov/browse/FFS-1092), because we still need a way to deal with multiple additional comments, which his captured in a previous ticket.

## Changes

Updates the summaries view page, but not the pdf. Includes some helpers that group and reduce payments information.

## Context for reviewers

This does not affect or change the PDF, but that will be updated in a future ticket.

## Testing

image
image

### DIFF
--- a/app/app/views/cbv/summaries/show.html.erb
+++ b/app/app/views/cbv/summaries/show.html.erb
@@ -3,52 +3,53 @@
 </h2>
 
 <% if @payments.any? %>
-  <p>
-    <%= t(".description") %>
-  </p>
-  <h3 class="site-preview-heading"><%= t(".total_payments", amount: number_to_currency(total_gross_income)) %></h3>
+  <div class="width-full margin-bottom-6">
+    <p>
+      <%= t(".description") %>
+    </p>
+    <h3 class="site-preview-heading"><%= t(".total_payments", amount: number_to_currency(total_gross_income)) %></h3>
 
-  <% payments_grouped_by_employer.each_with_index do |(account_id, summary), index| %>
-    <table class="usa-table usa-table--borderless width-full">
-      <caption>
-        <% if summary[:employer_name] %>
-          <%= t(".table_caption", number: index, employer_name: summary[:employer_name]) %>
-        <% else %>
-          <%= t(".table_caption_no_name", number: index + 1) %>
-        <% end %>
-      </caption>
-      <thead class="border-top-05">
-        <tr>
-          <th class="padding-3" style="background-color: #d9e8f6;" colspan="2">
-            <h3 class="margin-0">
-              <% if summary[:employer_name] %>
-                <%= t(".total_income_from", employer_name: summary[:employer_name], amount: summary[:total]) %>
-              <% else %>
-                <%= t(".total_income_from_no_employer_name", amount: number_to_currency(summary[:total])) %>
-              <% end %>
-            </h3>
-          </th>
-        </tr>
-      </thead>
-      <tbody>
-        <% summary[:payments].each do |payment| %>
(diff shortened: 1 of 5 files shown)


app/app/controllers/cbv/summaries_controller.rb
```diff
@@ -23,6 +24,24 @@ def update
...
+          total: payments.sum { |payment| payment[:amount] }
+        }] = payments
+
+        acc
+      end
+  end
+
+  def total_gross_income
+    @payments.reduce(0) { |sum, payment| sum + payment[:gross_pay_amount] }
+  end
```
tdooner, line 43: Interesting that this is summing the `:gross_pay_amount` field of all the payments, whereas the previous method is summing the `:amount` field of the payments. Is that because `:amount` is net of deductions? Or are they the same?

Can we standardize?

app/app/controllers/cbv/summaries_controller.rb
```diff
@@ -23,6 +24,24 @@ def update
...
+      .group_by { |payment| payment[:account_id] }
+      .reduce({}) do |acc, (account_id, payments)|
+        acc[{
+          account_id: account_id,
+          employer_name: payments.first[:employer],
+          total: payments.sum { |payment| payment[:amount] }
+        }] = payments
+
+        acc
+      end
```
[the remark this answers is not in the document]
tdooner: Wow yeah this is quite the hash. A couple thoughts:

1 - We should really have test coverage any time we are doing math on paycheck amount figures. This may point towards doing this kind of thing in helper modules where we could easily pass in fake/test data.

2 - This may be more readable / understandable if it had a simpler key (a single field, like `account_id`), rather than a hash. The other fields could be in the value, right? E.g.

```ruby
@payments.reduce({}) do |hash, payment|
  account_id = payment[:account_id]
  hash[account_id] ||= {
    employer_name: payment[:employer],
    total: 0,
    payments: [],
  }
  hash[account_id][:amount] += payment[:amount]
  hash[account_id][:payments] << payment
end
```

Ultimately I don't feel strongly enough about tests or the readability of this hash that it needs to be done as part of this PR, but this might be worth considering for cleaning up in a subsequent commit/task if time permits.

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -6,27 +6,39 @@
...
+      <caption>
+        <% if account[:employer_name] %>
+          <%= t(".table_caption", number: index, employer_name: account[:employer_name]) %>
+        <% else %>
+          <%= t(".table_caption_no_name", number: index + 1) %>
+        <% end %>
+      </caption>
+      <thead class="border-top-05">
+        <tr>
+          <th class="padding-3" style="background-color: #d9e8f6;" colspan="2">
```
tdooner, line 22: is there a CSS class for this color we can use? Or apply the styling to all `<th>`?

tdooner (approved):
Looks good. Some thoughts, of course.

app/app/controllers/cbv/summaries_controller.rb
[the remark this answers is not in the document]
tdooner: Thinking about this some more, one thing we should test is that we don't end up with floating point errors when adding a bunch of `payment[:amount]` values. E.g. testing that summing the paycheck amounts `[0.1, 0.1, 0.1]` does not equal `0.30000000000000004`

app/app/views/cbv/summaries/show.html.erb
GeorgeCodes19: Good callout. I also had to resort to doing inline css for this with the same table head color.

GeorgeCodes19 (approved):

PR #107, opened 2024-07-16
Support for multiple applicant comments. One comment per income source.

GeorgeCodes19:
This PR modifies the existing `additional_info` column of the `cbv_flows` table from string column to a JSONB column. The JSON object is a hash consisting of keys which reference applicant payment account_id to an end user supplied comment.

## Ticket

Resolves [FFS-1087](https://jiraent.cms.gov/browse/FFS-1087)

## Changes

> - Added migration that sets the default column value to an empty object e.g. `{}`
> - Adds new functionality to the Payment Details Controllers
> - Adds a comment box on the payment details page
> - Adds a form to the `payment_details` page that posts to the itself and then redirects to the `add_job` url
> - Adds unit and integration tests
> - Navigating back to the payment details page preserves the comment

Screenshot 2024-07-15 at 8 35 49 PM

## Context for reviewers

The continue button for the `payment_details` page now acts a form submission which submits a form to itself. Upon updating the comment for the respective payment details the user is redirected to the jobs page where they can add another source of income.

## Testing

Screenshot 2024-07-15 at 8 37 11 PM

### DIFF
--- a/app/app/controllers/cbv/payment_details_controller.rb
+++ b/app/app/controllers/cbv/payment_details_controller.rb
@@ -1,5 +1,4 @@
 class Cbv::PaymentDetailsController < Cbv::BaseController
-  before_action :set_cbv_flow
   helper_method :employer_name,
     :start_date,
     :end_date,
@@ -28,7 +27,7 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
       updated_at: Time.current
     }
     @cbv_flow.update(additional_information: additional_information.to_json)
-    redirect_to cbv_flow_add_job_path
+    redirect_to next_path
   end
 
   def account_comment
@@ -78,13 +77,6 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
     @payments
       .map { |payment| payment[:gross_pay_amount] }
       .reduce(:+)
-  end
-
-  def parse_additional_information(json_string)
-    return {} if json_string.nil?
-    JSON.parse(json_string)
-  rescue JSON::ParserError
-    {}
   end
 
   def sanitize_comment(comment)
--- a/app/app/controllers/cbv/payment_details_controller.rb
+++ b/app/app/controllers/cbv/payment_details_controller.rb
@@ -1,4 +1,5 @@
 class Cbv::PaymentDetailsController < Cbv::BaseController
+  before_action :set_cbv_flow
   helper_method :employer_name,
     :start_date,
     :end_date,
@@ -20,14 +21,13 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
 
   def update
     account_id = params[:user][:account_id]
-    cbv_flow = CbvFlow.find(session[:cbv_flow_id])
     comment = params[:cbv_flow][:additional_information]
-    additional_information = parse_additional_information(cbv_flow.additional_information)
+    additional_information = parse_additional_information(@cbv_flow.additional_information)
     additional_information[account_id] = {
       comment: sanitize_comment(comment),
       updated_at: Time.current
     }
-    cbv_flow.update(additional_information: additional_information.to_json)
+    @cbv_flow.update(additional_information: additional_information.to_json)
     redirect_to cbv_flow_add_job_path
   end
 
@@ -92,8 +92,7 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
   end
 
   def get_additional_information
-    cbv_flow = CbvFlow.find(session[:cbv_flow_id])
-    parse_additional_information(cbv_flow.additional_information)
+    parse_additional_information(@cbv_flow.additional_information)
   end
 
   def get_comment_by_account_id(account_id)
--- a/app/app/controllers/cbv/payment_details_controller.rb
+++ b/app/app/controllers/cbv/payment_details_controller.rb
@@ -22,7 +22,7 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
   def update
     account_id = params[:user][:account_id]
     comment = params[:cbv_flow][:additional_information]
-    additional_information = parse_additional_information(@cbv_flow.additional_information)
+    additional_information = @cbv_flow.additional_information
     additional_information[account_id] = {
       comment: sanitize_comment(comment),
       updated_at: Time.current
@@ -91,12 +91,7 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
     ActionController::Base.helpers.sanitize(comment)
   end
 
-  def get_additional_information
-    parse_additional_information(@cbv_flow.additional_information)
-  end
-
   def get_comment_by_account_id(account_id)
-    additional_information = get_additional_information
-    additional_information.fetch(account_id, { comment: nil, updated_at: nil })
+    @cbv_flow.additional_information[account_id] || { comment: nil, updated_at: nil }
   end
 end
--- a/app/app/controllers/cbv/payment_details_controller.rb
+++ b/app/app/controllers/cbv/payment_details_controller.rb
@@ -7,13 +7,33 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
     :employment_end_date,
     :employment_status,
     :pay_period_frequency,
-    :compensation_amount
+    :compensation_amount,
+    :account_comment
 
   def show
     account_id = params[:user][:account_id]
     @employment = pinwheel.fetch_employment(account_id: account_id)["data"]
     @income_metadata = pinwheel.fetch_income_metadata(account_id: account_id)["data"]
     @payments = set_payments account_id
+    @account_comment = account_comment
+  end
+
+  def update
+    account_id = params[:user][:account_id]
+    cbv_flow = CbvFlow.find(session[:cbv_flow_id])
+    comment = params[:cbv_flow][:additional_information]
+    additional_information = parse_additional_information(cbv_flow.additional_information)
+    additional_information[account_id] = {
+      comment: sanitize_comment(comment),
+      updated_at: Time.current
+    }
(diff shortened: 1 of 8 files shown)


GeorgeCodes19:
I'd like to add a few more test cases.

app/app/views/cbv/payment_details/show.html.erb
```diff
@@ -1,45 +1,53 @@
...
 
 <% if @payments.any? %>
   <p><%= t(".subheader", start_date: format_date(start_date), end_date: format_date(end_date)) %></p>
-    <h2><%= t(".total_gross_income", amount: number_to_currency(gross_pay)) %></h2>
-    <p><%= t(".total_gross_description") %></p>
+  <h2><%= t(".total_gross_income", amount: number_to_currency(gross_pay)) %></h2>
+  <p><%= t(".total_gross_description") %></p>
 
+  <%= form_with(model: @cbv_flow, url: cbv_flow_payment_details_path, method: :patch) do |form| %>
+    <%= hidden_field_tag "user[account_id]", params[:user][:account_id] %>
```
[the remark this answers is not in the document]
GeorgeCodes19: You're correct, the form only needs to wrap the continue/submit button and the hidden field. I had a personal note to update this.

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -6,7 +7,9 @@
   describe "#show" do
     render_views
 
-    let(:cbv_flow) { CbvFlow.create(case_number: "ABC1234", pinwheel_token_id: "abc-def-ghi") }
+    let!(:cbv_flow) { CbvFlow.create!(case_number: "ABC1234", pinwheel_token_id: "abc-def-ghi") }
```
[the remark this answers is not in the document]
GeorgeCodes19: We don't necessarily need it, but this is a convention that I often resort to when writing tests. 

Rather than having `cbv_flow` be a boolean `(false)` e.g. `?` I'd prefer that the test throw an exception the moment something goes wrong. In this instance that would be if we cannot create the CbvFlow for whatever reason (database connectivity issues, invalid params being passed, etc.). If cbv_flow resolves to false then we're deferring an error to later in the workflow. Most likely line 15 where we're attempting to get the `cbv_flow.id`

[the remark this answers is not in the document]
tdooner: FYI - this is a fun Rails subtlety - the exclamation marks mean slightly different things:
* `create!` - the exclamation mark means to raise an exception if the creation fails
* `let!` - the exclamation mark means that the `let` will be evaluated immediately, i.e. before the example's `it` block starts running. Typically `let` blocks (without the exclamation mark) are lazily evaluated the first time they are referenced in an `it` example block

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -17,8 +20,81 @@
...
+          account_id_2 => { comment: comment_2, updated_at: updated_at }
+        }
+
+        cbv_flow.update!(additional_information: additional_information)
+
+        # Verify that the comments were saved
+        loaded_info = cbv_flow.reload.additional_information
+        expect(loaded_info[account_id]["comment"]).to eq(comment)
+        expect(loaded_info[account_id]["updated_at"]).to eq(updated_at)
+        expect(loaded_info[account_id_2]["comment"]).to eq(comment_2)
```
GeorgeCodes19, line 63: I've added some more assertions to ensure that the `additional_information` column is indeed a hash with multiple values.

app/spec/controllers/cbv/summaries_controller_spec.rb
```diff
@@ -18,19 +18,5 @@
       get :show
       expect(response).to be_successful
     end
-
-    context "when saving additional information for the caseworker" do
```
GeorgeCodes19, line 22: This didn't seem right and I believe it'll be updated. An invalid test is not a valuable test so I've removed it.

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -1,3 +1,4 @@
+# spec/controllers/cbv/payment_details_controller_spec.rb
```
tdooner, line 1: ?? why put the filename in a comment here?

app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -59,4 +79,19 @@ def gross_pay
       .map { |payment| payment[:gross_pay_amount] }
       .reduce(:+)
   end
+
+  def parse_additional_information(json_string)
+    return {} if json_string.nil?
+    JSON.parse(json_string)
+  rescue JSON::ParserError
+    {}
+  end
```
tdooner, line 88: This method seems no longer used

app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -7,13 +8,32 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
...
+  def update
+    account_id = params[:user][:account_id]
+    comment = params[:cbv_flow][:additional_information]
+    additional_information = @cbv_flow.additional_information
+    additional_information[account_id] = {
+      comment: sanitize_comment(comment),
+      updated_at: Time.current
+    }
+    @cbv_flow.update(additional_information: additional_information.to_json)
+    redirect_to cbv_flow_add_job_path
```
tdooner, line 31: Would be good to use `next_path` here instead to keep the navigation consistent across all our pages.

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -17,8 +20,81 @@
...
+      let(:updated_at) { Time.current.iso8601 }
+
+      before do
+        additional_information = { account_id => { comment: comment, updated_at: updated_at } }
+        cbv_flow.update!(additional_information: additional_information)
+
+        # Verify that the comment was saved
+        loaded_info = cbv_flow.reload.additional_information
+        expect(loaded_info[account_id]["comment"]).to eq(comment)
+        expect(loaded_info[account_id]["updated_at"]).to eq(updated_at)
```
tdooner, line 37: Doing expectations in a `before` block is pretty weird. I see why you're doing it, but, I don't think it is necessary since ideally the actual implementation would break in the actual test cases if the comment wasn't saved properly.

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -17,8 +20,81 @@
...
+    let(:account_id) { SecureRandom.uuid }
+    let(:comment) { "This is a test comment" }
+
+    before do
+      session[:cbv_flow_id] = cbv_flow.id
+    end
+
+    it "updates the account comment" do
+      patch :update, params: { user: { account_id: account_id }, cbv_flow: { additional_information: comment } }
+      additional_information = JSON.parse(cbv_flow.reload.additional_information)
```
tdooner, line 96: `JSON.parse` here seems wrong - it should be a hash already without needing to explicitly parse it (and you have tests to verify that already)

app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -1,4 +1,5 @@
 class Cbv::PaymentDetailsController < Cbv::BaseController
+  before_action :set_cbv_flow
```
tdooner, line 2: Not necessary as this is inherited from `Cbv::BaseController`

tdooner (approved):
Looks fine, pending your review of my comments.

Architecturally, if we need to do stuff like this in the future, we should just make new tables for it that have foreign keys to `cbv_flows`. We're just setting ourselves up for future pain here since it will be hard to manage the structure of these JSON blobs later if we need to make changes.

app/app/views/cbv/payment_details/show.html.erb
```diff
@@ -1,45 +1,50 @@
...
+        <td><%= employment_status %></td>
+      </tr>
+      <tr>
+        <td><%= t(".pay_period_frequency") %></td>
+        <td><%= pay_period_frequency %></td>
+      </tr>
+      <tr>
+        <td><%= t(".hourly_rate") %></td>
+        <td><%= number_to_currency(compensation_amount) %></td>
+      </tr>
```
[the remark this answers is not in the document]
GeorgeCodes19: Your git differ isn't configured to ignore whitespace. There's a setting at the top of this page to disable whitespace diffs.

app/spec/controllers/cbv/payment_details_controller_spec.rb
GeorgeCodes19: I've used this practice for years with no issues. It's defensive coding.

Sure the test will _eventually_ break, but then you're attempting to trace down the origin of the error. I'd rather assert that my before block is properly configured and catch the error earlier in the stack.

app/spec/controllers/cbv/payment_details_controller_spec.rb
GeorgeCodes19: This is correct. It seems that the `reload` method will reload the entity, but doesn't deserialize the properties. There is an older [thread](https://github.com/rails/rails/issues/25594) that might be somewhat related to JSONB columns and serialization issues.

app/app/views/cbv/payment_details/show.html.erb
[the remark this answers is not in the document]
GeorgeCodes19: My RubyMine might be configured to make some changes (maybe replace tabs with spaces or something). I switched from VS Code to RubyMine (for debugging) while working on this ticket and noticed that it formatted a few of the files after I saved.

PR #109, opened 2024-07-18
start scans

JosephGasiorekUSDS:
## Ticket

## Changes

## Context for reviewers

> Testing instructions, background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.

## Testing


acouch:
Made some progress locally w/ anchore, got it from 200+ severities to a handful by moving updates to linux-libc-dev to the release target and upgrading rails:
image
Will push an update tomorrow.

acouch:
Having trouble with the last couple of packages for anchore. They appear to be correct, or not there, in the container, but grype says they are the wrong version.

image

acouch:
The rexml package comes from the image. Grype still sees the old package. It could still be present.
image

acouch:
The rexml package was located at `/usr/local/lib/ruby/gems/3.3.0` and being installed from the base image. I was able to remove it.

acouch:
This is green but there are a couple of packages that I've added to the grype file:
image

I tried to upgrade those, for example upgrading semver in package.json but it still reports the old version. For pip I'm not sure where it is located as it is not part available in the path. 

```
docker run iv-vuln-scan-5 python3 -m sysconfig
Platform: "linux-aarch64"
Python version: "3.11"
Current installation scheme: "posix_local"

Paths:
	data = "/usr/local"
	include = "/usr/include/python3.11"
	platinclude = "/usr/include/python3.11"
	platlib = "/usr/local/lib/python3.11/dist-packages"
	platstdlib = "/usr/lib/python3.11"
	purelib = "/usr/local/lib/python3.11/dist-packages"
	scripts = "/usr/local/bin"
	stdlib = "/usr/lib/python3.11"
```

It would prob take a couple of additional hours to resolve removing those.

app/.ruby-version
```diff
@@ -1 +1 @@
-3.3.0
+3.3.4
```
tdooner, line 1: Nice, I've been meaning to update the version of Ruby anyway.

app/Dockerfile
```diff
@@ -96,20 +98,32 @@ RUN SECRET_KEY_BASE_DUMMY=1 ./bin/rails assets:precompile
...
 RUN apt-get update -qq && \
-    apt-get install -y --no-install-recommends unzip python3-venv python-is-python3 curl libvips postgresql-client wkhtmltopdf && \
+    apt-get install -y --no-install-recommends unzip=6.0-28 \
+        python3-venv=3.11.2-1+b1 \
+        python-is-python3=3.11.2-1+deb12u1 \
+        curl=7.88.1-10+deb12u6 \
+        libvips42=8.14.1-3+deb12u1 \
+        postgresql-client=15+248 \
+        linux-libc-dev=6.1.99-1 \
+        wkhtmltopdf=0.12.6-2+b1 && \
```
tdooner, line 115: Thank you for sorting through all of these versions. Kind of annoying that `apt` doesn't allow you to specify a minimum version... but I guess at least our builds will be more reproducible this way.

tdooner (approved):
Looks good! Is there more that still needs to be done here?

PR #110, opened 2024-07-19
Add terraform config for CBV production environment

tdooner:
## Ticket

Resolves FFS-1048.

## Changes

This commit is the result of:
1. Adding AWS access keys to my local ~/.aws/credentials file (under a
   new section for "prod")
2. Running:
   ```
   AWS_PROFILE=prod make infra-set-up-account ACCOUNT_NAME=nava-ffs-prod
   AWS_PROFILE=prod make infra-configure-app-service APP_NAME=app ENVIRONMENT=prod
   AWS_PROFILE=prod make infra-configure-app-build-repository APP_NAME=app
   AWS_PROFILE=prod make infra-configure-app-database APP_NAME=app ENVIRONMENT=prod
   ```
   and their corresponding `update` make targets that actually create the AWS
   resources.
3. Reviewing and making minor updates to the deploy documentation

## Context for reviewers

One note is that this changes the shared build repository to be hosted
in production. It appears the infra template is architected around a
single shared container image registry, so, keeping that in production
probably makes more sense than having it in dev(/demo) and having
production pull from there.

## Testing

See it at verify-prod.navapbc.cloud.

### DIFF
--- a/.github/workflows/cd-app.yml
+++ b/.github/workflows/cd-app.yml
@@ -4,7 +4,6 @@ name: Deploy App
 run-name: Deploy ${{ github.ref_name }} to App ${{ inputs.environment || 'dev' }}
 
 on:
-  # !! Uncomment the following lines once you've set up the dev environment and ready to turn on continuous deployment
   push:
     branches:
       - "main"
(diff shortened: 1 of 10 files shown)


GeorgeCodes19 (approved):
Looks good to me!

PR #111, opened 2024-07-22
George/ffs 1147 confirmation number

GeorgeCodes19:
## Ticket

Resolves [FFS-1147](https://jiraent.cms.gov/browse/FFS-1147)

## Changes

- Added `confirmation_number` to `cbv_flow` table
- Added a `show` method to the `Successes Controller` which generates and persists a `confirmation_number` on the `cbv_flow` record if one does not exit. This means that an end-user can visit the success page multiple times and the existing confirmation number will not be overwritten. 
- Added Tests

## Context for reviewers
Technically I've generated a confirmation code rather than number. The code is prefixed with the ID of the `cbv_flow` to avoid creating a unique code that isn't a UUID (unless a prefix is specified- then the state prefix proceeds the `cbv_flow.id` which should originate from the config). The aim was to go with something that's unique and human readable.

Screenshot 2024-07-22 at 2 09 56 PM


app/app/controllers/cbv/successes_controller.rb
```diff
@@ -1,4 +1,23 @@
...
   end
+
+  private
+
+  def confirmation_number
+    @cbv_flow["confirmation_number"]
+  end
+
+  def generate_confirmation_number(prefix = nil)
+    confirmation_number = "#{@cbv_flow.id}-#{Time.now.to_i.to_s(36)}"
```
GeorgeCodes19, line 19: `.to_i..to_s(36)` converts the Unix timestamp to an integer (representing seconds) since epoch. It then converts the timestamp integer to a base-36 string. Base-36 uses the characters `0-9` and `a-z` to represent numbers, which makes the string more compact than a base-10 representation.

GeorgeCodes19: How do we feel about exposing the internal table ID here?

app/app/controllers/cbv/successes_controller.rb
```diff
@@ -1,4 +1,23 @@
 class Cbv::SuccessesController < Cbv::BaseController
+  helper_method :confirmation_number
+
   def show
+    if @cbv_flow.confirmation_number.blank?
+      prefix = session[:state]
```
GeorgeCodes19, line 6: I presume that the resolved state partner config will be stored in the session after the initial page load. I'm not sure if that would be nested under a config option. For now we can infer the state partner from the the top-level session object.

app/app/controllers/cbv/successes_controller.rb
tdooner: A couple pieces of feedback on this, based on my experience on other projects:
1. IMHO, exposing the ID is fine as long as the confirmation number includes some other harder-to-guess elements (which it does, with the encoded timestamp). The only issue I see here is that the length of the confirmation number will change as the project grows. It's not exactly necessary for conf numbers to always be the same length but it can be a nice property to have.
2. The Base-36 encoding is an interesting idea. I kind of like it. The only problem with alphanumeric is that we should consider normalizing characters that might be mistaken for each other, e.g. `O` and `0`, `I` and `1`, etc.

With that I think I'd recommend a slightly-tweaked generation approach:

```ruby
  # produce a confirmation number like "NY28J0123" representing a NY prefix, a
  # timestamp "28J", and a CBV flow ID of 123.
  confirmation_number = [
    prefix,
    (Time.now.to_i % 36 ** 3).to_s(36).upcase.tr("OISB", "0158").rjust(3, "0"),
    @cbv_flow.id.to_s.rjust(4, "0"),
  ].compact.join
```

app/app/controllers/cbv/successes_controller.rb
```diff
@@ -3,7 +3,8 @@ class Cbv::SuccessesController < Cbv::BaseController
 
   def show
     if @cbv_flow.confirmation_number.blank?
-      confirmation_number = generate_confirmation_number
+      prefix = session[:state]
+      confirmation_number = generate_confirmation_number(prefix)
       @cbv_flow.update(confirmation_number: confirmation_number)
```
tdooner, line 8: There's a couple issues with doing this in the successes controller:

1) It won't be present at the time we send the PDF to the caseworker on the previous page. Ideally, we would include the confirmation number on the PDF that we send to the caseworker as well.
2) It's unidiomatic to do data updates in a `GET` / `show` action since GET actions are typically considered idempotent.

To me those both point towards proposing you move this to the `Cbv::SharesController` `#update` method since that's the action that most closely represents the idea of the cbv flow results being "confirmed" by the applicant.

app/app/controllers/cbv/successes_controller.rb
tdooner: I was thinking it'd be stored in the `@cbv_flow` object in the session, i.e. `@cbv_flow.site_id`. For now this will work, and then I'll go through and refactor this once we actually have the field on the `@cbv_flow` object.

app/app/controllers/cbv/successes_controller.rb
```diff
@@ -1,4 +1,23 @@
...
+      @cbv_flow.update(confirmation_number: confirmation_number)
+    end
+    render :show
+  end
+
+  private
+
+  def confirmation_number
+    @cbv_flow["confirmation_number"]
+  end
```
tdooner, line 17: Why? `@cbv_flow` should be accessible in the view.

tdooner (changes requested):
Sorry but I do think the "which controller should generate the confirmation number" question is a blocker here because this implementation would prevent us from putting it on the PDF sent to the caseworker on the previous page. Looks good to me after that is addressed.

app/app/controllers/cbv/successes_controller.rb
GeorgeCodes19: Awesome. After reading your response I pictured something like this.

app/app/controllers/cbv/shares_controller.rb
```diff
@@ -18,4 +23,14 @@ def update
...
   end
+
+  private
+
+  def generate_confirmation_code(prefix = nil)
+    [
+      prefix,
+      (Time.now.to_i % 36 ** 3).to_s(36).upcase.tr("OISB", "0158").rjust(3, "0"),
+      @cbv_flow.id.to_s.rjust(4, "0")
+    ].compact.join.upcase
```
GeorgeCodes19, line 34: Added a call to `upcase` to ensure that the confirmation codes are all caps.

tdooner: Cool, good idea. We could remove the `upcase` from the middle chunk then.
image

app/db/schema.rb
```diff
@@ -38,7 +38,9 @@
     t.string "pinwheel_token_id"
     t.uuid "pinwheel_end_user_id", default: -> { "gen_random_uuid()" }, null: false
     t.jsonb "additional_information", default: {}
+    t.string "confirmation_number"
```
tdooner, line 41: Changing migrations after they are run is tricky business - you have to run `rails db:rollback` before changing the migration so it'll delete the old column. Since your local database is in a weird state, you should fix it by dropping the old `confirmation_number` column and then removing this line from the db/schema.rb file.

app/spec/controllers/cbv/shares_controller_spec.rb
```diff
@@ -22,12 +18,37 @@
...
+
+      before do
+        cbv_flow.update(confirmation_code: existing_confirmation_code)
+      end
+
+      it "does not override the existing confirmation code" do
+        expect(cbv_flow.reload.confirmation_code).to eq(existing_confirmation_code)
+        put :update
+        # ensure that the confirmation code is not changed
+        expect(cbv_flow.reload.confirmation_code).to eq(existing_confirmation_code)
```
tdooner, line 48: Fun rspec tip, they have a shorthand for this that you may see elsewhere:

```ruby
expect do
  put :update
end.not_to change { cbv_flow.reload.confirmation_code }
```

tdooner (approved):
Looks good. Please remove the old `confirmation_number` column from the db/schema.rb file before merging.

PR #113, opened 2024-07-23
Change site agency_name's to "HRA" and "DTA"

tdooner:
## Ticket

N/A

## Changes

We're going with acronyms for these since user testing found them to be
well-known by applicants.

## Context for reviewers

N/A

## Testing

N/A


tdooner:
Never mind, we'll keep the full name in this agreement box, per Slack thread.

PR #115, opened 2024-07-25
Require TLS when sending emails via AWS SES

tdooner:
## Ticket

Finishes FFS-1022.

## Changes

* Create a configuration set that requires TLS
* Set the default configuration set for the AWS SES sending identity

  Note: The `default_configuration_set` parameter is only allowed for
  `aws_sesv2_email_identity` resources (not the previously-used
  `aws_ses_domain_identity` resource), so this commit upgrades the
  resource to that. (It forces recreation.)
* Grant the task executor role permission to send email with the new
  configuration set

## Context for reviewers

See ticket.

## Testing

Tested by running in development environment and verifying that an email still
goes through. I'm not sure how to test the failure mode of a non-TLS email
server. I will run it in production after merging.

### DIFF
--- a/infra/app/service/main.tf
+++ b/infra/app/service/main.tf
@@ -96,7 +96,10 @@ resource "aws_iam_policy" "email_access_policy" {
 data "aws_iam_policy_document" "email_access_policy" {
   statement {
     actions   = ["ses:SendRawEmail", "ses:SendEmail"]
-    resources = ["arn:aws:ses:${data.aws_region.current.name}:${data.aws_caller_identity.current.id}:identity/*"]
+    resources = [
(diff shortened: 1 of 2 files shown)


GeorgeCodes19:
So if I'm understanding things correctly...

In order to support TLS we need to configure our Terraform email module i.e.  `aws_ses_configuration_set::delivery_options::tls_policy="Require"` , but doing so requires that we upgrade to `sesv2` as a prerequisite.

Consequentially, there is no `aws_sesv2_domain_identity` terraform module so we'll need to use `aws_sesv2_email_identity` ?

GeorgeCodes19 (approved):

tdooner:
@GeorgeCodes19 Yep, that's right!

PR #116, opened 2024-07-27
OWASP runner

joeyg:
## Ticket

Resolves [FFS-1083](https://jiraent.cms.gov/browse/FFS-1083)

## Changes

Make the OWASP running actually run.

## Context for reviewers
[Link](https://github.com/joeyg/iv-cbv-payroll/actions/runs/10120352661/job/27989999390)

There are a few failures. Not sure what the team would like to do about those:
```WARN-NEW: CSP: Notices [10055] x 6 
	http://localhost:3000/ (200 OK)
	http://localhost:3000/ (200 OK)
	http://localhost:3000/ (200 OK)
	http://localhost:3000/cbv/agreement (200 OK)
	http://localhost:3000/cbv/entry (200 OK)
WARN-NEW: Permissions Policy Header Not Set [10063] x 4 
	http://localhost:3000/ (200 OK)
	http://localhost:3000/cbv/agreement (200 OK)
	http://localhost:3000/cbv/entry (200 OK)
	http://localhost:3000/assets/application-5b1daa56fc52f867eb19dfecd7e4e06be4255b795514de338a8bdaebc71a9730.js (200 OK)
WARN-NEW: Dangerous JS Functions [10110] x 1 
	http://localhost:3000/assets/application-5b1daa56fc52f867eb19dfecd7e4e06be4255b795514de338a8bdaebc71a9730.js (200 OK)
FAIL-NEW: 0	FAIL-INPROG: 0	WARN-NEW: 3	WARN-INPROG: 0	INFO: 0	IGNORE: 0	PASS: 131
```

PR #122, opened 2024-07-31
owasp action

acouch:
## Ticket

Resolves [FFS-1083](https://jiraent.cms.gov/browse/FFS-1083)

## Changes

Make the OWASP running actually run. Started: https://github.com/DSACMS/iv-cbv-payroll/pull/116

This adds a couple of ignore: 

```
10110	IGNORE	Dangerous JS Functions
# There are necessary and properly scoped in content_security_policy.rb
10055	IGNORE	CSP: Notices
# Perms Policy not supported in rails https://github.com/rails/rails/issues/48878
# Also not fully adopted: https://caniuse.com/permissions-policy
10063	IGNORE	Permissions Policy Header Not Set
```

Unfortunately, [Zap does not support ignoring risk levels like LOW](https://groups.google.com/g/zaproxy-users/c/Vdan6QRHfRM), so ignoring a full check is the only way to filter.


.github/workflows/owasp-daily-scan.yml
```diff
@@ -1,10 +1,12 @@
 name: OWASP ZAP daily scan
 
 on:
-  schedule:
-    # cron format: 'minute hour dayofmonth month dayofweek'
-    # this will run at noon UTC every day (7am EST / 8am EDT)
-    - cron: '0 12 * * *'
+  pull_request:
+    branches: [main]
```
tdooner, line 5: nit: File is misnamed if this is no longer a daily scan

app/app/controllers/pages_controller.rb
```diff
@@ -1,4 +1,7 @@
 class PagesController < ApplicationController
   def home
+    unless params["format"].nil?
+      head 401, content_type: "text/html"
+    end
```
tdooner, line 5: I wonder if this is still necessary given your change to the `routes.rb` you've discussed in Slack

tdooner (approved):

.github/workflows/owasp-daily-scan.yml
acouch: Good catch, I renamed to `owasp-scan.yml` so the name is separate from the trigger.

app/app/controllers/pages_controller.rb
acouch: This was introduced in https://github.com/DSACMS/iv-cbv-payroll/pull/116 . 

@joeyg was there a reason this was added? I'm reading it as returning unauthorized if a request to the home page doesn't have a request format.

joeyg: @acouch One of the checks is for files like ._hg and the like.

acouch: @joeyg Thanks, I can confirm that https://github.com/DSACMS/iv-cbv-payroll/pull/124 fixes that. I'm going to remove from here and rebase once #124 is pulled in.

PR #126, opened 2024-08-01
1204: SSO Spike

allthesignals:
## Ticket

Resolves 1204

## Changes

Integrates pilot-specific SSO with test Azure credentials.

## Context for reviewers

Currently WIP. Integrates SSO with a combination of Devise, Omniauth, and an Azure AD Omniauth adapter gem.

## Testing

### DIFF
--- a/app/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/app/controllers/users/omniauth_callbacks_controller.rb
@@ -1,8 +1,23 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def ma_dta
     response_params = request.env["omniauth.auth"]["info"]
+    email = response_params["email"]
+
+    login_with_oauth(email, "ma")
+  end
+
+  def sandbox
+    response_params = request.env["omniauth.auth"]["info"]
+    email = response_params["email"]
+
+    login_with_oauth(email, "sandbox")
+  end
+
+  private
+
+  def login_with_oauth(email, site_id)
     # TODO: Check that the email is permissible according to its domain
-    @user = User.find_or_create_by(email: response_params["email"], site_id: :ma)
+    @user = User.find_or_create_by(email: email, site_id: site_id)
 
     if @user&.persisted?
       flash[:notice] = "Signed in!"
--- a/app/config/initializers/devise.rb
+++ b/app/config/initializers/devise.rb
@@ -287,6 +287,14 @@ Devise.setup do |config|
     }
   )
 
+  config.omniauth(
+    :sandbox,
+    {
+      **Rails.application.config.sites["sandbox"].sso,
+      strategy_class: OmniAuth::Strategies::AzureActivedirectoryV2
+    }
+  )
+
   # ==> Warden configuration
   # If you want to use other strategies, that are not supported by Devise, or
   # change the failure app, you can configure them inside the config.warden block.
--- a/app/config/site-config.yml
+++ b/app/config/site-config.yml
@@ -37,10 +37,5 @@
     client_id:     <%= ENV["AZURE_SANDBOX_CLIENT_ID"] %>
     client_secret: <%= ENV["AZURE_SANDBOX_CLIENT_SECRET"] %>
     tenant_id:     <%= ENV["AZURE_SANDBOX_TENANT_ID"] %>
-    redirect_uri:  <%= ENV["AZURE_SANDBOX_REDIRECT_URI"] %>
-    response_type: "id_token"
-    response_mode: "form_post"
     scope: "openid"
-    state: "12345"
-    nonce: "678910"
     name: "sandbox"
--- a/app/app/models/user.rb
+++ b/app/app/models/user.rb
@@ -1,5 +1,5 @@
 class User < ApplicationRecord
-  devise :omniauthable, omniauth_providers: %i[ma_dta nyc_dss]
+  devise :omniauthable, omniauth_providers: %i[ma_dta nyc_dss sandbox]
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
   devise :rememberable, :trackable, :timeoutable, :omniauthable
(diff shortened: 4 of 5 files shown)


app/app/controllers/cbv_flow_invitations_controller.rb
```diff
@@ -1,5 +1,8 @@
 class CbvFlowInvitationsController < ApplicationController
-  before_action :ensure_valid_params!
```
tdooner, line 2: We should still have this other `before_action` here.

app/app/controllers/application_controller.rb
```diff
@@ -2,6 +2,10 @@ class ApplicationController < ActionController::Base
   helper :view
   around_action :switch_locale
 
+  def new_session_path
+    new_user_session_path
+  end
```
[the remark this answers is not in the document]
tdooner: Hmm, yeah - where is `new_session_path` used? Can we just use `new_user_session_path` instead?

app/app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  def ma_dta
+    response_params = request.env["omniauth.auth"]["info"]
+    # TODO: Check that the email is permissible according to its domain
+    @user = User.find_or_create_by(email: response_params["email"])
+
+    if @user&.persisted?
+      flash[:notice] = "Signed in!"
+      sign_in_and_redirect @user, event: :authentication
+    else
+      flash[:alert] = "You have not yet an account!"
```
tdooner, line 11: "You have not yet an account!"

lol, what

app/app/views/cbv_flow_invitations/index.html.erb
tdooner, line 1: wat?

accidentally created file?

app/app/views/sso/_nyc_signin_button.html.erb
```diff
@@ -0,0 +1 @@
+<%= button_to "Sign in NYC DSS", user_nyc_dss_omniauth_authorize_path, data: { turbo: false }, class: "usa-button" %>
```
tdooner, line 1: Can we just implement this as we intend to actually do... instead of using the buttons, just make it so that going to /nyc/invitations/new, when logged-out, redirects to `user_nyc_dss_omniauth_authorize_path`?

app/config/site-config.yml
```diff
@@ -3,7 +3,11 @@
   transmission_method: shared_email
   transmission_method_configuration:
     email: <%= ENV['NYC_HRA_EMAIL'] %>
+  sso:
+    provider: nyc
```
tdooner, line 7:
```suggestion
    provider: nyc_dss
```

I'm not sure whether this config is accessible from the devise initialization but it would be nice to see if we can get the various config in here (tenant ID, client ID, client secret), just to keep all the environment variables visible in one place. We could take that on later.

app/app/views/sso/_nyc_signin_button.html.erb
tdooner: As discussed on slack - never mind - you're right that we will need these buttons for the eventual login page, so carry on.

.dockleconfig
```diff
@@ -1,4 +1,4 @@
 # This file is allows you to specify a list of files that is acceptable to Dockle
 # To allow multiple files, use a list of names, example below. Make sure to remove the leading #
-DOCKLE_ACCEPT_FILES=database.yml
+DOCKLE_ACCEPT_FILES=database.yml,omniauth.rb
```
tdooner, line 3: ah ha, nice.

app/config/initializers/devise.rb
```diff
@@ -0,0 +1,327 @@
...
+  # The default HTTP method used to sign out a resource. Default is :delete.
+  config.sign_out_via = :delete
+
+  # ==> OmniAuth
+  # Add a new OmniAuth provider. Check the wiki for more information on setting
+  # up on your models and hooks.
+  config.omniauth(
+    :ma_dta,
+    {
+      **Rails.application.config.sites["ma"].sso,
```
tdooner, line 277: Interesting - not sure I'm a huge fan of this (since it might be buggy) but I do see the convenience so I'm down to try it.

app/config/site-config.yml
```diff
@@ -6,13 +6,35 @@
...
+    client_id:     <%= ENV["AZURE_NYC_DSS_CLIENT_ID"] %>
+    client_secret: <%= ENV["AZURE_NYC_DSS_CLIENT_SECRET"] %>
+    tenant_id:     <%= ENV["AZURE_NYC_DSS_TENANT_ID"] %>
+    redirect_uri:  <%= ENV["AZURE_NYC_DSS_REDIRECT_URI"] %>
+    response_type: "id_token"
+    response_mode: "form_post"
+    scope: "openid"
+    state: "12345"
+    nonce: "678910"
+    name: "nyc_dss"
```
tdooner, line 19: Where did we get this list of config options from? I'm cross-referencing with [the Azure AD gem](https://github.com/RIPAGlobal/omniauth-azure-activedirectory-v2/tree/master?tab=readme-ov-file#configuration-options) and there's a bunch here that aren't listed. I just tried this locally with the following config and it seemed to work:

```suggestion
    client_id:     <%= ENV["AZURE_NYC_DSS_CLIENT_ID"] %>
    client_secret: <%= ENV["AZURE_NYC_DSS_CLIENT_SECRET"] %>
    tenant_id:     <%= ENV["AZURE_NYC_DSS_TENANT_ID"] %>
    scope: "openid"
    name: "nyc_dss"
```

I was kind of surprised to not need `redirect_uri`, nor the response_type ones, but I think they get added automatically. Can you try that and see if it works? I looked at this because it includes `state` and `nonce` which definitely feels wrong. (I believe those are meant to be parameters passed in uniquely for each authentication attempt.)

tdooner (approved):
Looks good conditional upon review of the configuration options. I think there are too many options provided and many of them don't do anything.

acouch:
Is there documentation for local dev that should be updated? Can a dev sign in locally?

PR #138, opened 2024-08-06
Listen for a combination of events

allthesignals:
Addresses [1263](https://jiraent.cms.gov/browse/FFS-1263)

## Changes

This PR tracks received events and ensures a combination of them are available. 

## Context for reviewers

Be sure to restart your server. We will need to add new webhook events for lower deployed environments:

```
    employment.added
    income.added
    paystubs.fully_synced
```

I don't love this approach. I think we should probably get an accounts model that belongs to a cbv_flow. That way we can track account logic better.

### DIFF
--- a/app/app/controllers/webhooks/pinwheel/events_controller.rb
+++ b/app/app/controllers/webhooks/pinwheel/events_controller.rb
@@ -25,7 +25,7 @@ class Webhooks::Pinwheel::EventsController < ApplicationController
       pinwheel_account = PinwheelAccount.find_by_pinwheel_account_id(params["payload"]["account_id"])
       pinwheel_account.update!(EVENTS_MAP[params["event"]] => Time.now) if pinwheel_account.present?
 
-      if pinwheel_account.has_full_synced?
+      if pinwheel_account.has_fully_synced?
         PaystubsChannel.broadcast_to(cbv_flow, {
           event: "cbv.payroll_data_available",
           account_id: params["payload"]["account_id"]
--- a/app/app/models/pinwheel_account.rb
+++ b/app/app/models/pinwheel_account.rb
@@ -1,7 +1,7 @@
 class PinwheelAccount < ApplicationRecord
   belongs_to :cbv_flow
 
-  def has_full_synced?
+  def has_fully_synced?
     paystubs_synced_at.present? && employment_synced_at.present? && income_synced_at.present?
   end
 end


GeorgeCodes19 (approved):

app/app/controllers/webhooks/pinwheel/events_controller.rb
```diff
@@ -1,26 +1,46 @@
...
   skip_before_action :verify_authenticity_token
 
-  def create
-    signature = request.headers["X-Pinwheel-Signature"]
-    timestamp = request.headers["X-Timestamp"]
+  DUMMY_API_KEY = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+  EVENTS_MAP = {
+    "employment.added" => "employment_synced_at",
+    "income.added" => "income_synced_at",
+    "paystubs.fully_synced" => "paystubs_synced_at"
```
tdooner, line 9: Just flagging that we will probably need to re-subscribe the webhooks in Demo/Prod in order to add this event to it.

app/app/controllers/webhooks/pinwheel/events_controller.rb
```diff
@@ -1,26 +1,46 @@
...
+    signature = request.headers["X-Pinwheel-Signature"]
+    timestamp = request.headers["X-Timestamp"]
+
     pinwheel = cbv_flow.present? ? pinwheel_for(cbv_flow) : PinwheelService.new(DUMMY_API_KEY, "sandbox")
     digest = pinwheel.generate_signature_digest(timestamp, request.raw_post)
     unless pinwheel.verify_signature(signature, digest)
-      return render json: { error: "Invalid signature" }, status: :unauthorized
+      render json: { error: "Invalid signature" }, status: :unauthorized
+    end
+  end
```
tdooner, line 25: nit: conventionally, `before_action`s are put below a `private` access modifier lower in the file. I love the idea of breaking it out into a `before_action`.

app/app/controllers/webhooks/pinwheel/events_controller.rb
```diff
@@ -1,26 +1,46 @@
...
+
+  def create
+    cbv_flow = CbvFlow.find_by_pinwheel_end_user_id(params["payload"]["end_user_id"])
```
tdooner, line 28: We could use a guard clause here if the cbv_flow is missing to simplify the future conditionals. Maybe something like
```suggestion
    cbv_flow = CbvFlow.find_by_pinwheel_end_user_id(params["payload"]["end_user_id"])
    unless cbv_flow
      Rails.logger.info "Unable to find CbvFlow for end_user_id: #{params["payload"]["end_user_id"]}"
      return render json: { status: "ok" }
    end
```

app/app/controllers/webhooks/pinwheel/events_controller.rb
```diff
@@ -1,26 +1,46 @@
...
+
+    if cbv_flow && params["event"] == "account.added"
+      pinwheel_account = PinwheelAccount.find_or_create_by(pinwheel_account_id: params["payload"]["account_id"])
+      pinwheel_account.update!(cbv_flow: cbv_flow)
```
tdooner, line 32: Here's a fun possibility (optional):
```suggestion
      pinwheel_account = PinwheelAccount.create_with(cbv_flow: cbv_flow).find_or_create_by(pinwheel_account_id: params["payload"]["account_id"])
```

app/app/models/pinwheel_account.rb
```diff
@@ -0,0 +1,7 @@
+class PinwheelAccount < ApplicationRecord
+  belongs_to :cbv_flow
+
+  def has_full_synced?
```
tdooner, line 4: nit: typo - `has_fully_synced?`

tdooner (approved):
Looks good. Some optional items you could consider, and either address now or subsequently if you just want to merge this to unblock Aaron.

app/app/controllers/webhooks/pinwheel/events_controller.rb
tdooner: Yeah that's right.

PR #150, opened 2024-08-07
Dynamically resolve the applicant (CBV) invite form based on site_id

GeorgeCodes19:
## Ticket
 resolves [FFS-1020](https://jiraent.cms.gov/browse/FFS-1020).
## Changes

- Infer the `site_id` from the params when loading the new invitation page, making it accessible to the responsible view handler.
- Update the view to load applicable state invitation form based on the `site_id`
- Updated invitation service to accept controller parameters over explicit args
- Organized test suites for respective `site_ids`
- Created a migration for new fields for the `cbv_flows_invitation` table
- Added new en locales for the form fields

Screenshot 2024-08-09 at 6 03 29 PM


app/app/controllers/cbv_flow_invitations_controller.rb
```diff
@@ -42,4 +43,13 @@ def cbv_flow_invitation_params
...
     )
   end
+
+  def sanitized_site_id
+    sanitize_site_id(params[:site_id])
+  end
+
+  def sanitize_site_id(site_id)
+    # Sanitize the site_id to prevent directory traversal attacks
+    site_id.to_s.gsub(/\A\//, '').gsub(/\A\.\.\//, '').gsub(/\/\.\.\//, '/')
```
GeorgeCodes19, line 53: Because `site_id` is used in a view to read a partial. This is just being particular about not allowing the `site_id` to have a leading `/` or `../` (or combination of) and enabling a malicious user to load .erb templates and partials since the project structure is public (open source)

tdooner: Why is checking the site ID is specified in the configuration file not sufficient here? If an attacker tried directory traversal, it would just fail the check since "../../../../../../../etc/passwd" is not a site we've configured in the YAML file... right?

app/spec/controllers/cbv_flow_invitations_controller_spec.rb
```diff
@@ -36,10 +42,36 @@
...
+          get :new, params: nyc_params
 
-        expect(response).to be_successful
+          expect(response.body).to include("first_name")
+          expect(response.body).to include("middle_name")
+          expect(response.body).to include("last_name")
+          expect(response.body).to include("client_id_number")
+          expect(response.body).to include("case_number")
+          expect(response.body).to include("email_address")
+          expect(response.body).to include("snap_application_date")
```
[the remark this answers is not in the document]
tdooner: I don't think `hash_including` will work because `response.body` isn't a hash, it's a string. I don't mind the repeated `include` statements, personally.

app/spec/controllers/cbv_flow_invitations_controller_spec.rb
```diff
@@ -10,8 +12,12 @@
     stub_environment_variable("CBV_INVITE_SECRET", invite_secret, &ex)
   end
 
+  let(:sandbox_params) { { site_id: "sandbox", secret: invite_secret } }
+  let(:nyc_params) { { site_id: "nyc", secret: invite_secret } }
+  let(:ma_params) { { site_id: "ma", secret: invite_secret } }
```
tdooner, line 17: FYI - we can remove the `invite_secret` bits from the test since we no longer use that for authentication. (We are using the `sign_in user` on line 40)

app/spec/controllers/cbv_flow_invitations_controller_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe CbvFlowInvitationsController do
   let(:invite_secret) { "FAKE_INVITE_SECRET" }
+
+  # hoisted here so that it's accessible to the test suite hooks .e.g. "before"
```
tdooner, line 6: This comment appears to be incorrect - each test case is specifying `site_id` separately.

tdooner (commented):
How did you make the database changes? Did you add all these as columns to `cbv_flow_invitations` table?

GeorgeCodes19:
> How did you make the database changes? Did you add all these as columns to `cbv_flow_invitations` table?

Please refer to the PR description. I wanted to separate the view and data persistence.

app/app/controllers/cbv_flow_invitations_controller.rb
GeorgeCodes19: Sure, either way we'll need a separate method rather than pulling the site_id directly from the params to satisfy the Brakeman check.

tdooner:
> I wanted to separate the view and data persistence.

Ah my bad, sorry for missing that in your description. I don't think it will work to split it into two PRs, though, as Rails requires the database fields to be present in order to render the form fields. You could add all the database fields in the first PR, I guess, but it's probably just easiest to do it all in one PR.

app/config/locales/en.yml
```diff
@@ -140,13 +140,29 @@ en:
       invite_failed: 'Error sending invitation to %{email_address}: %{error_message}.'
       invite_success: Successfully delivered invitation to %{email_address}.
     incorrect_site_id: Unable to send invitation due to missing site configuration.
+    ma:
+      invite:
+        agency_id_number: Client's agency ID number
+        beacon_id: Your BEACON ID
+        email_address: Email address
+        first_name: First name
```
GeorgeCodes19, line 148: Whoops, overlooked these locales. A few should mirror the nyc locales (first_name, middle_name, last_name, email_address). It would great to eventually move some of these to shared namespace.

app/spec/controllers/cbv_flow_invitations/auth_spec.rb
```diff
@@ -62,7 +70,7 @@
     before do
       allow_any_instance_of(CbvInvitationService)
         .to receive(:invite)
-        .with("test@example.com", "ABC1234", site_id)
+        .with(ActionController::Parameters.new(email_address: "test@example.com", case_number: "ABC1234", site_id: site_id).permit!)
```
GeorgeCodes19, line 73: There's some code duplication that can be cleaned up.

app/spec/controllers/cbv_flow_invitations/nyc_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+              expect(response.body).to include("email_address")
+              expect(response.body).to include("snap_application_date")
+            end
+        end
+    end
+
+  describe "#create" do
+    let(:cbv_flow_invitation_params) do
+      {
+        first_name: "John",
```
GeorgeCodes19, line 34: This looks like it's missing the `CIN` field

carmenrosalop (changes requested):
Looks good! 
There are 3 small things that need updating:
- When the calendar opens, it should look like this screenshot (closer to the component). Right now it's a bit far and floating
- I'd remove the word "Format" from the hint text "mm/dd/yyyy" because in other hint texts we won't add that word so we can keep it consistent
- Spacing between the "SNAP application date", hint text "mm/dd/yyyy" and the calendar should be corrected to look like the screenshot


Screenshot 2024-08-09 at 9 46 05 AM

app/spec/controllers/cbv_flow_invitations/nyc_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+              expect(response.body).to include("first_name")
+              expect(response.body).to include("middle_name")
+              expect(response.body).to include("last_name")
+              expect(response.body).to include("client_id_number")
+              expect(response.body).to include("case_number")
+              expect(response.body).to include("email_address")
+              expect(response.body).to include("snap_application_date")
+            end
+        end
+    end
```
tdooner, line 29: The ruby community is very opinionated that 2-space indents are the way to go. We could update rubocop to just do it for us automatically.

app/spec/controllers/cbv_flow_invitations/nyc_spec.rb
tdooner: CIN = `client_id_number` doesn't it? I see it below.

tdooner (approved):
Looks good to me, pending removal of the coercion to a regular hash (or an explanation of why it's necessary). ~Please pair with Carmen and run it by her before merging!~ Never mind I see she weighed in above.

carmenrosalop:
@GeorgeCodes19 I just caught this now, the text under the H1 needs to be updated. See image image

carmenrosalop (approved):
LGTM!

PR #156, opened 2024-08-09
Add support to infra template for cron jobs

tdooner:
## Ticket

Resolves FFS-1272.

## Changes

We want to run a job every hour to redact data older than a threshold (7
days). This is pretty similar to the support the infra template already
has for file-upload-triggered jobs, so I've put the configuration right
alongside those.

The task is scheduled with EventBridge Scheduler, which runs a
serverless cron. The Scheduler itself has a different service identifier
than the event manager used for the file-upload jobs, so I created a
separate IAM user and policy to allow it to AssumeRole.

## Context for reviewers

This can be merged independently of https://github.com/DSACMS/iv-cbv-payroll/pull/155.

## Testing

Tested by provisioning the scheduled task and watching it run at the top of the next hour.

### DIFF
--- a/infra/app/app-config/env-config/jobs.tf
+++ b/infra/app/app-config/env-config/jobs.tf
@@ -11,4 +11,18 @@ locals {
     #   task_command = ["python", "-m", "flask", "--app", "app.py", "etl", "<object_key>"]
     # }
   }
+
+  # Configuration for cron jobs to run in every environment.
+
+  # The `schedule_expression` supports cron format (min, hour, day_of_month,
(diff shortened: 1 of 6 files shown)


GeorgeCodes19 (commented):
Seems to be correctly configured, but I'll await Loren's feedback

tdooner:
I'm going to close this for now while awaiting the platform team's official implementation, because I'd rather implement it with whatever they come up with. If they don't implement something by the time we need this, I'll reopen.

See: https://github.com/navapbc/platform-test/pull/130

PR #164, opened 2024-08-12
Configure premailer to drop the stylesheet

allthesignals:
## Changes

Premailer was including the entire stylesheet with every email. This cause the email to be hidden under an "click to expand this message" link in email viewers.

## Context for reviewers

We configured it with the rule, `drop_unmergeable_css_rules`, which was not obvious at first. See discussion: https://github.com/fphilipe/premailer-rails/issues/184

## Testing

image


tdooner (approved):
💯

PR #165, opened 2024-08-13
Track error events

allthesignals:
## Ticket

Resolves [1288](https://jiraent.cms.gov/browse/FFS-1288)

## Changes

Adds error event columns and tracks accordingly.

## Context for reviewers

Does not yet handle for these errors.

## Testing

Tested manually using this account: https://docs.pinwheelapi.com/public/docs/sandbox#datanotavailable


app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -44,50 +43,62 @@ def account_comment
...
+    @pinwheel_account.supported_jobs.include?("employment") && @pinwheel_account.employment_errored_at.blank?
+  end
+
+  def has_paystubs_data?
+    @pinwheel_account.supported_jobs.include?("paystubs") && @pinwheel_account.paystubs_errored_at.blank?
+  end
+
   def employer_name
-    @employment["employer_name"]
+    @employment ? @employment["employer_name"] : I18n.t("cbv.payment_details.show.unknown")
```
tdooner, line 59: Why not use those fancy new methods here?

```suggestion
    return I18n.t("cbv.payment_details.show.unknown") unless has_employment_data?

    @employment["employer_name"]
```

tdooner (changes requested):
Looks good, but please write tests to verify that:
1. The payment details page renders successfully when income data is not available
2. The three `errored_at` timestamps are properly set when a pinwheel webhook returns the corresponding error. AFAIK, we only have seen the error occur with income data, so just a test for that one should at least prevent future regressions, if we're not able to easily write tests for all three.

app/app/controllers/cbv/payment_details_controller.rb
tdooner: Ah, I didn't consider that. I mean, yeah, it's kind of awkward. (You can always feel free to push back on my suggestions if you think they'll suck, heh.) The most important thing to me is that there is some test coverage of how these pages render with different subsets of data so we don't accidentally break it in the future if we were to improve the way these methods work later.

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -10,7 +10,19 @@
     let(:account_id) { SecureRandom.uuid }
     let(:comment) { "This is a test comment" }
     let(:supported_jobs) { %w[income paystubs employment] }
-    let!(:pinwheel_account) { PinwheelAccount.create!(cbv_flow: cbv_flow, pinwheel_account_id: account_id, supported_jobs: supported_jobs) }
+    let(:income_errored_at) { nil }
+    let(:paystubs_errored_at) { nil }
+    let(:employment_errored_at) { nil }
+    let!(:pinwheel_account) do
+      PinwheelAccount.create!(
```
GeorgeCodes19, line 17: Some recently merged changes now allow us to use a [factory](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/spec/support/Factories.rb) to create models. 

When you want to use the factory you can simply call the `create` method provided by `FactoryBot`. Example [here](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/spec/models/cbv_flow_spec.rb#L5)

https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#using-factories

GeorgeCodes19 (approved):
LGTM

docs/app/rendered/database-schema.pdf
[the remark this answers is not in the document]
tdooner: ah, my bad

PR #167, opened 2024-08-13
Required fields

allthesignals:
Closes [FFS-1300](https://jiraent.cms.gov/browse/FFS-1300)
## Changes

Adds not null constraint for fields 

## Context for reviewers

For previous fields, provides some default

## Testing

- I've confirmed locally that the data in the `cbv_flows_invitation` table is correct. The application also shows a successful flash notification.
- Users cannot proceed with an invalid date provided in the `snap_application_date` input field due to client-side validation.

Screenshot 2024-08-14 at 11 27 48 AM
Screenshot 2024-08-14 at 11 31 53 AM
## Javascript Disabled
Screenshot 2024-08-14 at 11 37 57 AM
## Required fields for MA
Screenshot 2024-08-14 at 1 30 07 PM
## Required fields for NYC
Screenshot 2024-08-14 at 1 30 07 PM


app/db/migrate/20240813181942_require_fields_for_cbv_invitation.rb
```diff
@@ -0,0 +1,10 @@
+class RequireFieldsForCbvInvitation < ActiveRecord::Migration[7.1]
+  def change
+    change_column_null :cbv_flow_invitations, :first_name, false, "Unknown"
+    change_column_null :cbv_flow_invitations, :middle_name, false, "Unknown"
+    change_column_null :cbv_flow_invitations, :last_name, false, "Unknown"
+    change_column_null :cbv_flow_invitations, :agency_id_number, false, "Unknown"
+    change_column_null :cbv_flow_invitations, :client_id_number, false, "Unknown"
```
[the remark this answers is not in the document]
GeorgeCodes19: Right, I think the CbvFlowInvitation model should be updated to include a new validation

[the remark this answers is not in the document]
GeorgeCodes19: Right. I'm not sure about the Rails-way of doing it, but we should probably keep this as optional. A bunch of tests will fail after this. The alternative would be to handle the required fields in the controller.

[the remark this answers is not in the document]
GeorgeCodes19: Yeah I can it on. I'm nearly done with it, but have a few tests to iron out.

app/app/models/cbv_flow_invitation.rb
```diff
@@ -30,4 +35,14 @@ def expired?
   def to_url
     Rails.application.routes.url_helpers.cbv_flow_entry_url(token: auth_token)
   end
+
+  private
+
+  def nyc_site?
```
GeorgeCodes19, line 41: Is there a better way of doing this? We can't directly compare in the validator itself due to `site_id` not being defined on the Model. Open to suggestions, but this gets the balling rolling.

app/db/schema.rb
```diff
@@ -26,12 +26,12 @@
...
-    t.string "first_name"
-    t.string "middle_name"
-    t.string "last_name"
+    t.string "first_name", null: false
+    t.string "middle_name", null: false
+    t.string "last_name", null: false
     t.string "agency_id_number"
     t.string "client_id_number"
-    t.date "snap_application_date"
+    t.date "snap_application_date", null: false
```
GeorgeCodes19, line 34: ehhh... I might walk this back.

app/spec/models/cbv_flow_spec.rb
```diff
@@ -2,7 +2,15 @@
 
 RSpec.describe CbvFlow, type: :model do
   describe ".create_from_invitation" do
-    let(:cbv_flow_invitation) { CbvFlowInvitation.create!(case_number: "ABC1234", site_id: "sandbox") }
+    let(:cbv_flow_invitation) { CbvFlowInvitation.create!(
+      first_name: "John",
```
GeorgeCodes19, line 6: Lots of code duplication regarding the fields needed to create a `CbvFlowInvitation` without getting an error for missing fields. This should probably be abstracted into a test helper/module

app/app/models/cbv_flow_invitation.rb
tdooner: This seems fine to me for now since we only have one page that's creating and modifying CbvFlowInvitation objects.

If we wanted to come up with an abstraction to move the site-specific field validation logic out of the model, a common Rails pattern is a "form object" which is a kind of wrapper of the `params` hash for a specific page, and can handle which fields are displayed and validated (i.e. there could be one form object for the MA form fields and one for the NYC form fields). I'm inclined to say we don't need that level of abstraction yet.

app/db/schema.rb
tdooner: Yes - please make sure the migration has a `change_column_null` for `snap_application_date`, or else we will have a different schema between dev/test and production (that'd be bad!)

The migration also allows NULLs for `agency_id_number` field, which seems correct, but unnecessary, since it already allows nulls. So you could remove that line from the migration.

app/spec/controllers/cbv/entries_controller_spec.rb
```diff
@@ -18,7 +18,14 @@
...
     context "when following a link from a flow invitation" do
-      let(:invitation) { CbvFlowInvitation.create(case_number: "ABC1234", site_id: "sandbox") }
+      let(:invitation) { CbvFlowInvitation.create(
+        first_name: "John",
+        middle_name: "Doe",
+        last_name: "Smith",
+        case_number: "ABC1234",
+        site_id: "sandbox",
+        email_address: "test@test.com",
+        snap_application_date: Date.today) }
```
tdooner, line 28: style nit: Use `do/end` for multiline blocks in Ruby

(e.g.
```ruby 
let(:invitation) do
  CbvFlowInvitation.create!(
    ...
  )
end
```
)

app/spec/models/cbv_flow_invitation_spec.rb
```diff
@@ -1,79 +1,125 @@
...
+      expect(invitation.errors[:agency_id_number]).to include("can't be blank")
     end
 
-    around do |ex|
-      Timecop.freeze(now, &ex)
+    it "requires beacon_id" do
+    invitation = CbvFlowInvitation.new(valid_attributes.merge(site_id: 'ma'))
+    invitation.valid?
+    expect(invitation.errors[:beacon_id]).to include("can't be blank")
+  end
```
tdooner, line 41: Indenting looks off here

app/spec/models/cbv_flow_spec.rb
tdooner: Agreed. The Rails-y thing to do is to use the [FactoryBot gem](https://github.com/thoughtbot/factory_bot) which gives you a method (a "factory" method) to create models with default sets of fields. It's quite easy and convenient to work with. We can make a note to implement it next time we add a required field to one of these models.

tdooner (commented):
Please fix the migration so it matches `db/schema.rb`.

Also, what is the user-facing behavior like? I don't see any changes to the controllers/views, so I'm assuming it's just going to fail deep down somewhere in the CbvInvitationService with a missing field? Let's make sure there is a good error message.

tdooner (changes requested):
Actually, I'm gonna use "Request Changes" for the migration/schema comment above, so we don't inadvertently merge it in the current state.

app/spec/controllers/cbv_flow_invitations/nyc_spec.rb
```diff
@@ -29,28 +28,18 @@
...
-      }
-    end
-
     it "creates a CbvFlowInvitation record with the nyc fields" do
       post :create, params: {
         secret: invite_secret,
-        site_id: nyc_params[:site_id],
-        cbv_flow_invitation: cbv_flow_invitation_params
+        site_id: "nyc",
+        cbv_flow_invitation: attributes_for(:cbv_flow_invitation, site_id: "nyc", client_id_number: client_id_number, case_number: "ABC1234"),
```
GeorgeCodes19, line 35: nit, this could be written a bit more clear

app/spec/mailers/applicant_mailer_spec.rb
```diff
@@ -13,7 +13,7 @@
   end
 
   let(:email) { 'me@email.com' }
-  let(:cbv_flow_invitation) { CbvFlowInvitation.create(email_address: email, site_id: 'nyc') }
+  let(:cbv_flow_invitation) { create(:cbv_flow_invitation, email_address: email, site_id: "nyc", client_id_number: "ABC123") }
```
GeorgeCodes19, line 16: adding `client_id_number` because the Model validator is actually working and forcing us to pass it when the `site_id` = `nyc`

app/spec/support/Factories.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :cbv_flow_invitation, class: "CbvFlowInvitation" do
+    first_name { "Jane" }
+    middle_name { "Sue" }
+    last_name { "Doe" }
+    case_number { "ABC1234" }
+    site_id { "sandbox" }
+    email_address { "test@example.com" }
+    snap_application_date { Date.today.strftime("%m/%d/%Y") }
```
GeorgeCodes19, line 9: Defaulting to the date format that the client will submit e.g. `08/13/2024`. This will ensure that the`before_validation` hook on the `CbvFlowInvitation` model responsible for casting the string to a Date object actually works.

app/spec/models/cbv_flow_invitation_spec.rb
```diff
@@ -1,79 +1,125 @@
...
-  describe "#expired?" do
-    let(:site_id) { "sandbox" }
-    let(:invitation_valid_days) { 14 }
-    let(:invitation) do
-      CbvFlowInvitation.create!(
-        email_address: "foo@example.com",
-        site_id: site_id,
-        created_at: invitation_sent_at
-      )
+  let(:valid_attributes) do
```
GeorgeCodes19, line 4: Could do `attributes_for` here (derived from the `FactoryBot` gem)

app/spec/models/cbv_flow_spec.rb
GeorgeCodes19: I opted to use FactoryBot 👍🏾

app/app/models/cbv_flow_invitation.rb
```diff
@@ -30,4 +39,26 @@ def expired?
...
+
+  def parse_snap_application_date
+    raw_snap_application_date = @attributes["snap_application_date"]&.value_before_type_cast
+    return if raw_snap_application_date.blank? || raw_snap_application_date.is_a?(Date)
+
+    begin
+      new_date_format = Date.strptime(raw_snap_application_date.to_s, "%m/%d/%Y")
+      self.snap_application_date = new_date_format
+    rescue Date::Error => e
+      errors.add(:snap_application_date, "is not a valid date")
```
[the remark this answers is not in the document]
GeorgeCodes19: Should be `errors`. I'm going to add a test case for this as well.

app/spec/controllers/cbv_flow_invitations/nyc_spec.rb
tdooner: Using factories here defeats the point of the test: it seeks to assert that when the form fields from the previous page are POSTed to the `create` endpoint, it actually creates the object. If we use the factory to dynamically include all the proper fields, we lose the fact that the list of params we're passing in here actually matches the form on the previous page.

tdooner (approved):
LGTM, good work, thank you! Thank you especially for all the screenshots and description of how you tested this.

app/app/models/cbv_flow_invitation.rb
```diff
@@ -30,4 +39,26 @@ def expired?
...
+    raw_snap_application_date = @attributes["snap_application_date"]&.value_before_type_cast
+    return if raw_snap_application_date.blank? || raw_snap_application_date.is_a?(Date)
+
+    begin
+      new_date_format = Date.strptime(raw_snap_application_date.to_s, "%m/%d/%Y")
+      self.snap_application_date = new_date_format
+    rescue Date::Error
+      self.errors.add(:snap_application_date, "is not a valid date")
+    end
+  end
```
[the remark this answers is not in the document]
GeorgeCodes19: That's good feedback. It remember toying with this for a good 20 minutes until inspected the class with the debugger. Googling didn't really provide any decent approaches.

PR #170, opened 2024-08-14
Fix downloading PDF after completing CBV flow

tdooner:
## Ticket

Resolves FFS-1302.

## Changes

It was broken because it was mistakenly redirecting to the expired
tokens URL. This commit tweaks how we do session management like so:

1. If the user hits the "Back" button after reaching the "success" page
   (or goes to any other /cbv/ url), they will be redirected back to the
   success page, with the CbvFlow still in their session (so the page
   can render). This will create consistency so people don't
   accidentally get locked out of their session after it's complete. We
   will rely on session expiration (coming in the future) so that
   reopening the browser after a long time does not return the user to
   the success page.
2. That makes the "expired" page only used when *starting* a new flow
   from a token.
3. After those changes, we can just skip the completeness check when
   downloading the PDF to ensure that it is always possible.

## Context for reviewers

N/A

## Testing

Tested the CBV flow locally: downloading the PDF works.

### DIFF
--- a/app/app/controllers/cbv/base_controller.rb
+++ b/app/app/controllers/cbv/base_controller.rb
@@ -15,6 +15,9 @@ class Cbv::BaseController < ApplicationController
       end
 
       @cbv_flow = invitation.cbv_flow || CbvFlow.create_from_invitation(invitation)
+      if @cbv_flow.complete?
+        return redirect_to(cbv_flow_expired_invitation_path)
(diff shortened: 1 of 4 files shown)


GeorgeCodes19 (approved):

tdooner:
Acceptance tested with Eryn, she approves!

PR #176, opened 2024-08-15
George/ffs 1264 search results

GeorgeCodes19:
Resolves FFS-1264

## Changes

- Added `@pinwheel_accounts_count` view helper which counts the number of `PinwheelAccounts` associated with a particular `CbvFlow`
- Added tests to assert that when there are no search results that the copy varies depending on if the user has linked pinwheel accounts

## Testing

I tested things locally by inspecting the views and modifying the underlying database tables and `@pinwheel_accounts_count` to ensure that the copy updates as expected. 

### link redirects to summary page
Screenshot 2024-08-15 at 3 34 59 PM

### link redirects to missing results path
Screenshot 2024-08-15 at 3 35 28 PM


app/spec/controllers/cbv/employer_searches_controller_spec.rb
```diff
@@ -21,6 +22,39 @@
       end
     end
 
+    context "when there are no employer search results" do
+      context "when the user at least one pinwheel_account associated with their cbv_flow" do
+        before do
+          sign_in nyc_user
+          stub_request_items_no_items_response
```
GeorgeCodes19, line 29: I could move this to parent `context` to prevent duplication

app/app/controllers/cbv/employer_searches_controller.rb
```diff
@@ -5,6 +5,7 @@ class Cbv::EmployerSearchesController < Cbv::BaseController
   def show
     @query = search_params[:query]
     @employers = @query.blank? ? [] : fetch_employers(@query)
+    @pinwheel_accounts_count = PinwheelAccount.where(cbv_flow_id: @cbv_flow[:id]).count
```
tdooner, line 8: nit: For readability, and to simplify the view, I'd recommend something like:

```ruby
@has_pinwheel_account = @cbv_flow.pinwheel_accounts.any?
```

And then simplify the conditionals in the view so they're not duplicating the same conditional three times.

app/spec/controllers/cbv/employer_searches_controller_spec.rb
tdooner: Good idea. Also `render_views`.

app/spec/factories/pinwheel_account.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :pinwheel_account, class: "PinwheelAccount" do
+    cbv_flow_id { "123456" }
```
tdooner, line 3: Since PinwheelAccounts are associated to a CbvFlow, you can just put `cbv_flow` here and FactoryBot will automatically use the `cbv_flow` factory to create a CbvFlow whenever creating the pinwheel account in test.

```suggestion
    cbv_flow
```

It's one of the nice features of FactoryBot - that it will create dependent resources for you automatically by default so your models are actually valid and represent real data.

tdooner (approved):

PR #181, opened 2024-08-16
Update the <title> doc tag for pages to reference the first H1 on a given page. 

GeorgeCodes19:
Closes FFS-1320

## Changes

- Adds a `content_for` view helper to each page with a `#show` method.
- Adds a `title_check_helper` Test helper which verifies that the title and top H1 tag in a page match
- - Only works on GET requests that return a 200 response.
- - Does not assert requests which are redirected to other pages
- - Fails when no H1 tag is present
- - Fails when no title attribute is found for the given page (not likely)

## Testing

I test this locally in various ways. For verifying that the <title> is updated I visually confirmed that the webpage's title was successfully updated and includes the `| SNAP Income Pilot` suffix. The `title_checker_helper` also uncovered several pages without titles or improper heading hierarchies. 
Screenshot 2024-08-16 at 5 17 12 PM

Screenshot 2024-08-16 at 5 31 35 PM
Screenshot 2024-08-16 at 5 19 21 PM
Screenshot 2024-08-16 at 5 18 43 PM
Screenshot 2024-08-16 at 5 18 13 PM

### DIFF
--- a/app/app/views/layouts/application.html.erb
+++ b/app/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="<%= I18n.locale %>">
   <head>
-    <title>Verify.gov</title>
+    <title><%= content_for?(:title) ? yield(:title) : "" %> | SNAP Income Pilot</title>
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
--- a/app/spec/support/title_checker_helper.rb
+++ b/app/spec/support/title_checker_helper.rb
@@ -1,0 +1,60 @@
+module AutoTitleTestHelper
+  H1_REGEX = /<h1>(.*?)<\/h1>/im
+  TITLE_REGEX = /<title>(.*?)<\/title>/im
+  INTERPOLATION_REGEX = /%\{.*?}/
+
+  def assert_title_contains_h1(response_body)
+    h1_content = decode_html_entities(extract_content(response_body, H1_REGEX))
+    expect(h1_content).to be_present, "H1 is missing"
+
+    title_content = decode_html_entities(extract_content(response_body, TITLE_REGEX))
+    expect(title_content).to be_present, "Title is missing"
+
+    cleaned_title = remove_interpolation_placeholders(title_content)
+    main_title = cleaned_title.split('|').first.strip
+
+    expect(h1_content).to include(main_title), "H1 and title content differ: `#{h1_content}` vs `#{main_title}`"
+  end
+
+  private
+
+  def extract_content(html, regex)
+    match = html.match(regex)
+    match ? match[1].strip : nil
+  end
+
+  def remove_interpolation_placeholders(text)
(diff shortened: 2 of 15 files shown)


app/spec/support/title_checker_helper.rb
```diff
@@ -0,0 +1,52 @@
+module AutoTitleTestHelper
+  H1_REGEX = /<h1>(.*?)<\/h1>/im
+  TITLE_REGEX = /<title>(.*?)<\/title>/im
+  INTERPOLATION_REGEX = /%\{.*?}/
```
GeorgeCodes19, line 4: Some titles contain headers which themselves contain interpolated values. These values are not yet interpolated in the `<title>`, but they are interpolated in the `H1` tag. 

We remove these un-interpolated strings from the `<title>` and ensure that the `H1` entry contains the `<title>` content minus the un-interpolated values i.e. `payment_details/show.html.erb` `    <%= t(".header", employer_name: employer_name) %>
`

app/app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="<%= I18n.locale %>">
   <head>
-    <title>Verify.gov</title>
+    <title><%= content_for?(:title) ? yield(:title) : "" %> | SNAP Income Pilot</title>
```
GeorgeCodes19, line 4: I'm not sure if there should be a default value?

acouch: suggestion: Would hide the pipe in the conditional and consider the "Snap Income Pilot" as the default. "Snap Income Pilot" should also be translated.

acouch:
Fantastic. 

One other way to go would be to add `meta_title` to each view in the `en.yml` and then add a method in the application helper to retrieve it: https://stackoverflow.com/a/73371607 . That way you wouldn't need to update the views and do the regex checks. This looks great as well though.

acouch (approved):

app/spec/support/title_checker_helper.rb
tdooner: Obligatory reference to [the best stackoverflow answer ever](https://stackoverflow.com/a/1732454).

(This seems fine for now since it's just in test. In the future we will probably bring in Capybara for integration testing, and it provides [a matcher](https://www.rubydoc.info/gems/capybara/Capybara/Node/DocumentMatchers#assert_title-instance_method) to make assertions like `expect(page).to have_title("Some title")` which presumably does all the escaping.)

app/spec/support/title_checker_helper.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  config.after(:each, type: :controller) do
+    if should_check_title?(request, response)
+      assert_title_contains_h1(response.body)
+    end
+  end
+
+  def should_check_title?(request, response)
+    (request.get? || request.params[:action] == 'show') &&
+      !%w[new create].include?(request.params[:action]) &&
```
tdooner, line 47: `new` is typically a GET request that should have a `<title>` tag - if the intent is to check all GET requests, it should be included

tdooner (approved):
Perfect, thanks. And thanks for writing a clever test for it!

PR #188, opened 2024-08-19
Agency specific copy and login links for the caseworker entry page

GeorgeCodes19:
## Ticket

Resolves [FFS-1188](https://jiraent.cms.gov/browse/FFS-1188)

## Changes

- Added a new controller for the caseworkers' entry pages `entries_controller`
- Added test to assert that agency specific copy is rendered when navigating to a specific agency entry page
- Fixed a bug whereas the application's home page had an error for a missing translation.

Screenshot 2024-08-20 at 2 22 20 PM
Screenshot 2024-08-20 at 2 22 13 PM


app/app/controllers/caseworker/entries_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  private
+
+  def current_site
+    site_config[params[:site_id]]
+  end
+
+  def agency_short_name
+    current_site.agency_short_name
+  end
```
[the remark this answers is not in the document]
GeorgeCodes19: ahh good catch! it's not being used anymore. I was previously passing it to the view, but found it easier to use a new translation entry.

app/app/views/pages/home.html.erb
```diff
@@ -1,6 +1,6 @@
-<% content_for :title, t(".header") %>
-
-<h1><%= t("shared.verify.welcome") %></h1>
+<% welcome = t("shared.verify.welcome") %>
```
GeorgeCodes19, line 1: This fixes the home page title bug.

carmenrosalop (changes requested):
The copy updates look good! The only thing that needs updating are the headers inside the process list. To align with how we implemented a previous /agreement page, the process list doesn't have headers but instead should have bolded body text like in the [Figma](https://www.figma.com/design/qzbsxwS2wsouKK1JmRtn2H/Consent-Based-Verification?node-id=1897-7643&t=HbXqWrhS8TgJWnkd-4) See images for reference.

359198150-940113b8-e9b1-4c68-8e90-eb7b36708a43
Screenshot 2024-08-19 at 12 48 36 PM

app/app/controllers/caseworker/entries_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Caseworker::EntriesController < ApplicationController
+  def index
+    @current_site = current_site
```
tdooner, line 3: This should be unnecessary if you rebase on `main` now

app/config/routes.rb
```diff
@@ -42,9 +42,12 @@
     # (remove this once people know about the new invitation URL)
     get "/invitations/new", to: redirect { |_, req| "/nyc/invitations/new?secret=#{req.params[:secret]}" }
 
-    scope "/:site_id", module: :caseworker do
-      get "/sso/", to: "sso#index", as: :new_user_session
-      resources :cbv_flow_invitations, as: :invitations, path: :invitations
+    constraints(site_id: Regexp.union(Rails.application.config.sites.site_ids)) do
+      scope "/:site_id", module: :caseworker do
```
tdooner, line 46: I think putting it on the `scope` itself is a bit more intuitive - since it's referring to the "site_id" parameter that is specified by `scope "/:site_id"` if you see what I mean? It's kind of weird to have the constraint at a higher level than the scope that defines the variable.

```suggestion
      scope "/:site_id", module: :caseworker, constraints: { site_id: Regexp.union(Rails.application.config.sites.site_ids) } do
```

tdooner (approved):

carmenrosalop (approved):
Thanks George! LGTM

PR #189, opened 2024-08-19
987: Client-facing PDF; Rename helper methods

allthesignals:
## Ticket

Addresses [987](https://jiraent.cms.gov/browse/FFS-987)

## Changes

Rename helpers

### DIFF
--- a/app/app/components/table_component.rb
+++ b/app/app/components/table_component.rb
@@ -3,8 +3,14 @@
 class TableComponent < ViewComponent::Base
   renders_one :header
   renders_many :rows, types: {
-    content: TableRowComponent,
+    content: {
+      renders: TableRowComponent,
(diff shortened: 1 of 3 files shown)


app/app/models/cbv_flow_invitation.rb
```diff
@@ -47,6 +48,10 @@ def to_url
     Rails.application.routes.url_helpers.cbv_flow_entry_url(token: auth_token)
   end
 
+  def paystubs_query_begins_at
+    PAYSTUB_REPORT_RANGE.before(snap_application_date)
```
acouch, line 52: Do we need some kind of fatal error here? There are no constraints that enforce that `snap_application_date` exists for the current `cbv_flow`

PR #196, opened 2024-08-20
Implement remaining product metrics

tdooner:
## Ticket

Resolves FFS-705.

## Changes

- **Track NewRelic event for CaseworkerLogin**
- **Add `site_id` to all NewRelic events**
- **Add `user_id` to ApplicantInvitedToFlow NewRelic events**
- **Add more metadata when tracking PDF share event**
- **Track seconds between invitation sent, flow start, and completion**

## Context for reviewers

When we initially implemented NewRelic, it was too early and we didn't have all
the constructs necessary to instrument these product metrics. Now, we do, so
I've added metadata to various parts in order to capture all the metrics in the
attached ticket.

## Testing

Now that the NewRelic account is set up, the easiest way to validate this change
is to merge this, build the dashboards, and then accept the ticket based off of
the dashboards as data flows in.

### DIFF
--- a/app/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/app/controllers/users/omniauth_callbacks_controller.rb
@@ -1,4 +1,6 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  after_action :track_event
+
   def nyc_dss
     response_params = request.env["omniauth.auth"]["info"]
     Rails.logger.info "Login successful from #{response_params["email"]} (name: #{response_params["name"]}, nickname: #{response_params["nickname"]})"
@@ -36,5 +38,14 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     else
       flash[:alert] = "Something went wrong."
     end
+  end
+
+  def track_event
+    return unless @user&.persisted?
+
+    NewRelicEventTracker.track("CaseworkerLogin", {
+      site_id: @user.site_id,
+      user_id: @user.id
+    })
   end
 end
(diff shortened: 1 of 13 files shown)


app/spec/controllers/cbv/summaries_controller_spec.rb
```diff
@@ -3,13 +3,20 @@
...
+  let(:cbv_flow) do
+    create(:cbv_flow, case_number: "ABC1234", pinwheel_token_id: "abc-def-ghi", created_at: flow_started_seconds_ago.seconds.ago)
+  end
   let(:cbv_flow_invitation) { cbv_flow.cbv_flow_invitation }
 
   before do
     session[:cbv_flow_invitation] = cbv_flow_invitation
   end
 
+  around do |ex|
```
GeorgeCodes19, line 16: what's going on here?

app/app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,4 +1,6 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  after_action :track_event
```
GeorgeCodes19, line 2: 😎

GeorgeCodes19 (approved):
LGTM

app/spec/controllers/cbv/summaries_controller_spec.rb
tdooner: I make an assertion in the test that it calculates the number of "seconds since flow started" correctly, so freezing time is necessary so the calculation is exact (and doesn't take into account the time it took for the test to run).

PR #201, opened 2024-08-21
George/ffs 1334

GeorgeCodes19:
## Ticket

Resolves [FFS-1334](https://jiraent.cms.gov/browse/FFS-1334)

## Changes

- Added guard clause and redirect with error when a user attempts to access a pinwheel account information for an account that's not in the session. The user is redirected to the start of the flow.
- Added tests

## Context for reviewers

This PR prevents users from accessing pinwheel information for a particular `account_id` that's not currently in the user session.

## Testing
Screenshot 2024-08-21 at 5 16 47 PM


app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -17,6 +17,12 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
   def show
     account_id = params[:user][:account_id]
     @pinwheel_account = PinwheelAccount.find_by_pinwheel_account_id(account_id)
+
+    # security check - make sure the account_id is associated with the current cbv_flow_id
+    if @pinwheel_account.nil? || @pinwheel_account["cbv_flow_id"] != session[:cbv_flow_id]
+      return redirect_to(cbv_flow_entry_url, flash: { slim_alert: { message: t("cbv.error_no_access"), type: "error" } })
+    end
```
tdooner, line 24: This will work, but, the slightly easier way to do it is:

```ruby
@pinwheel_account = @cbv_flow.pinwheel_accounts.find_by(account_id: account_id)
```

Since that query is scoped to the current `@cbv_flow`, it will only return the accounts associated with `@cbv_flow`, and thus the conditional can just check whether `@pinwheel_account` is `nil`.

app/config/locales/en.yml
```diff
@@ -162,6 +162,7 @@ en:
         your_login_credentials: Your login credentials for the payroll provider account
     error_invalid_token: The invitation link used is not valid. Double check the link and try again. If you continue experiencing issues, contact your caseworker.
     error_missing_token_html: "<strong>Your session has timed out due to inactivity.</strong> To continue where you left off, please click the link you received from your benefits agency in your email. If you encounter any issues, contact your benefits agency for assistance."
+    error_no_access: You do not have access to this resource. You have been redirected to the entry page.
```
tdooner, line 165: I think something a bit less stern/more user-friendly would be good in case we accidentally show this to a user through some fault of our own:

```suggestion
    error_no_access: We weren't able to load payroll data for this account. Please click the link you received from your benefits agency to try again.
```

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -158,6 +158,15 @@
         expect(response.body).not_to include("Empty deduction")
       end
     end
+
+    context "when a user attempts to access pinwheel account information not in the current session" do
+      it "redirects to the entry page" do
+        get :show, params: { user: { account_id: "1234" } }
```
tdooner, line 164: Ideally there would be a test case to verify that the logic matching up the Cbv Flow ID works properly. This one is only checking the `@pinwheel_account.nil?` part of that conditional.

tdooner (approved):

carmenrosalop (approved):
Thanks for the update! Looks good!

PR #204, opened 2024-08-21
Create new `CbvFlow` object every time an invite link is clicked

tdooner:
## Ticket

Resolves FFS-1416.

## Changes
This converts the invitation's `has_one :cbv_flow` into a `has_many`.
Every time the user clicks the tokenized link again, they will have a
new CbvFlow started. This will prevent any compromise of the tokenized
link during the time between a user linking their payroll account and
completing the flow from revealing their session's data.

## Context for reviewers

This is a launch requirement from NYC.

## Testing

I will acceptance test this with Eryn by walking through various scenarios of
linking Pinwheel accounts, letting the session expire, and then clicking the
link again to verify that the associated payroll data is gone.

### DIFF
--- a/app/app/controllers/cbv/base_controller.rb
+++ b/app/app/controllers/cbv/base_controller.rb
@@ -13,12 +13,11 @@ class Cbv::BaseController < ApplicationController
       if invitation.expired?
         return redirect_to(cbv_flow_expired_invitation_path)
       end
-
-      @cbv_flow = invitation.cbv_flow || CbvFlow.create_from_invitation(invitation)
-      if @cbv_flow.complete?
+      if invitation.complete?
         return redirect_to(cbv_flow_expired_invitation_path)
       end
 
+      @cbv_flow = CbvFlow.create_from_invitation(invitation)
       session[:cbv_flow_id] = @cbv_flow.id
       NewRelicEventTracker.track("ClickedCBVInvitationLink", {
         timestamp: Time.now.to_i,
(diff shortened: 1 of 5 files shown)


app/app/controllers/cbv/base_controller.rb
```diff
@@ -13,25 +13,17 @@ def set_cbv_flow
...
-
-      @cbv_flow = invitation.cbv_flow || CbvFlow.create_from_invitation(invitation)
-      if @cbv_flow.complete?
+      if invitation.complete?
         return redirect_to(cbv_flow_expired_invitation_path)
       end
 
+      @cbv_flow = CbvFlow.create_from_invitation(invitation)
       session[:cbv_flow_id] = @cbv_flow.id
       NewRelicEventTracker.track("ClickedCBVInvitationLink", {
```
GeorgeCodes19, line 22: Unrelated, but is the assumption to query the associated `cbv_flow_id` for the `site_id`? That might be annoying to do in NR. We may want to include the `site_id` here.

app/spec/controllers/cbv/entries_controller_spec.rb
```diff
@@ -33,54 +33,18 @@
       end
 
       context "when returning to an already-visited flow invitation" do
-        let(:existing_cbv_flow) { create(:cbv_flow, cbv_flow_invitation: invitation) }
+        let!(:existing_cbv_flow) { create(:cbv_flow, cbv_flow_invitation: invitation) }
```
GeorgeCodes19, line 36: Interesting... so vars declared with `let` in an rspec test aren't hoisted to the top of the test? I can see why it would be good convention to use a `!` here

GeorgeCodes19 (approved):

app/spec/controllers/cbv/entries_controller_spec.rb
tdooner: `let!` evaluates immediately, whereas `let` is lazily evaluated the first time it's accessed. I had to convert this to a `let!` because I'm not explicitly referring to the variable anymore in the tests, so it would never be lazily evaluated, and thus the test would actually not be testing what I had intended to test.

tdooner:
Acceptance tested with Eryn ✅

PR #191, opened 2024-08-20
Removing the deprecated `cbv/share` route

GeorgeCodes19:
## Ticket

Resolves FFS-1304

## Changes

- Removed the `cbv/share#show` route
- Removed tests pertaining to the `cbv/share#show route`
- Moved application transmission method to the Summaries update method. 

## Context for reviewers

The `cbv/share#show` route is now obsolete since we've moved the share with caseworker button to the summaries page.

PR #205, opened 2024-08-21
Implement designs for /cbv/success page

tdooner:
## Ticket

Resolves FFS-1237.

## Changes

Adds a redesigned /cbv/success page.

## Context for reviewers

N/A

## Testing

Tested locally.

### DIFF
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -3,9 +3,11 @@ name: CI Documentation Checks
 on:
   push:
     branches: [ main ]
-    paths: ['app/**']
+    paths:
+      - '**/*.md'
   pull_request:
-    paths: ['app/**']
+    paths:
+      - '**/*.md'
 
 jobs:
   lint-markdown:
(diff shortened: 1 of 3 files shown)


tdooner:
@carmenrosalop Here is the implemented design for the /cbv/success page.
image

The header length is kind of long, but I'm not sure what to do about it. Whatever we do, it should probably be across-the-board. If you have thoughts on that, I could work it into my other PR with the across-the-board style changes. Should I just make the headers max out at 8-out-of-12 columns wide or something?

carmenrosalop (changes requested):
Thanks Tom! You're right that the length of the H1 is too long. I'm wondering if updating the "measure" can help with making the length of the H1 match the updated length of the body https://designsystem.digital.gov/design-tokens/typesetting/measure/
Also, you might have already done this, but we need to remove this banner when users get to this success page since they already are getting the confirmation message in the H1. 
Screenshot 2024-08-21 at 6 04 07 PM

GeorgeCodes19 (approved):

tdooner:
@carmenrosalop Good news - the message was removed by George in #191.

carmenrosalop (approved):
Approved!

PR #210, opened 2024-08-22
Updated NYC Agreement to include Nava. Separate agency agreements.

GeorgeCodes19:
## Ticket

Resolves [FFS-1374](https://jiraent.cms.gov/browse/FFS-1374)

## Changes

- Separate agreements for state agencies
- Tests to assert that the proper agreement terminology is shown depending on the `site_id`

#### NYC
Screenshot 2024-08-22 at 2 19 12 PM

#### MA
Screenshot 2024-08-23 at 11 39 42 AM


app/config/locales/en.yml
```diff
@@ -107,7 +107,10 @@ en:
       create:
         error: You must check the agreement checkbox to proceed.
       show:
-        checkbox: Check this box to agree to let Verify.gov access your payment information and share it with the %{agency_name}.
+        checkbox:
+          ma: Check this box to agree to let Verify.gov access your payment information and share it with the Department of Transitional Assistance.
```
tdooner, line 111: Who's Verify.gov? Is that like Nava? 🙃

tdooner (changes requested):

app/config/locales/en.yml
tdooner: I see - the ticket only refers to NYC. So this is fine. I think it's still good to change this to Nava since we will probably want to do that at some point.

tdooner (approved):

PR #219, opened 2024-08-23
987: Adds Caseworker PDF view

allthesignals:
## Ticket

Resolves [987](https://jiraent.cms.gov/browse/FFS-987)

## Changes

Adds the additional caseworker-specific fields to the PDF view

## Context for reviewers

Request a new data request, identity. Could this be a failed job, as well? Might need an `errored_at`. 

## Testing

Automated and manual testing.


app/app/mailers/caseworker_mailer.rb
```diff
@@ -23,6 +23,9 @@ def set_params
     @payments = params[:payments] if params[:payments]
     @employments = params[:employments]
     @incomes = params[:incomes]
+    @identity = params[:identity]
+    # shows caseworker-specific data
+    @is_caseworker = true
```
tdooner, line 28: The other way to do this would be to change the `render_to_string` call to also have `locals: { is_caseworker: true }` and then use it without the `@` in the view. We could even update the other `render_to_string` call in the applicant-facing flow to pass in `is_caseworker: false` just to be explicit.

app/app/controllers/concerns/cbv/reports_helper.rb
```diff
@@ -64,6 +69,17 @@ def fetch_incomes_for_account_id(account_id)
     pinwheel.fetch_income_metadata(account_id: account_id)["data"]
   end
 
+  def fetch_identities
+    fetch_end_user_account_ids.map do |account_id|
+      next [] unless does_pinwheel_account_support_job?(account_id, "income")
```
tdooner, line 74:
```suggestion
      next [] unless does_pinwheel_account_support_job?(account_id, "identity")
```

app/app/mailers/caseworker_mailer.rb
tdooner: Well you don't need the variable anymore, now.

app/app/models/cbv_flow.rb
```diff
@@ -23,4 +23,7 @@ def self.create_from_invitation(cbv_flow_invitation)
       site_id: cbv_flow_invitation.site_id
     )
   end
+
+  def payroll_provider
+  end
```
tdooner, line 28: What is this? Looks like this was accidentally included?

app/app/views/cbv/summaries/show.pdf.erb
```diff
@@ -1,16 +1,56 @@
 <h1><%= t(".pdf.client.header") %></h1>
 
-<p><%= t(".pdf.client.description") %></p>
+<% if is_caseworker %>
+  <p><%= is_caseworker ? t(".pdf.caseworker.description") : t(".pdf.client.description") %></p>
```
tdooner, line 4: No need for this conditional since it's already nested within the same conditional.

```suggestion
  <p><%= t(".pdf.caseworker.description") %></p>
```

tdooner (approved):
Looks good once you address the comments.

PR #222, opened 2024-08-23
Fix `employer_name` display on summary page/PDF

tdooner:
## Ticket

Resolves FFS-1439.

## Changes

On the payment details page, we were rendering the `employer_name` based
off the `employments` API endpoint, which seems to contain data more
often than the individual paystub does.

This commit updates the summary page to also pull the employer name from
the employments hash.

Along the way, it fixes a bug that prevented rendering the PDF for my
Nava data due to a missing `employer_phone_number` hash.

## Context for reviewers

N/A

## Testing

Tested with my Nava data. (Redacted) screenshot below.


tdooner:
image

image

tdooner:
@carmenrosalop Screenshots above. I took the liberty of adding a dash between the employer name and "Employment Information" and "Pay Date" in the PDF only. Let me know if you don't like that. (I found it a bit hard to read when it was all together.)

carmenrosalop (commented):
The report is looking snazzy! I agree with your addition of the dash between the employer name and "employment info" and "pay date". I'll add a dash on the PDF design.
One small thing that I see is that ideally the text hierarchy for these text should be 
- Income verification report (H1)
- Employment and payment details (H2)
- Employer 1: Wegmans (H3)
- Wegmans - Employment information (H4)
Right now, Employer 1 seems smaller than the row below H4
Screenshot 2024-08-23 at 5 14 09 PM

PR #224, opened 2024-08-23
Return to caseworker entry page after logout

tdooner:
## Ticket

Resolves FFS-1440.

## Changes

We had been returning to the homepage, but we want to go back to the
entry page since there's not really a way to log in or do anything from
the homepage.

## Context for reviewers

N/A

## Testing

Will attach a recording of the updated behavior.

### DIFF
--- a/app/app/controllers/application_controller.rb
+++ b/app/app/controllers/application_controller.rb
@@ -11,6 +11,10 @@ class ApplicationController < ActionController::Base
     caseworker_dashboard_path(site_id: user.site_id)
   end
 
+  def after_sign_out_path_for(resource_or_scope)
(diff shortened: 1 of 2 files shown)


tdooner:
@carmenrosalop An easy one for acceptance:

https://github.com/user-attachments/assets/dc0d8b04-72ae-4e2e-bc78-00c8853cf3a6

carmenrosalop (approved):
Looks great! @tdooner 
Not part of this ticket but wanted to point out for a future ticket that we may want to update this banner to slim and to say "You have been successfully signed out."
Screenshot 2024-08-23 at 4 54 11 PM

PR #225, opened 2024-08-23
Implement /expired page

tdooner:
## Ticket

Resolves FFS-1437.

## Changes

The designs weren't ready when I initially implemented link expiry.
Let's implement them now.

Since the `@cbv_flow` is not available, we pass the site_id as a param
to the expiration controller.

## Context for reviewers

N/A

## Testing

Tested locally that an expired link takes us to the right version of the page.


tdooner:
@carmenrosalop 
image

carmenrosalop (approved):
Looks great!

PR #211, opened 2024-08-22
Update missed opportunities for factories

allthesignals:
## Ticket

N/A

## Changes

Uses FactorBot in more spec files; updates a method name

PR #228, opened 2024-08-26
Update to simplify factory spec creation

acouch:
## Ticket

I noticed some improvements to simplify the work started in: #211


app/spec/controllers/cbv/add_jobs_controller_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe Cbv::AddJobsController do
-  let(:cbv_flow) { create(:cbv_flow, case_number: "ABC1234", site_id: "sandbox") }
+  let(:cbv_flow) { create(:cbv_flow) }
```
acouch, line 4: The [factory](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/spec/factories/cbv_flow.rb#L3) already has `case_number` and `site_is` so these can be removed.

PR #233, opened 2024-08-26
987: PDF header and show pagination for caseworker email

allthesignals:
## Changes

Displays header for PDF; Shows pagination for mailer as well

image

### DIFF
--- a/app/app/views/layouts/pdf.pdf.erb
+++ b/app/app/views/layouts/pdf.pdf.erb
@@ -11,16 +11,6 @@
 
   <body>
     <div id="root">
-      <div>
-        <span class="cbv-header__pilot-name">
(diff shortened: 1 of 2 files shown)


app/app/views/layouts/pdf.pdf.erb
```diff
@@ -11,6 +11,16 @@
 
   <body>
     <div id="root">
+      <div>
+        <span class="cbv-header__pilot-name">
+          <%= t("shared.pilot_name") %>
+        </span>
+        <% if current_site %>
```
tdooner, line 18: I should hope this will be set!

tdooner (approved):
A screenshot would be nice

PR #234, opened 2024-08-27
George/ffs 1026 ma batch file upload

GeorgeCodes19:
## Ticket

Resolves FFS-1026

## Changes

- [x] Create an S3 file upload service
- [x] Provide a means to encrypt file contents
- [x] Create a service to convert views into a PDF
- [x] Add logic to zip the CSV and PDF file together
- [x] Test the S3 service using Rails console

## Considerations

- CSV and PDF generation result are only in-memory
- TAR and GPG files use the Ruby [Tempfile](https://ruby-doc.org/3.3.5/stdlibs/tempfile/Tempfile.html) to write files to disk. Tempfiles are automatically cleaned up by the garbage collector. 

## Testing

In an attempt to automate much of the user workflow I've added functionality to generate a GPG key pair rather than storing the key/pair in the codebase. We've verified that the CSV and PDF contents match what MA is expecting.


app/app/services/s3_service.rb
```diff
@@ -0,0 +1,51 @@
+# @see: https://github.com/ueno/ruby-gpgme
+require "gpgme"
+require "aws-sdk-s3"
+
+class S3Service
```
tdooner, line 5: By lines of code, this is more of a GPG Service than an S3 service. Maybe it should be named something that makes it clear it's actually responsible for both. Or, we could split out a GpgEncryptionService and keep this file cleanly responsible for uploading files to S3.

app/app/services/s3_service.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  private
+
+  def import_key(key_data)
+    unless GPGME::Key.valid?(key_data)
+      raise "Invalid GPG key: The provided key data is not valid"
+    end
+    GPGME::Key.import(key_data)
+  end
```
tdooner, line 50: Is this idempotent? Since we're calling it every time, is it going to keep adding the key repeatedly?

app/app/services/s3_service.rb
```diff
@@ -0,0 +1,51 @@
...
+
+
+  def gpg_encrypt_file(file_path)
+    import_key(@public_key)
+    crypto = GPGME::Crypto.new
+    recipient = GPGME::Key.find(:public, @public_key).first
+    encrypted_file_path = "#{file_path}.gpg"
+    File.open(encrypted_file_path, "wb") do |output|
+      crypto.encrypt File.open(file_path, "rb"), recipients: recipient, output: output
```
tdooner, line 26: This `File.open` should also be in a block, like the `output` file is, so that the file descriptor is closed after it's done being encrypted.

app/app/controllers/cbv/summaries_controller.rb
```diff
@@ -71,6 +71,33 @@ def transmit_to_caseworker
...
+          has_consent: has_consent
+        }
+      )
+      csv_path = "#{Rails.root}/tmp/cbv_flow_#{current_site.id}_#{Time.now.to_i}.csv"
+      File.open(csv_path, "w") do |file|
+        file.write("Employer,Pay Date,Gross Pay,Net Pay\n")
+        @payments.each do |payment|
+          file.write("#{payment[:employer_name]},#{payment[:pay_date]},#{payment[:gross_pay_amount]},#{payment[:net_pay_amount]}\n")
+        end
+      end
```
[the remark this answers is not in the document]
GeorgeCodes19: Yes, sorry I had unpublished PR comments myself. That was one of them. I'm doing TDD and just haven't gotten here yet. I kept note of what we promised to deliver. This is just a placeholder.

app/app/services/s3_service.rb
GeorgeCodes19: Agreed. I was thinking there would be a separate module for handling encryption and we'd use a bit of composition.

app/spec/spec_helper.rb
```diff
@@ -93,4 +99,13 @@
...
   Kernel.srand config.seed
+
+  # Mock the pinwheel method if it's not available in the test environment
+  config.before(:each, type: [ :controller, :request, :service ]) do
+    def pinwheel
+      @pinwheel ||= double('pinwheel')
+    end
+  end
+
+  config.include Cbv::PaymentsHelper if defined?(Cbv::PaymentsHelper)
```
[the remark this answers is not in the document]
GeorgeCodes19: This may not be needed. I'm not sure what happened to my review, but this was something else that I commented on. It shouldn't be needed anymore. The main reason for including it was because I was running into issues with the `PaymentsHelper` being invoked by the test and not having instance vars set. There was some weird code that I tossed out and this is residual stuff to toss out.

app/spec/services/pdf_service_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        template: 'cbv/summaries/show',
+        variables: variables
+      )
+    end
+
+    context "#generate" do
+      it 'generates a PDF file' do
+        expect(@file_path).not_to be_nil, "PDF generation failed"
+
+        if @file_path != nil
```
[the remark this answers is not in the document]
tdooner: If rubocop is wrong, you could use a comment to disable it:
https://docs.rubocop.org/rubocop/configuration.html#disabling-cops-within-source-code

app/spec/services/s3_service_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+  let(:s3_client) { instance_double(Aws::S3::Client) }
+
+  before(:all) do
+    # Ensure that generated gpg keys are within the rails tmp dir
+    @original_gpg_home = ENV['GNUPGHOME']
+    ENV['GNUPGHOME'] = Rails.root.join('tmp', 'gpghome').to_s
+    FileUtils.mkdir_p(ENV['GNUPGHOME'])
+
+    # Generate a new key pair non-interactively
+    key_script = <<-SCRIPT
```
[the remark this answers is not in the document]
tdooner: Seems fine to me. The slightly easier way would be to just generate one yourself and check it into the repo's spec/fixtures folder.

app/app/services/pdf_service.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def self.get_html
+    @html_content
+  end
+
+  def self.get_pdf
+    @pdf_content
+  end
+end
```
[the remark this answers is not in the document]
GeorgeCodes19: Thanks! This is exactly what I was trying to accomplish.

app/spec/controllers/concerns/gpg_encryption_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  let(:tmp_directory) { Rails.root.join('tmp') }
+  let(:test_file_content) { "This is a test file content" }
+  let(:test_file_path) { tmp_directory.join('test_file.txt').to_s }
+  let(:tar_file_path) { tmp_directory.join('test_file.tar').to_s }
+  let(:encrypted_tar_file_path) { "#{tar_file_path}.gpg" }
+  let(:untar_file_path) { File.join(tmp_directory, 'untarred') }
+
+  let(:encrypted_file_path) { "#{test_file_path}.gpg" }
+
+  before(:all) do
```
GeorgeCodes19, line 17: Violating some DRY principals here. How can I abstract this into a test helper so that we can call it whenever needed?

app/app/controllers/cbv/summaries_controller.rb
```diff
@@ -70,12 +74,94 @@ def transmit_to_caseworker
...
+        @cbv_flow.cbv_flow_invitation.middle_name,
+        @cbv_flow.cbv_flow_invitation.snap_application_date,
+        @cbv_flow.cbv_flow_invitation.beacon_id,
+        pinwheel_account.created_at,
+        @cbv_flow.created_at,
+        @cbv_flow.updated_at,
+        @cbv_flow.confirmation_code,
+        @cbv_flow.consented_to_authorized_use_at,
+        "application/pdf",
+        File.size(pdf_output["path"]),
```
GeorgeCodes19, line 159: I'll move this back into the pdf service

app/app/controllers/application_controller.rb
```diff
@@ -29,6 +30,8 @@ def site_config
 
   def current_site
     @current_site ||= site_config[params[:site_id]]
+    ApplicationHelper.current_site = @current_site
```
GeorgeCodes19, line 33: The `ApplicationHelper` receives the instance variable @current_site. Any code relying on `ApplicationHelper` can then retrieve the `ApplicationHelper`

app/app/controllers/cbv/base_controller.rb
```diff
@@ -1,5 +1,5 @@
 class Cbv::BaseController < ApplicationController
-  before_action :set_cbv_flow, :ensure_cbv_flow_not_yet_complete
+  before_action :set_cbv_flow, :ensure_cbv_flow_not_yet_complete, :current_site
```
GeorgeCodes19, line 2: Setting this as a `before_action` ensures that `@current_site` is set for `ApplicationHelper`

app/config/site-config.yml
```diff
@@ -25,8 +25,8 @@
   agency_contact_website: https://www.mass.gov/guides/how-to-contact-dta
   caseworker_feedback_form: https://forms.office.com/r/rfV04qjG9A
   pinwheel:
-    api_token: <%= ENV['PINWHEEL_API_TOKEN_DEVELOPMENT'] %>
-    environment: development
+    api_token: <%= ENV['PINWHEEL_API_TOKEN_SANDBOX'] %>
+    environment: sandbox
```
tdooner, line 29: Looks like this was committed by accident

app/app/controllers/concerns/gpg_encryptable.rb
```diff
@@ -0,0 +1,42 @@
+require "gpgme"
+require "tempfile"
+
+module GpgEncryptable
+  extend ActiveSupport::Concern
+
+  def gpg_encrypt_file(file_path, public_key)
+    crypto = GPGME::Crypto.new
+    begin
+      key_fingerprint = imported_key_fingerprint(public_key)
```
GeorgeCodes19, line 10: the public key needs to be imported to the GPG keyring prior to using. So we import it and then use the fingerprint to resolve the recipient key. Tempfile(s) are created ahead of time and then have data written to them.

app/app/controllers/concerns/tar_file_creatable.rb
```diff
@@ -0,0 +1,64 @@
+require "zlib"
+require "rubygems/package"
+require "tempfile"
+
+module TarFileCreatable
+  def create_tar_file(file_data)
+    temp_file = Tempfile.new([ SecureRandom.uuid, ".tar" ])
+    temp_file.binmode
+
+    Gem::Package::TarWriter.new(temp_file) do |tar|
```
GeorgeCodes19, line 10: I probably could get away with just using `TarWriter` instead of the fully qualified class name.

app/app/helpers/csv_helper.rb
```diff
@@ -0,0 +1,9 @@
+module CsvHelper
+  def create_csv(data)
```
GeorgeCodes19, line 2: This is only creating one header and row. I think @acouch might've done something more extensive around creating CSVs.

app/app/controllers/concerns/tar_file_creatable.rb
tdooner: I doubt it

app/db/schema.rb
```diff
@@ -51,8 +51,8 @@
     t.string "site_id"
     t.string "confirmation_code"
     t.datetime "transmitted_at"
-    t.datetime "redacted_at"
     t.datetime "consented_to_authorized_use_at"
+    t.datetime "redacted_at"
```
tdooner, line 55: In general we don't want to commit changes like this. It's harmless, because it just reflects a difference in column order between your dev machine and the schema file, but we don't want to commit it because someone else is just going to end up committing the opposite change next time their schema gets dumped to this file. So it's churn for no reason.

If there are diffs like this, it just means that at some point you should recreate your database with `bin/rails db:reset` so it conforms to this schema.

app/spec/spec_helper.rb
```diff
@@ -93,4 +93,11 @@
   # test failures related to randomization by passing the same `--seed` value
   # as the one that triggered the failure.
   Kernel.srand config.seed
+
+  # Mock the pinwheel method if it's not available in the test environment
+  config.before(:each, type: [ :controller, :request, :service ]) do
+    def pinwheel
+      @pinwheel ||= double('pinwheel')
+    end
+  end
```
tdooner, line 102: Is this necessary? It looks like the way you stub Pinwheel in tests doesn't rely on this, so I'm guessing this is residual from a previous attempt at stubbing pinwheel

tdooner (approved):
Looks good, this is mergeable as far as I see. Let's wait to get this bugfix release out first and then go ahead.

PR #235, opened 2024-08-27
987: Filter for known account IDs only

allthesignals:
## Ticket

N/A

## Changes

`fetch_end_user_account_ids` was returning any account ID that pinwheel knew about. 

Renamed it to `fetch_known_end_user_account_ids` and filter for those IDs in our database.

## Context for reviewers

There's a minor scenario where an account may be registered with Pinwheel but not with our app. This can lead to errors later in the flow.

## Testing

Manual tested and current automated test works.

### DIFF
--- a/app/app/controllers/concerns/cbv/payments_helper.rb
+++ b/app/app/controllers/concerns/cbv/payments_helper.rb
@@ -39,8 +39,8 @@ module Cbv::PaymentsHelper
   end
 
   def fetch_known_end_user_account_ids
-    pinwheel_account_ids = pinwheel.fetch_accounts(end_user_id: @cbv_flow.pinwheel_end_user_id)["data"].map { |account| account["id"] }
+    pinwheel_account_ids = pinwheel.fetch_accounts(end_user_id: @cbv_flow.pinwheel_end_user_id)["data"].pluck("id")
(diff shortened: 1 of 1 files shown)


app/app/controllers/concerns/cbv/payments_helper.rb
```diff
@@ -38,7 +38,9 @@ def parse_payments(payments)
     end
   end
 
-  def fetch_end_user_account_ids
-    pinwheel.fetch_accounts(end_user_id: @cbv_flow.pinwheel_end_user_id)["data"].map { |account| account["id"] }
+  def fetch_known_end_user_account_ids
```
GeorgeCodes19, line 41: Would `.pluck` make more sense here since we don't need to fetch the entire column, just the `id`?

```suggestion
def fetch_known_end_user_account_ids
  pinwheel_account_ids = pinwheel.fetch_accounts(end_user_id: @cbv_flow.pinwheel_end_user_id)["data"].pluck("id")
  PinwheelAccount.where(pinwheel_account_id: pinwheel_account_ids).pluck(:pinwheel_account_id)
end
```

GeorgeCodes19 (approved):

PR #237, opened 2024-08-27
Fix "submitted date" on caseworker email

tdooner:
## Ticket

N/A - Bug report in slack

## Changes

We used `transmitted_at` by accident, but that won't be set until the
email succeeds at sending. Let's use `consented_to_authorized_use_at`
instead, which represents the time we should initially consider the CBV
flow "complete" and that it is ready to transmit.

## Context for reviewers

N/A

## Testing

Screenshot of the mailer preview below.


tdooner:
image

PR #238, opened 2024-08-28
Add terraform to subscribe to email delivery/failure events

tdooner:
## Ticket

N/A - Relates to NYC debugging.

## Changes

This adds terraform code to get us most of the way towards processing
the delivery status of every email: it sets up delivery event
notifications for any successfully delivered email, or any email that is
not deliverable due to bounce.

The remaining work is to actually subscribe the application to receive
these events as HTTP webhooks, either directly via SNS or by creating a
Lambda to send the events. Until that work is done, we can get
visibility into email deliverability (for debugging purposes) by
subscribing our personal email addresses to it. (Again, only for
debugging purposes until we have actual real data.)

## Context for reviewers

N/A

## Testing

Tested by running this locally and creating all the resources in our demo environment.


tdooner:
It turns out this was unnecessary for NYC debugging, but we probably should still do it in tandem with handling these webhooks in Rails to track deliverability for product metrics reasons.

tdooner:
@allthesignals Yeah I think merging it is fine, I don't think it's too risky. We would still have to write the actual handler to subscribe to receive these events.

GeorgeCodes19 (approved):

PR #243, opened 2024-08-28
1492: Check Sync Status On Reconnect

allthesignals:
## Ticket

https://jiraent.cms.gov/browse/FFS-1492

## Changes

Previously, dropped connections would leave users stranded on a synchronization page. Further, if a user missed an update due to poor connection, they wouldn't receive an update on reconnect.

This update splits out the account synchronization step into an interstitial route which is responsible for:
1. Checking sync status on route load in case user accidentally refreshes the page or loses connection
2. Subscribing to websockets and listening for periodic updates on sync status

## Testing

I'm unsure how best to test this ActionCable behavior but it's been manually tested...

Screenshot 2024-09-30 at 3 56 45 PM
Screenshot 2024-09-30 at 3 56 48 PM
Screenshot 2024-09-30 at 3 56 52 PM
Screenshot 2024-09-30 at 3 56 57 PM

https://github.com/user-attachments/assets/d356d8ea-82e4-40b1-9d0b-8b9633407e73


app/app/controllers/cbv/synchronizations_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Cbv::SynchronizationsController < Cbv::BaseController
+  def show
+    account_id = params[:user][:account_id]
+
+    @pinwheel_account = PinwheelAccount.find_by_pinwheel_account_id(account_id)
```
tdooner, line 5: We need make sure the current user has permission to view this PinwheelAccount.

```suggestion
    @pinwheel_account = @cbv_flow.pinwheel_accounts.find_by(pinwheel_account_id: account_id)
```

^ should do the trick

tdooner (approved):
I pulled and tried locally. It works pretty well for me, nicely done. To allow us to do NYC hotfixes, let's hold off on merging this until you've incorporated the design.

GeorgeCodes19 (approved):

app/app/views/cbv/synchronizations/show.html.erb
GeorgeCodes19, line 1: This modal is a good candidate for some of that server rendered component work that we were talking about when addressing tech debt.

app/app/models/pinwheel_account.rb
```diff
@@ -23,10 +23,29 @@ def has_fully_synced?
...
     return nil unless error_column.present?
 
     supported_jobs.include?(job) && send(sync_column).present? && send(error_column).blank?
   end
+
+  def job_completed?(job)
+    error_column, sync_column = event_columns_for(job)
+    return nil unless error_column.present?
+
+    supported_jobs.include?(job) && (send(sync_column).present? || send(error_column).present?)
```
tdooner, line 36: I think we always set the `sync_column` timestamp regardless of whether it was successful, so you could probably only check the `sync_column` here. But, probably good to be careful and may as well check both.

app/app/models/pinwheel_account.rb
```diff
@@ -23,10 +23,29 @@ def has_fully_synced?
...
+  end
+
+  def relevant_cbv_jobs
+    supported_jobs.filter { |job| EVENTS_MAP.keys.any? { |key| key.start_with? job } }
+  end
```
tdooner, line 41: I don't see where this method is being used?

PR #247, opened 2024-08-29
Fix /:site/invitations to be a 404 instead of 500

tdooner:
## Ticket

Resolves FFS-1350.

## Changes

This error has happened a couple times. I couldn't reproduce why, so I'm
assuming that NYC staff just forgot the "/new" at the end of the link
and typed it in manually. To handle this, let's make it be a 404 rather
than 500 error.

While testing, I also cleaned up the parameters used on these redirects -
we don't have an invite secret anymore.

## Context for reviewers

N/A

## Testing

Tested invitation sending locally - still works.

### DIFF
--- a/app/config/routes.rb
+++ b/app/config/routes.rb
@@ -43,7 +43,7 @@ Rails.application.routes.draw do
       root to: "entries#index", as: :new_user_session
 
       resource :dashboard, only: %i[show], as: :caseworker_dashboard
-      resources :cbv_flow_invitations, as: :invitations, path: :invitations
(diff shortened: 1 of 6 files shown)


app/config/routes.rb
```diff
@@ -43,7 +43,7 @@
       root to: "entries#index", as: :new_user_session
 
       resource :dashboard, only: %i[show], as: :caseworker_dashboard
-      resources :cbv_flow_invitations, as: :invitations, path: :invitations
+      resources :cbv_flow_invitations, only: %i[new create], as: :invitations, path: :invitations
```
tdooner, line 46: ^ This is the actual fix. The rest is just related to the removal of the `secret` param

PR #248, opened 2024-08-30
Only permit authorized MA users

allthesignals:
## Ticket

Resolves [1036](https://jiraent.cms.gov/browse/FFS-1036)

## Changes

Only permitted certain emails to continue through authorization. 

## Context for reviewers

This was tricky to test. Because SiteConfig are loaded on application launch, rspec doesn't reload everything when stubbing the environment variable for the emails. The only way to get this to test properly was to define my own `site_config` or `sites` in the omniauth controller, which is reloaded every for every test (and therefore picks up my stub).

## Testing

Automated and manual testing

### DIFF
--- a/.grype.yml
+++ b/.grype.yml
@@ -20,3 +20,4 @@ ignore:
   - vulnerability: GHSA-c2qf-rxjj-qqgw
   - vulnerability: GHSA-cx63-2mw6-8hw5
   - vulnerability: GHSA-qwmp-2cf2-g9g6
+  - vulnerability: CVE-2024-4603
--- a/.trivyignore
+++ b/.trivyignore
@@ -7,3 +7,6 @@
 #  Issue:         Why there is a finding and why this is here or not been removed
 #  Last checked:  Date last checked in scans
 #The-CVE-or-vuln-id # Remove comment at start of line
+CVE-2024-2511
+CVE-2024-4603
+CVE-2024-4741
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -108,10 +108,10 @@ RUN apt-get update -qq && \
     apt-get install -y --no-install-recommends unzip=6.0-28 \
         python3-venv=3.11.2-1+b1 \
         python-is-python3=3.11.2-1+deb12u1 \
-        curl=7.88.1-10+deb12u6 \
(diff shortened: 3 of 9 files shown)


app/lib/site_config.rb
```diff
@@ -40,6 +41,7 @@ def initialize(yaml)
       @agency_name = yaml["agency_name"]
       @agency_short_name = yaml["agency_short_name"]
       @agency_contact_website = yaml["agency_contact_website"]
+      @authorized_emails = yaml["authorized_emails"] || ""
```
tdooner, line 44: Should we split this into an array or something? `(yaml["authorized_emails"] || "").split(",")` or similar?

app/spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -16,64 +16,78 @@
...
-      new_user = User.last
-      expect(new_user).to have_attributes(
-        email: test_email,
-        site_id: "ma"
-      )
-      expect(controller.current_user).to eq(new_user)
-    end
+    context "when the user is authorized" do
+      around do |ex|
+        stub_environment_variable("MA_DTA_ALLOWED_CASEWORKER_EMAILS", test_email, &ex)
```
tdooner, line 21: Yeah this is hard to stub because the file is evaluated at Rails load. I think you could have done:

```ruby
allow(Rails.application.config.sites["ma"])
  .to receive(:authorized_emails).and_return(test_email)
```

app/app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -14,7 +14,12 @@ def ma_dta
     Rails.logger.info "Login successful from #{response_params["email"]} (name: #{response_params["name"]}, nickname: #{response_params["nickname"]})"
     email = response_params["email"]
 
-    login_with_oauth(email, "ma")
+    if authorized?(email, "ma")
```
tdooner, line 17: I think we may want to normalize these emails by downcasing them just to make sure there's no ambiguity in the username. Devise will do it automatically but we should probably do it before we run our own logic. I think just calling `response_params["email"].downcase` should do the trick.

PR #250, opened 2024-09-03
Update Dockerfile and ignore files for vuln scans

acouch:
## Changes

Updates [curl](https://packages.debian.org/source/bookworm/curl) from `7.88.1-10+deb12u6` to `7.88.1-10+deb12u7` as 6 is no longer supported which is [breaking the CI](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/10684116756/job/29613763006#step:3:100).


app/.docker-compose.yml.swp
tdooner, line 1: It looks like this was accidentally included

app/local.env.example
tdooner, line 1: What is this file supposed to be an example of? If you wanted to add a `.env.development.local.example` file I could see that making sense with these environment variables as placeholders and instructions to copy it from 1password.

```
PINWHEEL_API_TOKEN_SANDBOX=
PINWHEEL_API_TOKEN_DEVELOPMENT=
PINWHEEL_API_TOKEN_PRODUCTION=

NEWRELIC_KEY=
SLACK_TEST_EMAIL=
AZURE_SANDBOX_CLIENT_ID=
AZURE_SANDBOX_CLIENT_SECRET=
AZURE_SANDBOX_TENANT_ID=
```

acouch: This was added by accident.

acouch:
This updates the Dockerfile for upgrades in installed packages and ignores the results that require upgrading the debian version which would be a bigger effort.

PR #251, opened 2024-09-03
Drop unused date-related methods

allthesignals:
## Ticket

N/A

## Changes

These two methods are no longer needed and were replaced by the application date.

## Context for reviewers

Removing to avoid misdirection

### DIFF
--- a/app/app/controllers/cbv/payment_details_controller.rb
+++ b/app/app/controllers/cbv/payment_details_controller.rb
@@ -2,8 +2,6 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
   include Cbv::PaymentsHelper
 
   helper_method :employer_name,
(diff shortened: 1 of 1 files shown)


app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -103,22 +103,6 @@ def compensation_amount
...
-  end
-
-  def end_date
-    return I18n.t("cbv.payment_details.show.unknown") unless has_paystubs_data?
-
-    @payments
-        .sort_by { |payment| payment[:end] }
-        .last[:end]
-  end
-
```
tdooner, line 121: They can be removed from the `helper_method` call above as well

tdooner (approved):

PR #253, opened 2024-09-04
Weekly Email Report

acouch:
## Ticket

Resolves FFS-1422

## Changes

Adds email with attached csv download.

## Testing


app/app/mailers/weekly_report_mailer.rb
```diff
@@ -0,0 +1,46 @@
...
+    items = CbvFlowInvitation.joins(:cbv_flows).where(cbv_flow_invitations: { 'site_id': site_id }).where(cbv_flows: { 'consented_to_authorized_use_at': previous_week_start..week_start }).select(cols)
+    rows = items.pluck(cols)
+    # The column "consented_to_authorized_use_at" should be reported as "completed_at"
+    cols[5] = "completed_at"
+    attachments[filename] = generate_csv(rows, cols)
+    mail(
+        to: email_address,
+        subject: "Weekyl report email",
+    )
+  end
```
tdooner, line 25: This method is really long - would be good to break it out into separate responsibilities. Ideally most of the logic for creating the report would be somewhere that's easier to test - e.g. in a service object.

app/app/mailers/weekly_report_mailer.rb
```diff
@@ -0,0 +1,46 @@
...
+    site_id = "nyc"
+    now = Time.now
+    previous_week_start = now.beginning_of_week - 7.days
+    week_start = now.beginning_of_week
+    cols = "client_id_number", "transmitted_at", "case_number", "created_at", "snap_application_date", "consented_to_authorized_use_at", "site_id"
+    to_date = now.beginning_of_week.strftime("%Y%m%d")
+    from_date = previous_week_start.beginning_of_week.strftime("%Y%m%d")
+    filename = "weekly_report_#{from_date}-#{to_date}.csv"
+    # Get flows from start of previous week to today
+    items = CbvFlowInvitation.joins(:cbv_flows).where(cbv_flow_invitations: { 'site_id': site_id }).where(cbv_flows: { 'consented_to_authorized_use_at': previous_week_start..week_start }).select(cols)
```
tdooner, line 16: After you started this, the requirements have changed and now we want to send *all* CBV flows regardless of completion. So it should probably use `created_at` rather than `consented_to_authorized_use_at`.

We also want to include the email address of the invitee as a column.

Also, no need to `select(cols)` if you are plucking the columns later.

app/app/mailers/weekly_report_mailer.rb
```diff
@@ -0,0 +1,46 @@
...
+    rows = items.pluck(cols)
+    # The column "consented_to_authorized_use_at" should be reported as "completed_at"
+    cols[5] = "completed_at"
```
tdooner, line 19: This seems liable to break, for example, if we added another column before this one.

app/app/mailers/weekly_report_mailer.rb
```diff
@@ -0,0 +1,46 @@
...
+    attachments[filename] = generate_csv(rows, cols)
+    mail(
+        to: email_address,
+        subject: "Weekyl report email",
+    )
+  end
+
+    private
```
tdooner, line 27: Some Ruby style conventions that would be nice to incorporate as you're finishing up this code:
* 2-space indent (lines 22 & 23)
* Use explicit array construction, not just a comma-separated list of items. You could [use %w[]](https://stackoverflow.com/questions/1274675/what-does-warray-mean) if you wanted. (line 11)
* Hash keys before a colon don't need quotes around them, they will be converted to symbols anyway (i.e. `{ 'site_id': ...` is equivalent to `{ site_id: ` and is accessed as `hash[:site_id]`) (line 16)

app/app/mailers/weekly_report_mailer.rb
acouch: Updated, thanks.

app/app/mailers/weekly_report_mailer.rb
acouch: Updated, thanks 👍🏿

app/spec/mailers/weekly_report_mailer_spec.rb
```diff
@@ -0,0 +1,57 @@
+require "rails_helper"
+require 'csv'
+
+RSpec.describe WeeklyReportMailer, type: :mailer do
+  let(:now) { Time.now.in_time_zone("America/New_York") }
+  let(:three_days_ago) { now - 5.days }
```
GeorgeCodes19, line 6: Is this correct? Wouldn't this be `:five_days_ago`?

app/app/mailers/weekly_report_mailer.rb
```diff
@@ -0,0 +1,61 @@
+require "csv"
+class WeeklyReportMailer < ApplicationMailer
+  helper_method :week_start_date_formatted, :prev_week_start_date_formatted
+
+  # Send email with a CSV file that reports on completed flows in past week
+  def report_email
+    email_address = site_config["nyc"].transmission_method_configuration["email"]
+    site_id = "nyc"
+    now = Time.now
+    previous_week_start_date = now.beginning_of_week - 7.days
```
GeorgeCodes19, line 10: This might work better
```suggestion
previous_week_start_date = now.beginning_of_week.prev_week
```

app/spec/mailers/weekly_report_mailer_spec.rb
acouch: Good catch, I updated that. Thanks.

tdooner:
Going to reopen a new PR.

PR #254, opened 2024-09-04
1146: PDFs for MA

allthesignals:
## Ticket

Resolves [1146](https://jiraent.cms.gov/browse/FFS-1146)

## Changes

Adds MA-specific fields to the MA PDF.

Adds a helper, current_site?(site_id).

[caseworker_ma.pdf](https://github.com/user-attachments/files/16880776/ma_caseworker.pdf)
[client_ma.pdf](https://github.com/user-attachments/files/16880778/client_ma.pdf)


tdooner (approved):
Looks good. I find it a bit weird that the component can conditionally decide not to render itself, but I suppose it does make sense not to render an empty table, so that's fine.

PR #256, opened 2024-09-05
Update every page's copy based on MA feedback

tdooner:
## Ticket

Resolves FFS-1455.

## Changes

- **Update ApplicationMailer#invitation_email copy**
- **Update /entry page with new copy**
- **Update /agreements page with MA copy**
- **Update /employer_searches with new copy**
- **Update /missing_results page with new copy**
- **Update /payment_details page with new copy**
- **Update /add_jobs page with new copy**
- **Update /summary page with new copy**
- **Update /success page with new copy**
- **Update /home with new copy**
- **Fix caseworker mailer "submitted on" date**

## Context for reviewers

Mostly just copy changes. I tried to minimize the number strings by passing in the `%{agency_acronym}` rather than creating a site-specific override.

## Testing

Went through the flow for MA, and it works. I will continue testing with NYC and verify with Makaela or Carmen for acceptance.


app/app/mailers/applicant_mailer.rb
```diff
@@ -6,7 +6,7 @@ class ApplicantMailer < ApplicationMailer
   def invitation_email
     mail(
       to: @cbv_flow_invitation.email_address,
-      subject: I18n.t("applicant_mailer.invitation_email.subject")
+      subject: view_context.site_translation("applicant_mailer.invitation_email.subject")
```
GeorgeCodes19, line 9: What's up with `view_context`? I did some Googling, but can't see how things are being stringed together. How is `site_translation` a method of `view_context`?

GeorgeCodes19 (approved):

app/app/mailers/applicant_mailer.rb
tdooner: This was a tricky one, the first time I've tried to do something like this. The problem here is that we want to use the `site_translation` view helper in the mailer. But view helpers are not available in the mailer/controller itself (only the view), because they are only mixed into the view context that the ERB is evaluated with.

If we were trying to use the view helper _in a controller,_ we would be able to do it [with `helpers.site_translation`](https://apidock.com/rails/ActionController/Helpers/ClassMethods/helpers), but no similar method exists for mailers. Thankfully, we can just call the method on the `view_context` object itself, which has all the helper methods mixed in already.

I also tried just doing `include ApplicationHelper` in the mailer to include the site_translation method directly, but that doesn't work either because the site_translation method calls other methods that Rails automatically adds to the view_context (the `t` method, among others).

PR #258, opened 2024-09-05
Temporarily replace MA configuration to match sandbox

tdooner:
## Ticket

N/A - Just for bug bash tomorrow.

## Changes

In order for us to test MA in demo, let's temporarily replace its
configuration with sandbox.

## Context for reviewers

N/A

## Testing

N/A


GeorgeCodes19 (approved):

GeorgeCodes19:
I wish there were a cleaner way to do this!

PR #267, opened 2024-09-10
1549: Fix logos in MA/NYC emails

tdooner:
## Ticket

Resolves FFS-1549.

## Changes

These logos weren't loading because there was no `asset_host` specified.
Let's re-use the `DOMAIN_NAME` env var for it.

## Context for reviewers

N/A

## Testing

Tested in development by copying this config into development.rb and
verifying it was used in the Rails mailer.


tdooner:
Yeah, interesting, it seems possibly related. I'm going to merge and we can accept this in demo.

PR #273, opened 2024-09-11
1567: Change Date Field

allthesignals:
## Ticket

Resolves [1567](https://jiraent.cms.gov/browse/FFS-1567).

## Changes

Changes the date label for MA, specifically recert date. Requests today's date. Renames the dictionary key for the label. 

## Testing
image

[client-ma.pdf](https://github.com/user-attachments/files/16970183/client-ma.pdf)
[caseworker-ma.pdf](https://github.com/user-attachments/files/16970184/caseworker-ma.pdf)


tdooner (approved):
Lgtm. We could pre-fill the field in the controller for MA if we wanted, like

```ruby
@cbv_flow_invitation = CbvFlowInvitation.new
if @site_id == "ma"
  @cbv_flow_invitation.snap_application_date ||= Date.today
end
```

PR #275, opened 2024-09-12
Add script to ease upgrading template-infra

tdooner:
## Ticket

Related to FFS-1422.

## Changes
* Add `bin/update-infra-template` script

## Context for reviewers

The template-infra repo contains a script to upgrade itself, but this
script didn't work for me because it applies patches as an
all-or-nothing operation, and of course we have made modifications to
our terraform that will cause the patch application to fail.

To make it easier on myself, I wrote a script that automatically looks
for new released versions of `template-infra` and steps through
upgrading them one version at a time. Each version is applied by
applying the individual commits, and the pool soul doing the upgrade
will have to resolve any conflicts commit-by-commit.

Using this script, I was able to upgrade 3 minor versions and 4 patch
versions in about 3 hours of work. I still did have to clean up the git
history after this, and make sure everything still works, so the script
could certainly be improved, but alas, I need to timebox this due to
launch concerns.

## Testing

This is going to be a pain. I tested this by verifying that `make
infra-update-app-service` doesn't do anything too weird, but we'll still have
to deploy this carefully.

### DIFF
--- a/bin/update-infra-template
+++ b/bin/update-infra-template
@@ -1,0 +1,126 @@
+#!/bin/bash
+# This script will guide you through updating the infra template. It replaces
+# the official `update-template` script since we have made so many
+# customizations to the infra code.
+#
+# To run this, make sure:
+# 1. You have a `template-infra` repository as a sibling directory to this
+#    repository's directory
+# 2. That the `template-infra` repo is up-to-date.
+#
+set -euo pipefail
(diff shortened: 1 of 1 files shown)


tdooner:
@allthesignals Since it's just a bash script, and it won't break anything by itself, I'm going to merge.

PR #279, opened 2024-09-13
1480: Add a basic maintenance mode option

allthesignals:
## Ticket

Resolves [1480](https://jiraent.cms.gov/browse/FFS-1480)

## Changes

Adds a maintenance mode view that is enabled with an environment variable

## Context for reviewers

Maintenance mode is enabled through an environment variable:

```
MAINTENANCE_MODE=true bin/dev
```

Terraform is updated with the new variable.

## Testing

image

### DIFF
--- a/app/spec/requests/pages_spec.rb
+++ b/app/spec/requests/pages_spec.rb
@@ -6,5 +6,29 @@ RSpec.describe "Pages", type: :request do
       get "/"
       expect(response).to have_http_status(:success)
     end
+
+    context "when the maintenance mode is enabled" do
+      around do |ex|
+        stub_environment_variable("MAINTENANCE_MODE", "true", &ex)
+      end
+
+      it "redirects to maintenance path" do
+        get "/"
+
+        expect(response).to redirect_to(maintenance_path)
+      end
+    end
+
+    context "when the maintenance mode is disabled" do
+      around do |ex|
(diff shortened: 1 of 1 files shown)


tdooner (commented):
Uhhh, is this intended to work without us having to deploy? It looks like this gem works by checking for whether a maintenance file exists on disk (the file is created by the rake task). But since we are operating in container-land, there's no way to make this file in the containers that are running in production, so we would have to create this file locally and then deploy it to production.

Can we just do it with an environment variable and a before_action in ApplicationController instead?

app/app/controllers/application_controller.rb
```diff
@@ -52,4 +53,10 @@ def add_newrelic_metadata
 
     NewRelic::Agent.add_custom_attributes(attributes)
   end
+
+  def redirect_if_maintenance_mode
+    if ENV["MAINTENANCE_MODE"] == "true"
+      redirect_to maintenance_path
+    end
+  end
```
tdooner, line 61: I know it's just a single `if` statement, but, a test would be nice to ensure that if we change the logic later, we don't accidentally break it. (Look at the bright side.. it's only a single `if` statement so it will be a very simple test.)

Also, a test that the maintenance page renders properly would be nice.

tdooner (approved):
Lgtm but would really enjoy seeing a test.

PR #281, opened 2024-09-17
Input validation for caseworker invitation forms.

GeorgeCodes19:
## Ticket

Resolves ([FFS-1412](https://jiraent.cms.gov/browse/FFS-1412))

## Changes

> Added model-level validation for the `CbvFlowInvitation` model
> Added Tests for validation
> Added [`en.yml` translations for Active Record Models](https://guides.rubyonrails.org/i18n.html#translations-for-active-record-models)
> Ignoring these same translations when running `I18n` unused task
> Updated USWDS form helper to support multiple model-level errors. .

## Testing

I tested forms locally and created tests for them. **The test coverage for this feature can be improved. Several areas where increased coverage would benefit:**

- When sending email via AWS fails - There is an exception that is caught and an error informing the case worker that the invitation email could not be delivered to the provided email. Currently it is assumed that this works, but the code has not yet been tested.
- The `datepicker` input could have some client-side validation enforcing the same rules as the model. _I think there may be a Rails way of doing this?_

Screenshot 2024-09-16 at 21-17-59 Send an invite link SNAP Income Pilot

### Git message prefix
Screenshot 2024-09-19 at 5 44 24 PM


app/config/locales/en.yml
```diff
@@ -1,5 +1,35 @@
...
+            first_name:
+              blank: Enter the client's first name.
+            last_name:
+              blank: Enter the client's last name.
+            middle_name:
+              invalid_format: must be letters only
+            snap_application_date:
+              invalid_date: is not a valid date
+              ma_future_date: Enter today's date or the date you contacted the client. This date must be today or in the past 30 days.
+              ma_too_old: Enter today's date or the date you contacted the client. This date must be today or in the past 30 days.
```
makaelastephens, line 30: Enter today's date or the date you contacted the client. This date must be today or in the past year.

makaelastephens:
The content looks right to me!

app/app/views/caseworker/cbv_flow_invitations/_sandbox.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <%= f.text_field :case_number %>
+</div>
+
+<div class="field_with_errors">
+  <%= f.date_picker :snap_application_date, label: t(".invite.snap_application_date"), hint: "mm/dd/yyyy" %>
+</div>
+
+<div class="field_with_errors">
+  <%= f.text_field :email_address %>
+</div>
```
tdooner, line 23: Why are all of these surrounded in a `field_with_errors` class? I would expect the UswdsFormBuilder to do that?

app/app/views/caseworker/cbv_flow_invitations/new.html.erb
```diff
@@ -4,8 +4,8 @@
 <div class="usa-prose">
   <%= t(".description_html", pay_income_days: current_site.pay_income_days) %>
 </div>
-
-<%= form_with(model: @cbv_flow_invitation, url: invitations_path, builder: UswdsFormBuilder) do |f| %>
-  <%= render partial: "#{@site_id}", locals: { f: f } %>
+<% site_id = @cbv_flow_invitation.site_id || @site_id %>
```
tdooner, line 7: Nice catch - it looks like `@site_id` is only set in the `new` action, not in `create`. It might be better to just initialize the object with the `site_id` attribute, and then we could consistently access this as `@cbv_flow_invitation.site_id`.

E.g. in the `new` action:
```ruby
@cbv_flow_invitation = CbvFlowInvitation.new(site_id: site_id)
```

Not a blocker though. We could do that later.

app/config/locales/en.yml
```diff
@@ -1,5 +1,35 @@
...
+              invalid_format: Enter the client's CIN in the correct format, XX00000X.
+            email_address:
+              blank: Enter the client's email address.
+              invalid_format: Enter an email address in the correct format, like name@example.com
+            first_name:
+              blank: Enter the client's first name.
+            last_name:
+              blank: Enter the client's last name.
+            middle_name:
+              invalid_format: must be letters only
```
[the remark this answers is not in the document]
tdooner: Yeah - this validation error can be deleted from the I18n strings since we aren't actually doing that validation anymore.

app/config/locales/en.yml
```diff
@@ -1,5 +1,35 @@
...
+            snap_application_date:
+              invalid_date: is not a valid date
+              ma_future_date: Enter today's date or the date you contacted the client. This date must be today or in the past 30 days.
+              ma_too_old: Enter today's date or the date you contacted the client. This date must be today or in the past 30 days.
+              nyc_future_date: SNAP application/recertification date must be today or in the past 30 days.
+              nyc_too_old: SNAP application/recertification date must be today or in the past 30 days.
```
tdooner, line 32: Question for @makaelastephens  - I think the intent of the story is that the error message should say "SNAP application date" in NYC and "SNAP recertification date" in MA. Is that right?

app/config/locales/en.yml
```diff
@@ -25,13 +55,12 @@ en:
   caseworker:
     cbv_flow_invitations:
       create:
-        invite_failed: 'Error sending invitation to %{email_address}: %{error_message}.'
```
tdooner, line 28: It looks like you renamed this to `invite_error` but might have forgotten to add that string to this file?

app/spec/factories/cbv_flow_invitation.rb
```diff
@@ -1,23 +1,33 @@
...
-    snap_application_date { Date.today.strftime("%m/%d/%Y") }
+    snap_application_date { Time.zone.today.strftime("%m/%d/%Y") }
     user
 
+    trait :with_provider do
+      after(:build) do |invitation|
+        invitation.define_singleton_method(:provider) do
+          CbvFlowInvitationProvider
+        end
+      end
```
tdooner, line 18: What? This is really complicated... Why is including the provider even necessary? I'd prefer we just inline those `generate_*` methods into the factory, personally, since this is the only file that is using it.

app/spec/mailers/applicant_mailer_spec.rb
```diff
@@ -46,7 +46,15 @@
...
   context "for a MA CbvFlowInvitation" do
-    let(:cbv_flow_invitation) { create(:cbv_flow_invitation, :ma, email_address: email) }
+    let(:cbv_flow_invitation) do
+      begin
+        create(:cbv_flow_invitation, :ma, email_address: email)
+      rescue => e
+        puts "Exception caught in cbv_flow_invitation creation: #{e.class} - #{e.message}"
+        puts e.backtrace.join("\n")
+        raise e
+      end
```
tdooner, line 56: This seems like some vestigial code you were using during development? We shouldn't be catching exceptions here in general - rspec will happily tell us if these `let`s have an exception.

config/locales/en.yml
```diff
@@ -0,0 +1 @@
+
```
tdooner, line 1: Looks like this file was inadvertently created. We shouldn't have a top-level `config/locales/en.yml` in the monorepo root.

tdooner (changes requested):

app/app/models/cbv_flow_invitation.rb
```diff
@@ -1,16 +1,46 @@
...
+  MA_AGENCY_ID_REGEX = /\A\d{7}\z/
+
+  # Massachusetts: 6 alphanumeric characters
+  MA_BEACON_ID_REGEX = /\A[a-zA-Z0-9]{6}\z/
+
+  # New York City: 11 digits followed by 1 uppercase letter
+  NYC_CASE_NUMBER_REGEX = /\A\d{11}[A-Z]\z/
+
+  # New York City: 2 uppercase letters, followed by 5 digits, followed by 1 uppercase letter
+  NYC_CLIENT_ID_REGEX = /\A[A-Z]{2}\d{5}[A-Z]\z/
```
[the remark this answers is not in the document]
makaelastephens: I didn't know we were doing this.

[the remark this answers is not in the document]
makaelastephens: I think we might want to see if Casey can test the validation before we merge to prod (NYC eligibility specialist)

app/config/locales/en.yml
makaelastephens: No - DTA changed it.
The field NYC enters is "SNAP application/recertification date"
The field for DTA is "Today's date".

app/config/locales/en.yml
GeorgeCodes19: @makaelastephens Looks like after you and I did a review there were some changes that I didn't push. We actually got rid of `ma_too_old ` and `ma_future_date` and kept a single `ma_invalid_date` and `nyc_invalid_date`. 

```suggestion
            snap_application_date:
              invalid_date: is not a valid date
              ma_invalid_date: Enter today's date or the date you contacted the client. This date must be today or in the past year.
              nyc_invalid_date: SNAP application/recertification date must be today or in the past 30 days.
```

app/spec/factories/cbv_flow_invitation.rb
GeorgeCodes19: So I've tried a number of different approaches to resolve this. I kept running into: 
`FactoryBot::MethodDefinitionError: Defining methods in blocks (trait or factory) is not supported `

I don't find this approach to be overly complex. I'm actually a fan of how FactoryBot provides hooks and that it's even possible to augment the model instance. In attempt to add methods to the FactoryBot instance I broke 45 tests. So I'm tempted to stick with this unless you have clear direction on how to accomplish this.

.githooks/pre-commit
```diff
@@ -29,3 +29,26 @@ do
...
+if [ -n "$TICKET" ]; then
+    # Read the current commit message
+    COMMIT_MSG=$(cat $COMMIT_MSG_FILE)
+
+    # Check if the commit message already starts with the ticket number
+    if [[ $COMMIT_MSG != FFS-$TICKET:* ]]; then
+        # Prepend the ticket number to the commit message
+        sed -i.bak "1s/^/FFS-$TICKET: /" $COMMIT_MSG_FILE
+    fi
+fi
```
tdooner, line 54: Cool, will be interested to try this!

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
```diff
@@ -5,24 +5,38 @@ class Caseworker::CbvFlowInvitationsController < Caseworker::BaseController
...
+      error_header = "#{"error".pluralize(error_count)} occurred"
+
+      # Collect error messages without attribute names
+      error_messages = @cbv_flow_invitation.errors.messages.values.flatten.map { |msg| "<li>#{msg}</li>" }.join
+      error_messages = "<ul>#{error_messages}</ul>"
+
+      flash[:alert_heading] = error_header
+      flash[:alert] = error_messages.html_safe
+
+      return render :new
```
tdooner, line 39: No `return` keyword necessary here.

app/app/models/cbv_flow_invitation.rb
```diff
@@ -1,16 +1,46 @@
...
+  before_validation :format_case_number, if: :nyc_site?
+
   validates :site_id, inclusion: Rails.application.config.sites.site_ids
-  validates :case_number, presence: true, if: :nyc_site?
-  validates :agency_id_number, presence: true, if: :ma_site?
-  validates :beacon_id, presence: true, if: :ma_site?
-  validates :email_address, presence: true
+  validates :first_name, presence: true
+  validates :last_name, presence: true
+  validates :middle_name, presence: false
```
tdooner, line 27: I think we could just remove this line since it's not validating anything, right?
```suggestion
```

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
```diff
@@ -5,24 +5,38 @@ class Caseworker::CbvFlowInvitationsController < Caseworker::BaseController
...
+      @cbv_flow_invitation = CbvInvitationService.new.invite(invitation_params, current_user)
+    rescue => e
       flash[:alert] = t(".invite_failed",
                         email_address: cbv_flow_invitation_params[:email_address],
-                        error_message: ex.message
-                       )
-      Rails.logger.error("Error sending CBV invitation: #{ex.class} - #{ex.message}")
-      return redirect_to new_invitation_path(site_id: params[:site_id])
+                        error_message: e.message)
+      redirect_to caseworker_dashboard_path(site_id: params[:site_id])
```
tdooner, line 25: Sorry I didn't notice this earlier. We need the `return` here or else the logic below will continue executing and we will get an error with the `render :new` (a "double render" error when you try to tell Rails two different things to do with the response).

```suggestion
      return redirect_to caseworker_dashboard_path(site_id: params[:site_id])
```

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
tdooner, line 39: nitpick: No return keyword is necessary here.

```suggestion
     render :new
```

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
```diff
@@ -72,7 +86,8 @@ def cbv_flow_invitation_params
       :email_address,
       :snap_application_date,
       :agency_id_number,
-      :beacon_id
+      :beacon_id,
+      :welid
```
tdooner, line 90: The field is still named `beacon_id` even though it is labeled in the UI as the "WELID", right?

```suggestion
      :beacon_id
```

app/app/models/cbv_flow_invitation.rb
tdooner, line 27: We could probably remove this since it's not validating anything, right?
```suggestion
```

app/spec/mailers/applicant_mailer_spec.rb
```diff
@@ -15,7 +15,7 @@
   let(:email) { 'me@email.com' }
   let(:cbv_flow_invitation) { create(:cbv_flow_invitation, email_address: email) }
   let(:mail) { ApplicantMailer.with(cbv_flow_invitation: cbv_flow_invitation).invitation_email }
-
+  let(:valid_agency_id_number) { "1234567" }
```
tdooner, line 18: It doesn't look like this is being used elsewhere in the test.
```suggestion
```

app/spec/mailers/caseworker_mailer_spec.rb
```diff
@@ -43,13 +46,17 @@
...
+      email_body = strip_tags(email_html).gsub(/\s+/, ' ').strip
+      expected_date = format_parsed_date(Time.zone.today)
+      request_date = format_parsed_date(cbv_flow.created_at)
       expect(email_body).to include("Attached is an Income Verification Report PDF with confirmation number #{cbv_flow.confirmation_code}")
       expect(email_body).to include("confirm that their information has been submitted to HRA")
       expect(email_body).to include("This report is associated with the case number ABC1234 and CIN #{invitation.client_id_number}")
-      expect(email_body).to include("It was requested by #{caseworker_email} on #{format_parsed_date(cbv_flow.created_at)}")
-      expect(email_body).to include("submitted by the client on #{format_parsed_date(Date.today)}")
+      expect(email_body).to include("It was requested by #{caseworker_email} on #{request_date}")
+      expect(email_body).to include("submitted by the client on #{expected_date}")
```
tdooner, line 59: Interesting. Was this test failing due to the wrong date being in there? Maybe it was another test that implicitly relied on being run before 8pm ET. Thanks for fixing it.

app/app/models/cbv_flow_invitation.rb
```diff
@@ -71,11 +99,45 @@ def parse_snap_application_date
...
+
+  def nyc_snap_application_date_not_more_than_30_days_ago
+    if snap_application_date.present? && snap_application_date < 30.day.ago.to_date
+      errors.add(:snap_application_date, :nyc_invalid_date)
+    end
+  end
+
+  def format_case_number
+    if case_number.present? && case_number.length == 12
+      self.case_number = case_number.rjust(12, "0")
```
tdooner, line 140: This logic still looks wrong to me, please double check before merging.

tdooner (approved):
Looks good to me! Thanks for addressing everything I came up with before. The `format_case_number` thing is the last blocker; after fixing that, feel free to merge.

app/spec/mailers/caseworker_mailer_spec.rb
GeorgeCodes19: I was actually seeing date/time differences. Hence why I switched to a different date initializer. `Time.zone.today`

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
GeorgeCodes19: Removing the return here results in a double render

```
AbstractController::DoubleRenderError: Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...); return".

  0) Caseworker::CbvFlowInvitationsController#create requires the ma fields
     Failure/Error: redirect_to caseworker_dashboard_path(site_id: params[:site_id])

     AbstractController::DoubleRenderError:
       Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...); return".
     # ./app/controllers/caseworker/cbv_flow_invitations_controller.rb:46:in `create'
     ```

GeorgeCodes19: Removing this return statement results in double render.

tdooner: My bad, I misread the code. Sorry for the wild goose chase.

PR #284, opened 2024-09-19
Fix MA GPG encryption in production

tdooner:
## Ticket

N/A - Discovered during E2E testing

## Changes

In production (only), our docker containers are read-only, which prevents the
GPG key from being added to the local keyring.

We should probably redo this to happen during build-time, but until then, let's
tell GPG to create our keyring in the writable directory.

## Context for reviewers
N/A

## Testing

Tested locally with an incantation of `docker run --read-only` and bind
mounting.


GeorgeCodes19 (approved):
🤦🏾‍♂️

PR #285, opened 2024-09-23
George/ffs 1412 input validation

GeorgeCodes19:
## Ticket

Resolves [FFS-1412](https://jiraent.cms.gov/browse/FFS-1412)

## Changes

- Removed error entries causing `"can't be blank"` error messages
- Updated tests to assert for new validation conditions

## Context for reviewers

> Per agency feedback, there are several fields that were showing error messages about being blank without providing field names. Error messages pertaining to an input field being blank should use the language entries pertaining to the respective field having an invalid value instead.

### MA DTA
Screenshot 2024-09-23 at 4 10 06 PM

### NY HRA
Screenshot 2024-09-23 at 4 10 26 PM


app/app/models/cbv_flow_invitation.rb
```diff
@@ -18,14 +18,13 @@ class CbvFlowInvitation < ApplicationRecord
 
   has_secure_token :auth_token, length: 36
 
-  before_validation :parse_snap_application_date
+  before_validation :parse_snap_application_date, if: -> { !:snap_application_date.empty? }
```
GeorgeCodes19, line 21: Does this look correct? Is it really needed?

makaelastephens (commented):
For **NYC** can we change 
 snap_application_date: SNAP application date

to 
 snap_application_date: SNAP application or recertification date

app/config/locales/en.yml
```diff
@@ -9,8 +9,7 @@ en:
               blank: Enter a valid agency ID number.
               invalid_format: Agency ID number must be 7 digits.
             beacon_id:
-              blank: can't be blank
-              invalid_format: Your WelID must be 6 characters. It can only include letters and numbers.
+              invalid_format: Your WELID must be 6 characters. It can only include letters and numbers.
```
tdooner, line 12: Remember to capitalize the WELID in the field label as well

app/db/schema.rb
```diff
@@ -51,8 +51,8 @@
     t.string "site_id"
     t.string "confirmation_code"
     t.datetime "transmitted_at"
-    t.datetime "consented_to_authorized_use_at"
     t.datetime "redacted_at"
+    t.datetime "consented_to_authorized_use_at"
```
tdooner, line 55: Ideally we don't commit this but it's not harming anything.

tdooner (approved):

PR #288, opened 2024-09-24
1425: Import Spanish Translations

GeorgeCodes19:
## Ticket

Resolves [FFS-1425](https://jiraent.cms.gov/browse/FFS-1425)

## Changes

> Added a TranslationService that I will attempt to make as generic as possible in order to import various other translations

- `TranslationService` includes the ability to register multiple `modifiers` in order to update an individual row's values
- `TranslationService` includes the ability to skip rows by registering a method to skip rows.

I believe that the two middlewares above can be joined into a single function that operates on a per row basis. Returning `false` could skip the row while other operations could allow the row value to be modified in the lambda/closure.

## Context for reviewers
We'll still want to walk through each page and translation.

## Testing

Tested using the Rails console, rspec, and visually.

Screenshot 2024-09-23 at 8 26 05 PM

Screenshot 2024-09-23 at 8 20 14 PM
Screenshot 2024-09-23 at 8 20 53 PM
Screenshot 2024-09-23 at 8 19 53 PM


app/spec/services/translation_service_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    context 'with Spanish translations' do
+      it 'generates a YAML file with Spanish translations and skips rows based on condition' do
+        result = TranslationService.generate(
+          csv_path.to_s,
+          output_path.to_s,
+          target_locale: 'es',
+          skip_row_conditions: [
+            TranslationService.skip_no_translation,
```
GeorgeCodes19, line 26: It's probably best to create a different `skip_no_translation` callable to pass. Maybe creating an inline CSV or putting a new CSV stub in the support/fixtures directory and using column values that will allow a `skip_no_translation` callable to exhibit the behavior that we expect.

app/app/services/translation_service.rb
```diff
@@ -0,0 +1,97 @@
...
+    csv_content = csv_lines.join
+
+    CSV.parse(csv_content, headers: true, header_converters: :symbol, skip_blanks: true).each_with_index do |row, index|
+      # Skip the first row as it's a document header
+      next if index == 0
+
+      row_count += 1
+
+      # Apply row modifiers
+      row_modifiers.each do |modifier|
```
GeorgeCodes19, line 39: We can likely combine `modifiers` and `skip conditions` into a single middleware that returns a boolean. Whereas `false` implies that the CSV row should not be translated and imported.

carmenrosalop (commented):
Thanks @GeorgeCodes19 I'm so glad the translations process was smooth. The CBV pages look great! I'd like to see how the client-facing PDF report in Spanish looks like. Could you please add it here?

GeorgeCodes19:
> Thanks @GeorgeCodes19 I'm so glad the translations process was smooth. The CBV pages look great! I'd like to see how the client-facing PDF report in Spanish looks like. Could you please add it here?

Yes! I'll actually send you the PDFs docs and attach images.

app/spec/support/fixtures/I18n/snap_income_pilot_translations.csv
GeorgeCodes19, line 1: Should we delete this or have a version that's scrubbed a bit more before adding it to the repo?

tdooner: Personally I don't think it's very useful to commit to the repo. If you want to commit something so you can test your script, can you just commit the first couple lines?

tdooner (approved):
I didn't really review the script, because if it worked to import the spanish CSV into the YAML here, that's good enough for me. Some documentation might be useful as a comment in that file that describes how to call the code.

PR #289, opened 2024-09-24
Bump webrick

allthesignals:
## Ticket

https://github.com/DSACMS/iv-cbv-payroll/actions/runs/11016602537/job/30592509373?pr=282

## Changes

Version bump webrick


tdooner (approved):
Seems fine to me, but would be even better if it didn't add it to the Gemfile

PR #291, opened 2024-09-25
1412: Email validation fixes

GeorgeCodes19:
## Ticket

N/A

## Changes

Bugfix that replaces the email validation regex supplied by Rails with one that's a bit more strict with one that requires that email addresses contain a TLD.

## Context for reviewers
https://nava.slack.com/archives/C06FC5TPAR3/p1727269222636339
## Testing

Screenshot 2024-09-25 at 11 09 17 AM
Screenshot 2024-09-25 at 11 09 06 AM


app/spec/models/cbv_flow_invitation_spec.rb
```diff
@@ -2,11 +2,49 @@
...
+
+          # This test is to provide evidence that URI::MailTo::EMAIL_REGEXP is NOT what we want to use
+          # https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+          #
+          # EXCERPT: This requirement is a willful violation of RFC 5322, which defines a syntax for email addresses
+          # that is simultaneously too strict (before the "@" character), too vague (after the "@" character),
+          # and too lax (allowing comments, whitespace characters, and quoted strings in manners unfamiliar to most users)
+          # to be of practical use here.
+          it "URI::MailTo::EMAIL_REGEXP considers an email without a TLD to be valid" do
+            expect(invalid_email_no_tld).to match(URI::MailTo::EMAIL_REGEXP)
```
GeorgeCodes19, line 30: I need to revisit this. It's a placeholder for now. This isn't making the assertion that want it to. We want to assert that when using this REGEX that no validation error is thrown provided an email address without a TLD.

app/app/models/cbv_flow_invitation.rb
```diff
@@ -1,6 +1,5 @@
 class CbvFlowInvitation < ApplicationRecord
-  EMAIL_REGEX = URI::MailTo::EMAIL_REGEXP
-
+  EMAIL_REGEX = /\A([\w+\-].?)+@[a-z\d\-]+(\.[a-z]+)*\.[a-z]+\z/i
```
tdooner, line 2: Sigh, this is a classically hard problem to solve. This regex would be invalid for domains that have subdomains and numbers, i.e. "foo@test1.test2.com". That seems unlikely so I'll approve this but my point here is that it's going to be hard to figure out all the edge cases/possibilities.

Personally the best easy validation I've seen is just to verify that the email address has an "@" in it and a "." following that, i.e. something like.:

```ruby
/\A(.*)@(.*)\.(.*)\z/
```

app/spec/models/cbv_flow_invitation_spec.rb
tdooner, line 30: IMO, it's not useful for us to have a test that isn't testing the correctness of our code, and consequently that will never fail. Taken to its extreme, if we were to do this often, we could be spending test suite time running tests that don't provide value.

app/spec/models/cbv_flow_invitation_spec.rb
```diff
@@ -2,11 +2,49 @@
...
+          end
+        end
+
+        context "when email address is invalid" do
+          invalid_email_addresses = %w[johndoe@gmail johndoe@gmail..com johndoe@gmail.com..com johndoe@gmail\ .\ com]
+          invalid_email_addresses.each do |email|
+            it "validates #{email} format" do
```
tdooner, line 37: nit: A better message here, like `determines #{email} is invalid` would make it slightly easier to tell from the test failure message what test failed.

tdooner (approved):

app/app/models/cbv_flow_invitation.rb
GeorgeCodes19: Screenshot 2024-09-25 at 2 05 23 PM

That regex doesn't seem to satisfy some email addresses that are clearly invalid.

app/spec/models/cbv_flow_invitation_spec.rb
GeorgeCodes19: How else would you recommend that we take note of approaches that have not worked? I think we also have some weirdness with DateTime/Time zones that we could also note.

tdooner: I think the comment approach you adopted looks good.

app/app/models/cbv_flow_invitation.rb
tdooner: What I meant to say is that the point of this regex is just to prevent common caseworker typos, not to perfectly validate an email address format. The worst case scenario is we have a false positive where the regex says an email address is invalid, but it actually is valid, and then the caseworker gets stuck. So at the minimum, I think it would be good to handle the edge case I noticed by changing the regex to allow numbers and dashes in the domain components, since those may happen.

```suggestion
  EMAIL_REGEX = /\A([\w+\-].?)+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z\d\-]+\z/i
```

PR #293, opened 2024-09-26
Fix `invalidated_session_ids` system error when logging out twice

tdooner:
## Ticket

N/A - Fixes system error in logs.

## Changes

The "before_logout" hook can be called with a `nil` user if the user has
already logged out (or their session timed out, presumably). In this
case, we can't do anything to invalidate the session because Devise
won't tell us which user it was related to. That's fine - we don't need
to invalidate the session, as Devise will already consider the session
timed out.

This commit fixes the error that comes up with `user` is nil in this
case.

## Context for reviewers

N/A

## Testing

N/A

### DIFF
--- a/app/app/services/session_invalidation_service.rb
+++ b/app/app/services/session_invalidation_service.rb
@@ -33,6 +33,8 @@ class SessionInvalidationService
   end
 
   def valid?
+    return false unless @user.present?
+
     (@user.invalidated_session_ids || {}).exclude?(@session_id)
   end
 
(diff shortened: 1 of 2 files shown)


app/app/services/session_invalidation_service.rb
```diff
@@ -33,6 +33,8 @@ def initialize(user, session_id)
   end
 
   def valid?
+    return false unless @user.present?
```
GeorgeCodes19, line 36: Forgetting to add guard clauses has been biting us a little.

GeorgeCodes19 (approved):

PR #296, opened 2024-09-27
1425: Dynamic applicant language selection.

GeorgeCodes19:
## Ticket

## Changes

- [x] Add `language` column to CbvFlowInvitation
- [x] Enforce the selection of the applicant's preferred language by adding form validation
- [x] Added language toggle button to site header
- [x] Ensure that the language is being passed to New Relic event when a CBV Flow has been completed
- [x] Pass language to Pinwheel link to ensure that Pinwheel modal loads in the selected language
- [x] Update UI to match design specs
- [x] Write tests


app/app/helpers/view_helper.rb
```diff
@@ -2,11 +2,11 @@ module ViewHelper
...
   def format_active_locale(locale_string)
-    link_classes = "usa-nav__link"
-    if locale_string.to_sym == I18n.locale
-      link_classes = "#{link_classes} usa-current"
-    end
-    link_to t("shared.languages.#{locale_string}"), root_path(locale: locale_string), class: link_classes
+    # using request.fullpath to include query parameters so that
+    # switching languages doesn't lose the query parameters
+    current_path = request.fullpath.sub(/\A\/#{I18n.locale}/, "")
+    new_path = "/#{locale_string}#{current_path}"
```
tdooner, line 8: There is a better rails-y way to do it than with regular expressions:

```suggestion
    new_path = url_for(params.permit(:controller, :action, :site_id).merge(locale: locale_string))
```

app/app/helpers/view_helper.rb
```diff
@@ -2,11 +2,11 @@ module ViewHelper
...
+    link_to t("shared.languages.#{locale_string}"), new_path, class: "usa-nav__link"
```
tdooner, line 9: nit: This method is not very well-named. It doesn't format the active locale, it outputs a link that switches the locale to something else. Maybe a name like `switch_locale_link` would be more apt?

app/app/controllers/api/pinwheel_controller.rb
```diff
@@ -8,7 +8,8 @@ def create_token
     token_response = pinwheel.create_link_token(
       response_type: token_params[:response_type],
       id: token_params[:id],
-      end_user_id: @cbv_flow.pinwheel_end_user_id
+      end_user_id: @cbv_flow.pinwheel_end_user_id,
+      language: session[:locale]
```
tdooner, line 12: Let's pass the `locale` as a param to this endpoint, so we can access it the same way as the other `token_params` for `response_type` and `id`. It will be simpler to test and debug that way.

app/app/javascript/utilities/pinwheel.js
```diff
@@ -25,12 +30,13 @@ export function initializePinwheel(Pinwheel, linkToken, callbacks) {
...
 }
 
-export const fetchToken = (response_type, id) => {
+export const fetchToken = (response_type, id, locale) => {
   return fetch(PINWHEEL_TOKENS_GENERATE, {
     method: 'post',
     headers: {
       'X-CSRF-Token': CSRF.token,
       'Content-Type': 'application/json',
+      'Accept-Language': resolveLanguage(locale),
```
tdooner, line 39: We don't need to pass this header. `Accept-Language` is a header intended for browsers to hint what language their user speaks... if we're communicating from our own JS to our own API endpoint, it seems more idiomatic to just pass this data in as a query param instead (the same way we pass `response_type` and `id` as params to the endpoint)

app/app/views/caseworker/cbv_flow_invitations/_ma.html.erb
```diff
@@ -11,3 +11,25 @@
...
+          type="radio"
+          name="<%= f.object_name %>[language]"
+          value="<%= value %>"
+          <%= 'checked="checked"' if f.object.language == value %>
+        >
+        <label class="usa-radio__label" for="language-<%= value %>-<%= index %>">
+          <%= label %>
+        </label>
+      </div>
+    <% end %>
```
tdooner, line 33: Did you consider using `f.radio_button` for this? It might simplify this quite a bit. Something like...

```suggestion
    <% language_options.each do |value, label| %>
      <% f.radio_button :language, value, label: label %>
    <% end %>
```

app/app/views/caseworker/cbv_flow_invitations/_sandbox.html.erb
```diff
@@ -6,6 +6,34 @@
 
 <%= f.text_field :case_number, label: t(".invite.case_number") %>
 
-<%= f.date_picker :snap_application_date, label: t(".invite.todays_date") %>
+<%= f.text_field :beacon_id, label: t(".invite.beacon_id") %>
+
+<%= f.text_field :agency_id_number, label: t(".invite.agency_id_number") %>
+
+<%= f.text_field :client_id_number, label: t(".invite.client_id_number") %>
```
tdooner, line 13: nit: Did you add these fields to the sandbox invitation page for some particular reason? They seem a bit unnecessary to me.

app/app/controllers/cbv/summaries_controller.rb
```diff
@@ -191,7 +191,8 @@ def track_transmitted_event(cbv_flow, payments)
       paystub_count: payments.count,
       account_count_with_additional_information:
         cbv_flow.additional_information.values.count { |info| info["comment"].present? },
-      flow_started_seconds_ago: (Time.now - cbv_flow.created_at).to_i
+      flow_started_seconds_ago: (Time.now - cbv_flow.created_at).to_i,
+      language: I18n.locale
```
tdooner, line 195: We should also be storing the `language` on the CbvFlow when the PDF is sent. I see you added the database column, but I didn't see anywhere that the database column value is updated.

app/app/controllers/cbv/base_controller.rb
```diff
@@ -79,4 +79,26 @@ def agency_url
...
+
+    # Don't redirect if the user has selected the default locale, "en"
+    # Adding /en/ to the URL is unnecessary and will break many tests
+    return unless I18n.locale != I18n.default_locale
+
+    # Redirect only if the current URL doesn't match the selected locale
+    unless request.path.start_with?("/#{I18n.locale}")
+      redirect_to url_for(locale: I18n.locale, params: request.query_parameters)
+    end
+  end
```
tdooner, line 103: Hmm, you're overthinking this.... If we're using the URL param (`params[:locale]`) to indicate which locale the user wants to use, why don't we just update the invitation URL to have the correct locale in it?

What I mean is... if the invitation language is spanish, we can just update the email's URL to take the user to `/es/cbv/entries?token=1234` rather than `/cbv/entries?token=...`. I don't think we should need to do any of this stuff if we just send the user to the correct link to begin with.

tdooner (commented):
Looks like you're on the right track! The only blockers I would like to have resolved before merging are:
1. Removing the `request.path.split("/")` thing
2. Actually saving the `language` when the PDF is sent
3. Simplifying the invitation URL handling logic so we don't need the `set_locale` method

It would be nice to:
* Remove the storage of locale in the `session`
* Clean up and simplify some of the other suggestions

tdooner (approved):

PR #298, opened 2024-09-30
Get README Setup cleaned up, working

millerti:
# No Ticket
This does not resolve any ticket. I just proactively made these changes while trying to get the repo to run locally.

## Changes
1. Updated the README in general, primarily with up-to-date instructions in the Setup section.
1. Added a new migration that enables the pgcrypto extension, which is required for the db:migrate step to work on a fresh install.

## Testing Instructions
1. Clone a fresh copy of this repo
2. Follow the instructions in the Setup section of the README
3. At the end, the site should be working!


app/db/migrate/20240423000001_add_pgcrypto_extension.rb
GeorgeCodes19, line 1: I'm not sure if we need `pgcrypto` at the moment. Is there a reason why we'd want to enable this extension right now?

millerti: The migrations don't run without it. I got errors like the one [here](https://nava.slack.com/archives/C06FC5TPAR3/p1727713124753769). I suppose I am just putting a lot of trust in Matt that this is the correct solution... do you know another?

millerti: Matt may have found this solution from [this thread](https://stackoverflow.com/questions/47064090/rails-postgres-migration-why-am-i-receiving-the-error-pgundefinedfunction).

tdooner: It's odd that none of the rest of us hit this issue (although I have seen issues like this in the past, but not here). It makes me wonder whether you installed a different version of postgres somehow?

app/db/migrate/20240423000001_add_pgcrypto_extension.rb
```diff
@@ -0,0 +1,5 @@
+class AddPgcryptoExtension < ActiveRecord::Migration[7.1]
+  def change
+    enable_extension "pgcrypto"
+  end
+end
```
tdooner, line 5: Whenever you add a migration, you should run it and also commit the corresponding automatic change to `db/schema.rb`.

millerti: Got it, thanks! I definitely ran it. 😄 But didn't know that this file should be included.

tdooner (approved):

app/db/migrate/20240423000001_add_pgcrypto_extension.rb
millerti: Hmm... I feeeeeel like I remember I just had it installed (the Odin Project specifically talked about it and why I shouldn't install a new one, I think), which ime means that maybe I'm using a very outdated version

tdooner: @millerti Ah yeah, maybe worth checking your version of Postgres. We're running 14 in production, so ideally you have that one or something slightly newer.

millerti: Looks like I'm working with 14.13.

millerti: Here's a theory, please weigh in:

I deleted the migration I added and re-ran db:migrate. This completed successfully! This, to me, indicates that maybe pgcrypto is now installed in some way, whereas before it was not.

Therefore, I hypothesize that the reason you three did not see this issue is that you already had it installed somehow. Does that sound plausible?

If you agree, I'd suggest we keep the migration in place, because the fact of the matter appears to be that we depend on a function defined in pgcrypto.

millerti: @tdooner @GeorgeCodes19

tdooner: I agree - I doubt it'll hurt anything. The main thing to watch is whether we will be able to run this command on RDS when we are deploying - so I would just make sure that the deploy after you merge this succeeds.

tdooner:
I just re-triggered the OWASP scan for you. It often fails by hitting some "secondary rate limit". I haven't had a chance to debug it yet - retriggering it usually seems to fix the issue.

app/db/migrate/20240423000001_add_pgcrypto_extension.rb
tdooner: Also - in the future when generating migrations, make sure to do it with the generate command: `bin/rails generate migration AddPgcryptoExtension`. This will automatically set the timestamp in the filename correctly (yours is set to 4/23/2024, which is wrong, but I believe this will still work)

app/db/migrate/20240423000001_add_pgcrypto_extension.rb
millerti: Added back, sorted it first this time per advice on the internet. @GeorgeCodes19 , if you're satisfied, please resolve this conversation.

app/db/migrate/20240423000001_add_pgcrypto_extension.rb
millerti: No, I did that on purpose and manually, because I thought that this migration needed to happen before the others (to ensure that, when they needed the function that pgcrypto had, it was already there). But FYI that I did generate the migration this way, I just changed the filename.

PR #302, opened 2024-10-01
1781: Specify charset

allthesignals:
## Ticket

Closes https://jiraent.cms.gov/browse/FFS-1781

## Changes

Specifies charset in the PDF template

## Context for reviewers

Manually tested

image


tdooner (approved):
nit: When merging, if you use a title like "1781: Fix rendering of smart quotes in PDFs", it will be more clear to non-technical folks when reading the release notes what this commit is addressing.

PR #303, opened 2024-10-02
1492: Synchronizations follow up

allthesignals:
## Ticket

https://jiraent.cms.gov/browse/FFS-1492

## Changes

Switches to Turbo Streams for indicator views. Shows job statuses

## Context for reviewers

Test manually with https://docs.pinwheelapi.com/public/docs/sandbox#income-and-employment-job-errors

## Testing

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ Income Verification: Consent-Based Verification (Payroll)
 
 # About the Project
 
-Consent-Based Verification (CBV) is an approach to allow benefit applicants to opt to verify their income via products that pull directly from payroll providers. This repository implements a product to demonstrate this technology for testing and validation purposes.
+Consent-Based Verification (CBV) is a prototype that allows benefit applicants to verify their income directly using payroll providers. It is currently being piloted for testing and validation purposes.
 
 # Development and Software Delivery Lifecycle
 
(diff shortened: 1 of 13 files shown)


app/app/models/pinwheel_account.rb
```diff
@@ -1,6 +1,10 @@
 class PinwheelAccount < ApplicationRecord
   belongs_to :cbv_flow
 
+  after_update_commit {
+    broadcast_replace target: self, partial: "cbv/synchronizations/indicators", locals: { pinwheel_account: self }
```
tdooner, line 5: This is probably rendering in the default locale because there's no connection to the locale the current user is using.

You may be able to get the partial to render in the right locale by surrounding this in:

```ruby
I18n.with_locale(cbv_flow.cbv_flow_invitation.language) do
  broadcast_replace # ...
end
```

but that won't take into account any locale change during the flow. We could store the locale on the Cbv Flow as well (updating it every time it changes), which would be more resilient to this.

app/app/models/pinwheel_account.rb
```diff
@@ -29,11 +33,14 @@ def job_succeeded?(job)
     supported_jobs.include?(job) && send(sync_column).present? && send(error_column).blank?
   end
 
-  def job_completed?(job)
+  def synchronization_status?(job)
```
tdooner, line 36: If it doesn't return a boolean, it shouldn't end with a question mark

app/app/models/pinwheel_account.rb
```diff
@@ -29,11 +33,14 @@ def job_succeeded?(job)
...
     error_column, sync_column = event_columns_for(job)
     return nil unless error_column.present?
 
-    supported_jobs.include?(job) && (send(sync_column).present? || send(error_column).present?)
+    return :unsupported unless supported_jobs.include?(job)
+    return :succeeded if job_succeeded?(job)
+    return :in_progress if supported_jobs.include?(job) && (send(sync_column).blank? && send(error_column).blank?)
+    :failed if supported_jobs.include?(job) && (send(error_column).present?)
```
tdooner, line 43: I found it a bit hard to follow this logic - despite looking like guard clauses, these conditionals are all coupled together so they don't really allow a reader to ignore them as preconditions of the actual method's logic (in the way that a guard clause would).

I might suggest just using the old if/else constructs....

```suggestion
    if supported_jobs.exclude?(job)
      :unsupported
    elsif job_succeeded?(job)
      :succeeded
    elsif supported_jobs.include?(job) && (send(sync_column).blank? && send(error_column).blank?)
      :in_progress
    elsif supported_jobs.include?(job) && (send(error_column).present?)
      :failed
    end
```

This simpler code also makes it apparent that we could also simplify the last two conditions by removing `supported_jobs.include?(job)` since it will always be true if we've reached that branch (as the first branch will be executed instead if `supported_jobs.exclude?(job)`)

app/app/helpers/application_helper.rb
```diff
@@ -68,4 +68,9 @@ def feedback_form_url
       APPLICANT_FEEDBACK_FORM
     end
   end
+
+  # some job statuses we consider completed even if they failed
+  def coalesce_to_completed(status)
+    [ :unsupported, :failed ].include?(status) ? :completed : status
+  end
```
tdooner, line 75: I feel like this is a bit weird for a couple reasons, 1) it introduces a `completed` status that isn't in the `synchronization_status` method, and 2) it means that we have logic for overriding the sync status of certain jobs in a different place than the others.

My suggestion, which you can take or leave, would be to add yet another constant to PinwheelAccount like:

```ruby
  # A required job must succeed for the user's session to proceed
  REQUIRED_JOBS = %w[paystubs]
```

and then update the logic of `synchronization_status` to use it. I think the right logic would be to change the `if` statement for `:succeeded` to be `if job_succeeded?(job) || REQUIRED_JOBS.exclude?(job)`.

app/spec/models/pinwheel_account_spec.rb
```diff
@@ -71,4 +71,34 @@
...
+      end
+    end
+
+    context "when status is unsupported" do
+      it "returns unsupported" do
+        pinwheel_account.update!(supported_jobs: supported_jobs.reject { |job| job == 'income' })
+        expect(pinwheel_account.synchronization_status?('income')).to eq(:unsupported)
+      end
+    end
+  end
```
tdooner, line 103: Thanks for the tests, hopefully they will make the refactor I proposed easy!

tdooner (approved):
Looks good. Some ideas for refactoring that you can take if you agree that they would be simpler.

app/app/models/pinwheel_account.rb
GeorgeCodes19: Yup, it's [set on every request](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/app/controllers/application_controller.rb#L20).

PR #305, opened 2024-10-03
FFS-1803: Spanish Smoke Test Feedback

GeorgeCodes19:
## Ticket

[FFS-1798](https://jiraent.cms.gov/browse/FFS-1798) fix untranslated body paragraph in email
[FS-1802](https://jiraent.cms.gov/browse/FFS-1802) fix missing language button in mobile
[FFS-1803](https://jiraent.cms.gov/browse/FFS-1803) fix nyc incoming report sending in Spanish
[FFS-1800](https://jiraent.cms.gov/browse/FFS-1800) fix untranslated dates and date ranges

## Additional Changes

- Added translations for some Pinwheel data such as pay frequency

### Invitation email
Screenshot 2024-10-02 at 18-40-55 Mailer Preview for applicant_mailer#invitation_email_dta
Screenshot 2024-10-02 at 18-06-45 Su información de pago de Acme Corporation SNAP Income Pilot (1)


app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -112,6 +112,31 @@ def gross_pay
...
+      "biweekly" => :bi_weekly,
+      "daily" => :daily,
+      "hourly" => :hourly,
+      "monthly" => :monthly,
+      "permile" => :per_mile,
+      "semimonthly" => :semi_monthly,
+      "semiweekly" => :semi_weekly,
+      "variable" => :variable,
+      "weekly" => :weekly
+    }
```
tdooner, line 133: Where'd you get this list? Per https://docs.pinwheelapi.com/public/reference/get_income_v1_accounts__account_id__income___get,  only include a few of these are valid values. I guess better safe than sorry - I'm just wondering.

app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -112,6 +112,31 @@ def gross_pay
...
+
+    yaml_key = frequency_map[key] || frequency
+
+    I18n.t("payment_frequencies.#{yaml_key}", locale: :es, default: frequency)
```
tdooner, line 137: This could bite us later if we need to add support for more languages. It's fine for now since we don't, but, ideally we wouldn't be special-casing things like this. A general pattern you could use is to just put the pinwheel values as the I18n keys and then just check whether the key exists, like:

```ruby
if I18n.exists?("payment_frequencies.#{key}")
  I18n.t("payment_frequencies.#{yaml_key}")
else
  Rails.logger.warn "Unknown pay_frequency from Pinwheel: #{frequency}"
  I18n.t("cbv.payment_details.show.unknown")
end
```

app/app/helpers/view_helper.rb
```diff
@@ -7,33 +7,35 @@ def switch_locale_link(locale)
     link_to t("shared.languages.#{locale}"), path, class: "usa-nav__link", data: { "turbo-prefetch": false }
   end
 
-  def format_parsed_date(date)
+  def format_parsed_date(date, format = :default)
     begin
-      date.strftime(DATE_FORMAT)
+      I18n.l(date, format: format, locale: I18n.locale)
```
tdooner, line 12: I don't think you should need to pass `locale: I18n.locale` to these calls, since that's kind of the point of setting the locale as we surround the request with `I18n.with_locale(...)`

app/app/mailers/caseworker_mailer.rb
```diff
@@ -28,6 +28,8 @@ def set_params
   end
 
   def generate_pdf
+    # caseworkers should receive the report in English
+    I18n.locale = "en"
```
tdooner, line 32: Ok, sure, this is probably fine. Though, there would be fewer side effects to do it using `I18n.with_locale`, e.g.

```ruby
I18n.with_locale(:en) do
  # caseworkers should receive the report in English
  WickedPdf.new.pdf_from_string(
      render_to_string(template: "cbv/summaries/show", layout: "pdf", formats: [ :pdf ], locals: { is_caseworker: true }),
      footer: { right: "Income Verification Report | Page [page] of [topage]", font_size: 10 }
    )
end
```

tdooner (approved):
No blockers, but some suggestions (of course).

app/app/controllers/cbv/payment_details_controller.rb
GeorgeCodes19: These are in the original CSV. I wrongly assumed that these had already been cross referenced.

app/app/mailers/caseworker_mailer.rb
GeorgeCodes19: Oh wow, this approach is pretty interesting. More magic.

app/app/helpers/view_helper.rb
GeorgeCodes19: Thanks. That's a good callout.

PR #308, opened 2024-10-04
an attempt that seems to kind of fix the issue, but I think may be ov…

millerti:
…erkill

## Ticket

## Changes

## Context for reviewers

> Testing instructions, background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.

## Testing


millerti:
Question for Tom: I see the trivy-scan wants us to bump the version of linux-libc-dev in our Dockerfile. Should I just do that? I'm not aware of how we usually test changes to the Dockerfile.

tdooner (approved):

PR #311, opened 2024-10-08
1488: Explicitly set margin to something slightly higher than the default

allthesignals:
## Ticket

https://jiraent.cms.gov/browse/FFS-1488

## Changes

Explicitly sets margins during PDF generation

## Context for reviewers

See discussion here: https://nava.slack.com/archives/C06FC5TPAR3/p1728411846834959?thread_ts=1724784739.697159&cid=C06FC5TPAR3

When margins are not set (or set to 0), the margin page numbers are not included. This is just an experiment to see.


tdooner (approved):
Worth a shot.

PR #313, opened 2024-10-08
Include cbv flow id in pinwheel link error events

allthesignals:
## Ticket

NA

## Changes

This adds thie CbvFlow ID to the browser-emitted NewRelic events so that we can see where/why people are failing on Employer Search

## Testing

Not sure we have test support for this just yet at least not without Capybara... with some feature specs we can stub the NREUM thing and assert that way?

### DIFF
--- a/app/app/views/cbv/employer_searches/show.html.erb
+++ b/app/app/views/cbv/employer_searches/show.html.erb
@@ -3,7 +3,7 @@
   <%= t(".header") %>
 </h1>
 
-<div data-controller="cbv-employer-search" data-cbv-employer-search-cbv-flow-id-value=<%= @cbv_flow.id %>>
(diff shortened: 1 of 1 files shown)


app/app/views/cbv/employer_searches/show.html.erb
```diff
@@ -3,7 +3,7 @@
   <%= t(".header") %>
 </h1>
 
-<div data-controller="cbv-employer-search">
+<div data-controller="cbv-employer-search" data-cbv-employer-search-cbv-flow-id-value=<%= @cbv_flow.id %>>
```
tdooner, line 6: Lol at the length of the name.

Let's put the value in quotes though, just to be a bit more HTML-y?

```suggestion
<div data-controller="cbv-employer-search" data-cbv-employer-search-cbv-flow-id-value="<%= @cbv_flow.id %>">
```

tdooner (approved):

PR #314, opened 2024-10-08
Fix data redaction to only redact records once

tdooner:
## Ticket

N/A - Bugfix.

## Changes

While running data redaction, I noticed that it was redacting every record
every time. To avoid wasting time on this every time redaction is run, let's
skip redacting any records that have previously been redacted.

## Context for reviewers

Also, I noticed the error wasn't reported to NewRelic. I've added rake task
monitoring in an attempt to fix that.

## Testing

Unit tests included. Will test manually in demo.

### DIFF
--- a/app/app/models/concerns/redactable.rb
+++ b/app/app/models/concerns/redactable.rb
@@ -24,6 +24,11 @@ module Redactable
     uuid: "00000000-0000-0000-0000-000000000000"
   }
 
+  included do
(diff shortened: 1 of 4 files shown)


GeorgeCodes19 (approved):
Looks good 👍🏾 Thanks for this. I got to spend to some time learning about scopes and how they differ from methods.

PR #316, opened 2024-10-09
Add help text for caseworker invite fields

tdooner:
## Ticket

N/A - I did this since I keep forgetting the format of these fields, and
figured that caseworkers may appreciate the hint as well.

## Changes

Since we validate the format of these fields, let's provide a visual
hint of the correct format as the field's help text.

**NYC:**
image

**MA:**
image

## Context for reviewers

N/A

## Testing

N/A


tdooner:
@carmenrosalop What do you think about adding placeholders to the caseworker fields that have format validation? (See screenshots above)

tdooner:
@carmenrosalop Updated! New screenshots in the original message above ^

carmenrosalop (approved):
Looks great! Thanks Tom!

GeorgeCodes19 (approved):
Makes sense to me.

PR #317, opened 2024-10-10
1773: Include overtime hours in "Number of hours worked" PDF field

tdooner:
## Ticket

Resolves FFS-1773.

## Changes

* Add `earnings` entries with category="overtime" into the hours calculation.

## Context for reviewers

See the comment in https://jiraent.cms.gov/browse/FFS-1773 for a
longwinded explanation of why this seems like the best way to make the
"Number of hours worked" field align better with reality.

## Testing

Unit tests included for a number of cases I copied from real data. Only the
"earnings" hash is copied from real data, and I anonomyzied it somewhat by
rounding the numbers.

### DIFF
--- a/app/app/controllers/concerns/cbv/payments_helper.rb
+++ b/app/app/controllers/concerns/cbv/payments_helper.rb
@@ -23,13 +23,10 @@ module Cbv::PaymentsHelper
 
   def parse_payments(payments)
     payments.map do |payment|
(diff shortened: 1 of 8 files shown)


GeorgeCodes19 (approved):
It is possible to use these stubs with the Capybara integration once we enable network requests?

tdooner:
> It is possible to use these stubs with the Capybara integration once we enable network requests?

Possibly! Not sure what those stubs will look like. It would be nice to be able to, for sure!

PR #319, opened 2024-10-11
1857: Add email open/deliver/bounce events to NewRelic

tdooner:
## Ticket

Resolves FFS-1857.

## Changes

This commit updates our email terraform to forward email delivery events
to NewRelic via EventBridge. (We don't need an SNS topic, as a result.)

The Email events are in NewRelic as custom events named "AWSSESEvent",
broken down with attributes for the SES event type and the message ID,
so we can make a funnel.

## Context for reviewers

Unfortunately, we won't be able to tie these to the CBV Flow or
invitation just yet, but perhaps we can add a custom email header.

If we want to tie these in with CBV Flow or invitations, we could
instead send these to our application as HTTP webhooks using
EventBridge (by duplicating the code here).

## Testing

Tested by running terraform locally to deploy this to our demo environment.
Then I sent a few emails to myself, and it seems to work!

### DIFF
--- a/infra/app/service/main.tf
+++ b/infra/app/service/main.tf
@@ -258,6 +258,8 @@ module "email" {
   source             = "../../modules/email"
   hosted_zone_domain = local.network_config.domain_config.hosted_zone
   domain             = local.service_config.domain_name
+  newrelic_account_id = "4619676"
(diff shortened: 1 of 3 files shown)


tdooner:
Cool, thanks for de-linting it for me @allthesignals .

PR #321, opened 2024-10-16
1488: Reset PDF margins back to 10

allthesignals:
## Ticket

Addresses 1488

## Changes

Sets back to the default. The previous size was too large. 

This _did_ fix a minor issue where defaults were apparently ignored in production.

## Context for reviewers

It doesn't address any of the substance of 1488, unfortunately.


tdooner (approved):

tdooner:
Merging despite test failures because the test failures are either addressed on a different branch (Bundle audit) or will be addressed separately (anchore-scan).

PR #325, opened 2024-10-16
Fix permissions on Github Actions token

tdooner:
We need to give the deploy workflow `write` permission so it can post a
tag after the deploy finishes. I had forgotten to do it in my last commit:
https://github.com/DSACMS/iv-cbv-payroll/pull/315


tdooner:
Going to self-merge to try it and see if it works. If anyone has feedback, happy to address as a follow-up.

PR #326, opened 2024-10-17
Importing more Spanish Translations via Rake Task

GeorgeCodes19:
## Ticket

Resolves [FFS-1908](https://jiraent.cms.gov/browse/FFS-1908)
Resolves [FFS-1892](https://jiraent.cms.gov/browse/FFS-1892)

## Changes
- Renamed SpanishTranslationService -> TranslationService
- Created a rake task from the TranslationService
- Updated the TranslationService to be more generic and work with multiple languages
- Added a locale import meta file so that we can track when locale imports are run
- Ensuring that the locale parameter passed in the URL matches the supported locales

## Testing

`bundle exec translations:import[es,true]`

Screenshot 2024-10-17 at 12 08 29 AM

### Example txt output
```
Import Date: 2024-10-18 13:44:48 -0400
Overwrite Mode: true
Total Entries: 323
Successfully Imported: 15
Empty Rows: 0
Skipped Rows: 290
Failed Imports: 3
Collisions: 0

Failed Imports Details:
  - Key: en.cbv.summaries.show.payment_details.show.pay_date, Reason: English key does not exist
  - Key: en.cbv.summaries.show.pay_date, Reason: English key does not exist
  - Key: en.cbv.summaries.show.report_details_table_header, Reason: English key does not exist

Collisions Details:
```


app/app/services/translation_service.rb
```diff
@@ -0,0 +1,169 @@
...
+    translation_value = (row[@locale.to_sym]).to_s.strip
+
+    if translation_value.empty?
+      @results[:empty_row_count] += 1
+      return
+    end
+
+    lookup_key = translation_key.sub(/^en\./, "")
+
+    unless @existing_translations["en"].dig(*lookup_key.split("."))
```
GeorgeCodes19, line 63: This forces us to ensure that there's an `en` translation key for the equivalent in another language. I'm not opposed to importing entries even though there aren't corresponding `en` entries.

app/app/services/translation_service.rb
```diff
@@ -0,0 +1,169 @@
+require "csv"
+require "yaml"
+
+class TranslationService
+  attr_reader :results
+
+  def initialize(locale = "es", overwrite = false)
```
GeorgeCodes19, line 7: I'm not sure if it's best practice to use an object/hash for declaring args. That way may be a bit more explicit.

app/app/services/translation_service.rb
```diff
@@ -0,0 +1,169 @@
...
+  def load_existing_translations
+    YAML.load_file(Rails.root.join("config", "locales", "en.yml"))
+  end
+
+  def load_current_locale_translations
+    file_path = Rails.root.join("config", "locales", "#{@locale}.yml")
+    File.exist?(file_path) ? YAML.load_file(file_path) : { @locale => {} }
+  end
+
+  def generate_metadata_file
```
GeorgeCodes19, line 138: Is this a good idea? It seems cool, but I can see where generating a file could clutter up the codebase. Is the information in the file important enough to warrant persisting it for referencing later? Probably not. Maybe we should just log this info and store a file containing a summary of the job.

app/spec/services/translation_service_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      YAML.load_file(current_locale_translations_path)
+    )
+  end
+
+  def create_mock_csv_file
+    contents = [
+      %w[key en es],
+      [ 'en.test.key1', 'Hello', 'Hola' ],
+      [ 'en.test.key2', 'World', 'Mundo' ],
+      [ 'en.test.key3', 'Skip', '' ]
```
GeorgeCodes19, line 95: Note that the total entries in the test being asserted as `2` on [line 52](https://github.com/DSACMS/iv-cbv-payroll/pull/326/files#diff-a3f8b6eda485e35d909d8fe176fdade7fb4a19760f5c82d481fc04a5680043b6R52) because the test CSV purposely contains an empty value for the `es` column.

app/spec/controllers/application_controller_spec.rb
GeorgeCodes19, line 1: Testing that https://jiraent.cms.gov/browse/FFS-1892 is handled appropriately.

app/app/controllers/application_controller.rb
```diff
@@ -18,7 +18,8 @@ def after_sign_out_path_for(resource_or_scope)
   end
 
   def switch_locale(&action)
-    locale = params[:locale] || I18n.default_locale
+    requested_locale = params[:locale]
+    locale = CbvFlowInvitation::VALID_LOCALES.include?(requested_locale) ? requested_locale : I18n.default_locale
```
GeorgeCodes19, line 22: Resolves https://jiraent.cms.gov/browse/FFS-1892

app/app/services/translation_service.rb
tdooner: Yeah, I'm not a huge fan of checking in to the repo files that aren't going to have a long-term purpose. I think logging it in the terminal (and maybe copying it to the Github PR?) is about as far as this should go.

app/spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+class DummyController < ApplicationController
+  def test_action
+    switch_locale do
+      render plain: I18n.locale.to_s
+    end
+  end
+end
+
+RSpec.describe ApplicationController, type: :controller do
+  controller(DummyController) do
```
tdooner, line 12: This could just be `controller(ApplicationController)` (or even just `controller do`) since the first argument is the base class that this anonymous controller inherits from, and it defaults to the controller currently being tested.

The definition of `DummyController` class shouldn't be necessary

app/spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+  end
+
+  describe '#switch_locale' do
+    before do
+      I18n.default_locale = :en
+      routes.draw do
+        get 'test_action', to: 'dummy#test_action'
+      end
```
tdooner, line 25: Do you need to make this route? I thought that by defining an anonymous controller, rspec will set up the routes for you to just do `get :test_action` without having to make the routes yourself? (Very possible I am wrong about this, though.)

app/spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  def test_action
+    switch_locale do
+      render plain: I18n.locale.to_s
+    end
+  end
+end
+
+RSpec.describe ApplicationController, type: :controller do
+  controller(DummyController) do
+    def test_action
```
tdooner, line 13: Side note but in your anonymous controller here, the `switch_locale` around_action will still be running (since it inherits from ApplicationController, which defines it.) You may want to add a `skip_around_action :switch_locale` just to make sure that the switch_locale that's running in your test action here is the only time that that function is being called.

app/app/services/translation_service.rb
```diff
@@ -0,0 +1,169 @@
...
+        file.puts "    Old Value: #{collision[:old_value]}"
+        file.puts "    New Value: #{collision[:new_value]}"
+      end
+    end
+
+    Rails.logger.info "Metadata file generated: #{metadata_file}"
+    file_contents = File.read(metadata_file)
+    Rails.logger.info file_contents
+  end
+end
```
tdooner, line 169: Overall seems like a nice script. I hope we continue to find it useful! I imagine we will.

tdooner (approved):

app/spec/controllers/application_controller_spec.rb
GeorgeCodes19: Oh I wasn't aware. I'll give it a shot.

app/spec/controllers/application_controller_spec.rb
GeorgeCodes19: Ahh interesting. I assumed that the second call would be memoized by Rails and only called once.

PR #332, opened 2024-10-22
1930: Refactor Synchronizations to Patch to Itself

allthesignals:
## Ticket

Addresses https://jiraent.cms.gov/browse/FFS-1930

## Changes

Instead of having the form issue a GET to the next page in the flow, it patches to itself which redirects to the next page.

This is done to move flow control logic into the controller where we need to determine account status. 

## Testing

Added a test to check for the redirect

### DIFF
--- a/app/spec/controllers/cbv/synchronizations_spec.rb
+++ b/app/spec/controllers/cbv/synchronizations_spec.rb
@@ -5,14 +5,14 @@ RSpec.describe Cbv::SynchronizationsController do
 
   let(:cbv_flow) { create(:cbv_flow) }
 
+  let(:pinwheel_account) { create(:pinwheel_account, cbv_flow: cbv_flow) }
+
   before do
     session[:cbv_flow_id] = cbv_flow.id
   end
 
   describe "#update" do
     it "redirects to the payment details page" do
-      pinwheel_account = create(:pinwheel_account, cbv_flow: cbv_flow)
-
       patch :update, params: { user: { account_id: pinwheel_account.pinwheel_account_id } }
 
       expect(response).to redirect_to(cbv_flow_payment_details_path(user: { account_id: pinwheel_account.pinwheel_account_id }))


app/app/views/cbv/synchronizations/show.html.erb
```diff
@@ -6,7 +6,7 @@
   <p class="usa-prose">
     <%= t(".fetching_payroll_description") %>
   </p>
-  <%= form_with url: next_path, method: :get, class: "display-none", data: { 'cbv-synchronizations-target': "form" } do |f| %>
+  <%= form_with url: cbv_flow_synchronizations_path, method: :patch, class: "display-none", data: { 'cbv-synchronizations-target': "form" } do |f| %>
```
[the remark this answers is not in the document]
tdooner: I think if you omit `url: cbv_flow_synchronizations_path` it will just post back to the current URL/controller.

app/spec/controllers/cbv/synchronizations_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  describe "#update" do
+    it "redirects to the payment details page" do
+      pinwheel_account = create(:pinwheel_account, cbv_flow: cbv_flow)
+
+      patch :update, params: { user: { account_id: pinwheel_account.pinwheel_account_id } }
+
+      expect(response).to redirect_to(cbv_flow_payment_details_path(user: { account_id: pinwheel_account.pinwheel_account_id }))
+    end
+  end
```
tdooner, line 20: yay. thanks.

app/spec/controllers/cbv/synchronizations_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  let(:cbv_flow) { create(:cbv_flow) }
+
+  before do
+    session[:cbv_flow_id] = cbv_flow.id
+  end
+
+  describe "#update" do
+    it "redirects to the payment details page" do
+      pinwheel_account = create(:pinwheel_account, cbv_flow: cbv_flow)
```
tdooner, line 14: Creating this object in a `let(:pinwheel_account) { ... } ` would make it easier for future us to add more tests in this `describe` context.

tdooner (approved):

PR #340, opened 2024-10-30
FFS-1801 - API Response Localization

GeorgeCodes19:
## Ticket

Resolves [FFS-1801](https://jiraent.cms.gov/browse/FFS-1801)

## Changes

- Adds a `pinwheel` namespace to the `config/locales/es.yml` file. These entries are dynamically used and not explicitly called so they're also added to the `config/i18n-tasks.yml` file under the `ignore_missing` and `ignore_unused` entries
- Adds tests for the `view_helper.rb` files. Specifically testing the `translate_pinwheel_value` method
- Augments the existing methods in the `pinwheel_data_component.rb` file. Methods that render variable locale data now use the `translate_pinwheel_value` view helper to resolve to their respective values in a target locale.

## Context for reviewers
The goal of this PR is to provide translations for data returned by the upstream Pinwheel API in respect to the end-user's chosen locale. 

### Before
img_4431_720

### After
Screenshot 2024-10-30 at 2 00 20 PM
Screenshot 2024-10-30 at 2 00 34 PM


app/Gemfile
```diff
@@ -61,7 +61,7 @@ gem "actioncable-enhanced-postgresql-adapter"
 gem "aws-sdk-rails"
 
 # https://www.ruby-lang.org/en/news/2024/08/01/dos-rexml-cve-2024-41123/
-gem "rexml", "~> 3.3.6"
+gem "rexml", "~> 3.3.9"
```
GeorgeCodes19, line 64: Fixing a [vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2024-49761)

app/app/controllers/cbv/payment_details_controller.rb
```diff
@@ -112,36 +112,6 @@ def gross_pay
       .reduce(:+)
   end
 
-  def translate_payment_frequency(frequency)
```
GeorgeCodes19, line 115: This logic is no longer need. It's been replaced by the `ApiResponseLocalizer` which essentially does the same thing, but can be extended any class `*_service.rb`

app/app/services/pinwheel_service.rb
```diff
@@ -142,4 +144,7 @@ def generate_signature_digest(timestamp, raw_body)
   def verify_signature(signature, generated_signature)
     ActiveSupport::SecurityUtils.secure_compare(signature, generated_signature)
   end
+
+  localize_methods :fetch_items, :fetch_accounts, :fetch_paystubs,
```
GeorgeCodes19, line 148: `ApiResponseLocalizer` needs to know which methods it should localize. This is done by specifying `localize_methods` after defining the methods.

app/lib/api_response_localizer.rb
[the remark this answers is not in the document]
GeorgeCodes19: > It seems unusual to do this level of surgical injection of translations in the api service wrapper layer where you'd want to keep things strictly "business logic," no?

I don't think I've worked on any two software systems with the same assumptions, requirements and caveats so I can't agree that the API layer is strictly for business logic. I believe that most HTTP libraries actually offer hooks and event lifecycles to hook into and perform these type of surgical operations. 

A `ViewHelper` could be a great place for this type of work. I don't yet have the Rails paradigms in my head so my mind doesn't go there by default. 

Your last point is what I had in mind...
> That said, while I don't think this level of complexity is yet warranted, this is still a really novel, powerful approach to a situation where we may be dealing with tons of different APIs and lots of translations.

There's been discussion about developing the service so that it can scale. (Keyword scale). We may not need the flexibility to translate between multiple API's, but the next team might, or at the very least we can say that it's something we can support. This seems less to do about complexity and more about placement/timing IMO.

app/config/locales/es.yml
```diff
@@ -303,17 +306,6 @@ es:
...
-    annually: anual
-    bi_weekly: cada dos semanas
-    daily: diario
-    hourly: por hora
-    monthly: mensual
-    per_mile: por milla
-    semi_monthly: dos veces al mes
-    semi_weekly: cada dos semanas
-    variable: variable
-    weekly: semanal
```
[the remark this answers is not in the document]
GeorgeCodes19: I've seen `hourly` in the report somewhere. I think it'll make sense to place a comment above groups of translation entries or use yaml object/array syntax with some better nesting.

tdooner (commented):
After catching up on the discussion here and [in Slack](https://nava.slack.com/archives/C06FC5TPAR3/p1730473891889139?thread_ts=1730473786.015589&cid=C06FC5TPAR3), here's where I come down on this:
* I think it's a good idea to make the `translate_payment_frequency` method more generic and reusable as we're likely to need it short-term for the earnings breakdown ticket ("sick" vs "premium" time).
  * Thanks for taking the initiative to do this!
* This approach is a bit too powerful.
  * It might be the right answer if we had 5+ APIs with many different potential value-returning methods that needed translation. But over the short-to-medium-term, I don't think we're likely to have that many.
  * I think we can achieve the same goal of translating API response values without resorting to metaprogramming.
* I also agree with Matt that keeping the translation logic as close to the view layer is preferable as it allows us to more easily reason about the Pinwheel data.
  * What I mean by this is that by overriding the `fetch_paystubs` method here, we're taking the Pinwheel API response and translating it before it makes it into the controller or view. What if we wanted to do some conditional logic on these (e.g. display a special blurb for "bi-weekly" values)? We wouldn't be able to do so easily, because the value returned by `fetch_paystubs` would either be "biweekly" or "cada dos semanas".

So, here's what I'd propose in terms of a simpler approach that accomplishes your goals:

First, let's keep the YAML file together as multiple YAML files are confusing and make our translation tooling more complicated. We could namespace all the pinwheel key/values underneath a "pinwheel" key:
```yaml
# config/locales/es.yml
pinwheel:
  payment_frequencies:
    annually: anual
    # ...
```

Then, how would we render those? Let's take the `translate_payment_frequency` helper method and make it a bit more generic so it will work with future types of API fields as well. Maybe we call it `translate_pinwheel_value`? Pseudocode would be something like:
```ruby
def translate_pinwheel_value(field, value)
  i18n_key = "pinwheel.#{field}.#{value}"
  # if I18n key exists, render it
  # else, raise error in dev/test ("unknown pinwheel value for #{field}: #{value}") and fall back to raw value in production
end
```

Then, how do we use that helper? We could put it in the view directly, or in the ViewComponent layer. If we put it in the view it would look something like Matt suggested in Slack:
```erb
<%= table.with_data_point(:pay_frequency, translate_pinwheel_value(:payment_frequencies, summary[:income]["pay_frequency"])) %>
```

If we put it in the ViewComponent layer, then we would just pass the raw value from Pinwheel into the `PinwheelDataPoint` component, and the component itself would be responsible for translating it. Something like:
```ruby
# in the PinwheelDataPointComponent
def pay_period_with_frequency(start_date, end_date, pay_frequency)
  translated_pay_frequency = translate_pinwheel_value(:payment_frequencies, pay_frequency)

  {
    label: I18n.t("cbv.summaries.show.pdf.caseworker.pay_period", pay_frequency: translated_pay_frequency),
    value: I18n.t("cbv.payment_details.show.pay_period_value", start_date: format_date(start_date), end_date: format_date(end_date))
  }
end
```

I kind of like the idea of putting it in the ViewComponent, so that all the translation occurs there and we're just passing in the raw Pinwheel data value (`summary[:income]["pay_frequency"]`) into the component.

I know it's not as fancy as the refactor you cooked up here, but I think the simpler approach still accomplishes your goals of enabling scale down the road. When/if we introduce a non-pinwheel API that needs translation, we could again make the pattern more general by removing `pinwheel` from the helper method's name.

app/config/locales/es.yml
```diff
@@ -293,17 +293,21 @@ es:
...
+  pinwheel:
+    deductions:
+      other: Otro
```
tdooner, line 298: Although the ticket explicitly calls out "other" and "employed" as the strings that need translating, it would be good for us to translate all possible values so we don't keep discovering more and more of these untranslated strings down the road. We could do it as a follow-up though, since the translation process is quite frustrating at the moment.

app/app/components/pinwheel_data_point_component.rb
```diff
@@ -72,23 +74,26 @@ def employment_end_date(end_date)
...
+    translated_pay_frequency = translate_pinwheel_value("payment_frequencies", frequency)
     {
       label: I18n.t("cbv.payment_details.show.pay_frequency"),
-      value: frequency
+      value: translated_pay_frequency
     }
   end
 
   def hourly_rate(amount, unit)
+    translated_unit = translate_pinwheel_value("payment_frequencies", unit)
```
tdooner, line 93: nit: It looks like the "payment_frequencies" translation key includes **both** the API values for
* `pay_frequencies` (`bi-weekly daily monthly semi-monthly variable weekly`), and
* `compensation_unit` (`annually bi-weekly daily hourly monthly per_mile semi-monthly semi-weekly variable weekly`)

I see why we did this because there is so much overlap between the two sets of values. That said, I do think this could be a good opportunity to separate the lists of values, just to make it more clear what possible options exist for both API responses.

Not a blocker, but just another idea on making this a bit more understandable. (Also, should we make "payment_frequencies" match the Pinwheel field name `pay_frequencies`?)

app/app/components/pinwheel_data_point_component.rb
```diff
@@ -72,23 +74,26 @@ def employment_end_date(end_date)
   end
 
   def employment_status(status)
+    translated_status = translate_pinwheel_value("employment_statuses", status)
```
tdooner, line 77: I really like how this reads. Thanks for incorporating my feedback!

app/spec/helpers/view_helper_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+RSpec.describe ViewHelper, type: :helper do
+  describe '#translate_pinwheel_value' do
+    # Store the original locale before all tests
+    let(:original_locale) { I18n.default_locale }
+
+    # Reset locale after each test
+    after { I18n.locale = original_locale }
+
+    context 'when locale is :es' do
+      before { I18n.locale = :es }
```
tdooner, line 12: An easier way to set the locale for tests within a context would be to use an `around` block:

```ruby
around do |ex|
  I18n.with_locale(:es, &ex)
end
```

If you did it that way, you could get rid of the `original_locale` setup and `after` teardown.

app/spec/helpers/view_helper_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      it 'returns the translated value if translation exists' do
+        I18n.backend.store_translations(:es, {
+          pinwheel: {
+            namespace: {
+              existing_value: 'Translated Value'
+            }
+          }
+        })
```
tdooner, line 21: TIL about `store_translations` - cool. This is clever (I think I would have just stubbed I18n.t when called with `pinwheel.namespace.existing_value`). The downside of using this method is probably that these tests will remain stored in the I18n backend even after this test case finishes running. Not a big deal though, since I don't think it's likely to break anything else.

tdooner (approved):
Lots of nitpicks/additional feedback (of course...) but I think this is looking very good and could even be merged as-is.

app/config/i18n-tasks.yml
```diff
@@ -134,6 +134,8 @@ ignore_missing:
     - "cbv.summaries.show.pdf.caseworker.*"
     # Applicant invitation reminder emails taht are only supported in English:
     - "applicant_mailer.invitation_reminder_email.*"
+    # Pinwheel values that are only supported in English:
+    - "pinwheel.*"
```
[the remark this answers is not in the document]
GeorgeCodes19: hmmm... I do like the idea of being able to alter what we get back from Pinwheel even when `en` is set as the locale. I'll add these entries to the `en.yml` file.

app/app/helpers/view_helper.rb
```diff
@@ -38,4 +35,26 @@ def format_view_datetime(timestamp_string)
   def format_money(dollars_in_cents)
     number_to_currency(dollars_in_cents.to_f / 100)
   end
+
+  def translate_pinwheel_value(namespace, value)
+    # if the key isn't in spanish, just return the original value
+    return value unless I18n.locale == :es
```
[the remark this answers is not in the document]
GeorgeCodes19: Great, feedback. Implemented.

app/config/locales/es.yml
GeorgeCodes19: Yeah, let's pull all the possible values and ship them for translation in the next batch. I'll go ahead and add them to `en.yml` so that we don't have to track them down in Pinwheel.

PR #346, opened 2024-11-01
Track manually switching locale

allthesignals:
## Ticket

N/A

## Changes

Track when language is manually switched

## Context for reviewers

Based on this conversation: https://nava.slack.com/archives/C0700F01KKP/p1730478392526959

## Testing

image

### DIFF
--- a/app/app/views/application/_header.html.erb
+++ b/app/app/views/application/_header.html.erb
@@ -51,7 +51,7 @@
           <% if show_translate_button? %>
             <% I18n.available_locales.each do |locale| %>
               <% next if locale == I18n.locale %>
-              <li class="usa-nav__primary-item">
+              <li class="usa-nav__primary-item" onclick="window.NREUM && window.NREUM.addPageAction('LanguageManuallySwitched', { requested_locale: <%= locale %> })">
                 <%= switch_locale_link(locale) %>
               </li>
             <% end %>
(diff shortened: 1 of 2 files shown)


erynsobing:
curious if this would this track the page in which the user toggled the language button? i.e. if i query the logs, is there an attribute that'll tell me which page the event was triggered in

app/app/views/application/_header.html.erb
```diff
@@ -51,7 +51,7 @@
           <% if show_translate_button? %>
             <% I18n.available_locales.each do |locale| %>
               <% next if locale == I18n.locale %>
-              <li class="usa-nav__primary-item">
+              <li class="usa-nav__primary-item" onclick="window.NREUM && window.NREUM.addPageAction('LanguageManuallySwitched', { requested_locale: <%= locale %> })">
```
tdooner, line 54: I'm getting lost in all the levels of interpolation here but I think it needs to be JSON in there...

(I might also consider calling the attribute "new_locale"?)

```suggestion
              <li class="usa-nav__primary-item" onclick="window.NREUM && window.NREUM.addPageAction('LanguageManuallySwitched', { new_locale: <%= JSON.generate(locale) %> })">
```

tdooner (changes requested):
We probably should

PR #348, opened 2024-11-05
FFS-1836: Email invitation reminders

millerti:
## Changes 
Added a new invitation reminder email that is triggered by a cron job at 9am ET every morning. This will send an email reminder for any CBV invitations that are older than 3 days, have not been completed, have not expired, and have not already had a reminder sent. Check out [the jira ticket](https://jiraent.cms.gov/browse/FFS-1836) and [the figma](https://www.figma.com/design/qzbsxwS2wsouKK1JmRtn2H/Consent-Based-Verification?node-id=3935-3570&node-type=canvas&t=P4BToCiQTAmb2AwV-0) for more detail on what the email should look like.

### Check that the email sends and looks ok
In order to see emails when running locally, follow the steps below. You may want to run bin/rails db:reset before you start your testing, since the reminder sender will iterate over all existing CbvFlowInvitations and may choke on old or bad data.
1. Add to gemfile:
	gem 'letter_opener'
	gem 'launchy'
2. Get a fresh CbvFlowInvitation with decent data by sending an invite locally
3. In rails c, run the following commands:
	a = CbvFlowInvitation.last
	a.created_at = 4.days.ago
	a.updated_at = 4.days.ago
	a.save!
4. You will need to comment out this line in cbv_flow_invitation.rb: before_validation :parse_snap_application_date (this prevents saving the modified record... not sure why)
5. In development.rb, add this line: config.action_mailer.delivery_method = :test
4. Optionally, remove the 'invitation.update(invitation_reminder_sent_at: DateTime.now)' at the end of invitation_reminders.rake. This is what prevents the emails from being sent multiple times (which you should test), but if you just need to see the email, removing this line is convenient.
6. Add the command 'Launchy.open ActionMailer::Base.deliveries.last.message.to_s' after the ApplicantMailer line in invitation_reminders.rake
7. At the command line, call 'rake invitation_reminders:send_all'
Verify that a browser pops up with the rendered email and that it looks like the [desired layout in the figma](https://www.figma.com/design/qzbsxwS2wsouKK1JmRtn2H/Consent-Based-Verification?node-id=3935-3570&node-type=canvas&t=P4BToCiQTAmb2AwV-0).

### Verify the cron job
This rake task is scheduled to run once a day at 9am Eastern (or 10am during DST). This is configured in scheduled_jobs.tf. Verify that the configuration looks correct. Tom says this is very difficult to test locally, so I think this may be the best we can do until this is in the Demo environment.

Once in demo, also make sure that the NewRelic email sent event is being logged and can be distinguished from other email sent events (eg you can tell how many reminders are being sent as compared to original invite emails).


app/spec/mailers/applicant_mailer_spec.rb
```diff
@@ -1,71 +1,147 @@
...
+    it "has the correct subject line for ma" do
+      invitation.update!(site_id: "ma")
+      Rake::Task['invitation_reminders:send_all'].execute
+      email = ActionMailer::Base.deliveries.last
+
+      expect(email.subject).to eq("Reminder: Verify your income to renew your SNAP benefits")
+    end
+
+    it "has the correct subject line for nyc" do
+      invitation.update!(site_id: "nyc", case_number: "00012345678A", client_id_number: "AB00000C", snap_application_date: invitation_sent_at - 30.minutes)
```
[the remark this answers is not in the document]
millerti: It's because we used the factory to create a default version, and now it lacks some of the required things. We're reusing the same invite for all the tests, essentially.

app/app/views/applicant_mailer/invitation_reminder_email.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <h3>
+      <span class="text-primary"><%= t("shared.pilot_name") %> | </span>
+      <% if current_site.id == "ma" %>
+        <%= image_tag "dta_logo.png", class: "text-middle", style: "width: 140px" %>
+      <% elsif current_site.id == "nyc" %>
+        <%= image_tag "hra_logo.png", class: "text-middle", style: "width: 170px" %>
+      <% else %>
+        <%= site_translation("shared.header.cbv_flow_title") %>
+      <% end %>
+    </h3>
```
tdooner, line 15: Hmm, this duplicated code is unfortunate (copied from the invitation email itself). I don't feel so strongly as to block the PR but it might be nice to DRY it up by creating an email layout for applicant-facing emails.

app/app/views/applicant_mailer/invitation_reminder_email.html.erb
```diff
@@ -0,0 +1,43 @@
...
+
+      <p class="margin-y-3">
+        <a href="<%= @cbv_flow_invitation.to_url %>" class="usa-button width-auto" type="button">
+          <%= t(".verify_income_button" ) %>
+        </a>
+      </p>
+      <p><%= t(".instructions", agency_acronym: current_site.agency_short_name) %></p>
+      <h2><%= t(".feedback_header") %></h2>
+      <p>
+        <a href="<%= "https://docs.google.com/forms/d/1E65H8Bcl3nYL8_OS5uwOqeN6p9G4aBXth6aHDoLlIoc/viewform?edit_requested=true" %>" class="usa-button usa-button--outline width-auto" type="button">
```
tdooner, line 31: Did this have to be an ERB interpolation for any particular reason? I think it'd be a bit more clear as simply:

```suggestion
        <a href="https://docs.google.com/forms/d/1E65H8Bcl3nYL8_OS5uwOqeN6p9G4aBXth6aHDoLlIoc/viewform" class="usa-button usa-button--outline width-auto" type="button">
```

app/config/locales/en.yml
```diff
@@ -48,6 +48,25 @@ en:
       subject:
         default: Verify your income for your SNAP application or renewal
         ma: Verify your income to renew your SNAP benefits
+    invitation_reminder_email:
+      agency_acronym:
+        default: DTA
+        nyc: NYC
+      agency_name:
+        default: Department of Transitional Assistance
+        nyc: Human Resources Administration Department of Social Services
```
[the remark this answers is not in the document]
tdooner: Yeah, we do have the acronyms at least in site-config.yml, and (slightly-different) agency names as well.

That said, I've been thinking we should just standardize on putting them in the I18n files, though. It will facilitate translation of them and be more intuitive since they are strings being used in views.

That said, some concrete feedback here:
1. For the acronyms, just use the `current_site.agency_acronym` in the view. We can move that to the I18n files later.
2. For the `agency_name`, we already have a variant of them in the `shared:` section of the I18n files (`shared.agency_full_name`). If we could use those, even though they are slightly different, that would be ideal.
3. If we keep these here, let's not use "default" for the MA strings, as it will appear like the invitations are from MA when we're sending invitations in the sandbox site. I'd like us to standardize on using "default" only when there is a specific override for a single state. So, change "default" to "ma" and then add a "sandbox" key as well that is "CBV Test Agency".

tdooner (approved):

app/app/views/applicant_mailer/invitation_reminder_email.html.erb
millerti: No, I think it's fine to leave that out

PR #353, opened 2024-11-15
No ticket: deduplicated expired event, refactored ClickedCBVInvitationLink slightly

millerti:
## Ticket
No ticket. Fixes to work done in [FFS-1839](https://jiraent.cms.gov/browse/FFS-1839)

## Changes
Made ApplicantLinkExpired unique by cbv_flow_id.
Refactored ClickedCBVInvitationLink to its own function to simplify set_cbv_flow.

## Testing
Trigger both changed events and verify that they fire correctly and that the ApplicantLinkExpired event contains a valid cbv_flow_id. 

You can easily expire a link by setting its created_at date to 15.days.ago.


app/app/controllers/cbv/base_controller.rb
```diff
@@ -14,22 +14,20 @@ def set_cbv_flow
       if invitation.blank?
         return redirect_to(root_url, flash: { alert: t("cbv.error_invalid_token") })
       end
+
+      @cbv_flow = CbvFlow.create_from_invitation(invitation)
+      session[:cbv_flow_id] = @cbv_flow.id
```
millerti, line 19: I moved these lines up so I could use the cbv_flow_id earlier (line 22). But the unit tests that touch the expired events page explicitly test that session[:cbv_flow_id] is unchanged, and of course now it has changed. So I wonder: is there some important reason to test this and I'm breaking something unintentionally by having the session[:cbv_flow_id] set on the expired page? Or was this just something we could conveniently unit test, and I can get rid of it?

app/app/controllers/cbv/base_controller.rb
```diff
@@ -96,4 +92,25 @@ def track_timeout_event
   rescue => ex
     Rails.logger.error "Unable to track NewRelic event (ApplicantTimedOut): #{ex}"
   end
+
+  def track_expired_event(cbv_flow_id)
+    NewRelicEventTracker.track("ApplicantLinkExpired", {
+      cbv_flow_id: cbv_flow_id,
```
tdooner, line 98: Looks like you're passing in the invitation's ID into the method, but calling it `cbv_flow_id`. I'd recommend passing in both as you've done in the other methods.

tdooner (changes requested):

tdooner (approved):

app/app/controllers/cbv/base_controller.rb
tdooner: Ah true, there is no cbv_flow_id if the invitation is expired. The invitation_id is great, thanks.

PR #360, opened 2024-11-19
1782: Capture User Agent and other data per page

allthesignals:
## Ticket

Resolves https://jiraent.cms.gov/browse/FFS-1782

## Changes

Captures browser, device, device type along with flow id and invitation id per page view for applicants only

## Context for reviewers

Added a new gem

## Testing
image

### DIFF
--- a/app/spec/spec_helper.rb
+++ b/app/spec/spec_helper.rb
@@ -99,6 +99,11 @@ RSpec.configure do |config|
   # as the one that triggered the failure.
   Kernel.srand config.seed
 
+  # Stub out NewRelicEventTracker so it doesn't make calls in test
+  config.before(:each) do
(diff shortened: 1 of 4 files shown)


app/spec/spec_helper.rb
```diff
@@ -99,6 +99,11 @@
   # as the one that triggered the failure.
   Kernel.srand config.seed
 
+  # Stub out NewRelicEventTracker so it doesn't make calls in test
+  config.before(:each) do
+    allow(NewRelicEventTracker).to receive(:track)
```
GeorgeCodes19, line 104: oh, I like this pattern.

GeorgeCodes19 (approved):

PR #364, opened 2024-11-21
Fix test for invitation reminder language

tdooner:
## Ticket

N/A

## Changes

* Fix test for invitation reminder language

## Context for reviewers

The previous test was never setting the language of the invitation to Spanish, so of course it was going to send in English.

## Testing

It, itself, is a test.

### DIFF
--- a/app/spec/mailers/applicant_mailer_spec.rb
+++ b/app/spec/mailers/applicant_mailer_spec.rb
@@ -144,10 +144,16 @@ RSpec.describe ApplicantMailer, type: :mailer do
       Rake::Task['invitation_reminders:send_all'].execute
     end
 
-    it "renders the subject and body in English because Spanish is not yet supported" do
-      Rake::Task['invitation_reminders:send_all'].execute
-      email = ActionMailer::Base.deliveries.last
-      expect(email.subject).to eq(I18n.t('applicant_mailer.invitation_reminder_email.subject.ma'))
+    context "for an invitation in Spanish" do
+      before do
+        invitation.update(language: "es")
+      end
+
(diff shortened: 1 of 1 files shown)


app/spec/mailers/applicant_mailer_spec.rb
```diff
@@ -144,10 +144,16 @@
       Rake::Task['invitation_reminders:send_all'].execute
     end
 
-    it "renders the subject and body in English because Spanish is not yet supported" do
-      Rake::Task['invitation_reminders:send_all'].execute
-      email = ActionMailer::Base.deliveries.last
-      expect(email.subject).to eq(I18n.t('applicant_mailer.invitation_reminder_email.subject.ma'))
+    context "for an invitation in Spanish" do
+      before do
+        invitation.update(language: "es")
```
millerti, line 149: Is there any reason to think that you need to set it back to "en" for the other tests, since it randomizes their order? Or does context somehow protect from that?

millerti (commented):
Tests pass. Nice catch!

app/spec/mailers/applicant_mailer_spec.rb
tdooner: Good question! Because [we use transactional fixtures](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/spec/rails_helper.rb#L47-L50), all tests run in separate database transactions (which get rolled back after every test). There's some Rails docs [here](https://edgeguides.rubyonrails.org/testing.html#transactions) and rspec docs [here]( https://rspec.info/features/6-0/rspec-rails/Transactions/).

PR #370, opened 2024-11-21
No ticket: added two missing Spanish translations for reminder emails

millerti:
## Changes
Two Spanish strings were not translated in the NYC reminder email:
image

Now, it will look like this. Note that, after discussion, we decided NOT to translate the header string, since it's used in multiple places (not just these emails), matches the URL, and is too long in Spanish.
image

## Testing
Take a look at es.yml and convince yourself that all Spanish translations are now present.

Once this is merged and in demo, trigger a reminder email by sending a new invite to your own email in Spanish, then call the following:
last = CbvFlowInvitation.last
last.created_at = 4.days.ago
last.snap_application_date = 1.hour.ago
last.save!
ApplicantMailer.with(cbv_flow_invitation: last).invitation_reminder_email.deliver_now

Verify the email you receive is completely in Spanish.

### DIFF
--- a/app/config/locales/es.yml
+++ b/app/config/locales/es.yml
@@ -388,7 +388,7 @@ es:
       fr: Francés
       zh: Chino
     not_applicable: N/C
-    pilot_name: Programa piloto de ingresos de SNAP
+    pilot_name: SNAP Income Pilot
     skip_link: Saltar al contenido principal
   time:
     formats:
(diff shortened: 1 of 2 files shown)


tdooner:
Could you provide a screenshot of the "after", both of the email and of the web layout? The translated product name strikes me as very long and potentially likely to break the styling.

millerti:
> Could you provide a screenshot of the "after", both of the email and of the web layout? The translated product name strikes me as very long and potentially likely to break the styling.

Good catch. Added the shot... you were right! Not sure what we want to do about it though.

tdooner (approved):

PR #372, opened 2024-11-21
FFS-2104 - Database Model (CbvClient) for PII Fields

GeorgeCodes19:
## Ticket

Resolves [FFS-2104](https://jiraent.cms.gov/browse/FFS-2104)

## Changes

- [x] Create initial `CbvClient` model and migration
- [x] Set up the model associations (`belongs_to`)
- [x] Create data migration script to move over the data
- [x] Update the invitation sending to write to both the Invitation **and** the cbv client table

## Testing
Unit tests were added to this PR. No user testing is needed at the moment.


app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
```diff
@@ -39,6 +39,9 @@ def create
       return render :new
     end
 
+    # hydrate the cbv_client with the invitation if there are no cbv_flow_invitation errors
+    @cbv_client = CbvClient.create_from_invitation(@cbv_flow_invitation)
```
GeorgeCodes19, line 43: I moved this lower in the call-stack after the `CbvFlowInvitation` data has been validated because passing nil to create a `cbv_flow_invitation` in one of our tests caused the `NOT NULL` constraint violations to propagate up to the view layer (since hydrating the `CbvClient` record and creating a new `CbvFlowInvitation` were in the same begin/rescue block.

app/app/models/cbv_client.rb
```diff
@@ -0,0 +1,21 @@
...
+      first_name: cbv_flow_invitation.first_name,
+      middle_name: cbv_flow_invitation.middle_name,
+      last_name: cbv_flow_invitation.last_name,
+      agency_id_number: cbv_flow_invitation.agency_id_number,
+      client_id_number: cbv_flow_invitation.client_id_number,
+      snap_application_date: cbv_flow_invitation.snap_application_date,
+      beacon_id: cbv_flow_invitation.beacon_id
+    )
+    cbv_flow_invitation.cbv_client = result
+    cbv_flow_invitation.save(validate: false)
```
GeorgeCodes19, line 17: I tried several different approaches including saving the `cbv_flow_invitation` to the `CbvClient` within this `create` call, but ultimately reverted back to this.

GeorgeCodes19: This has been fixed! The issue was due to the `snap_application_date` being in ISO 8601 format by default after the `CbvFlowInvitation` has been saved and queried from the database.

app/spec/models/cbv_client_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      cbv_client = CbvClient.create_from_invitation(cbv_flow_invitation)
+      cbv_flow_invitation.reload
+      expect(cbv_flow_invitation.cbv_client).to eq(cbv_client)
+    end
+
+    context 'when cbv_flow_invitation is nil' do
+      it 'raises an error' do
+        expect {
+          CbvClient.create_from_invitation(nil)
+        }.to raise_error(NoMethodError)
```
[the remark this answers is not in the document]
GeorgeCodes19: Agreed. Deleted

app/app/models/cbv_client.rb
```diff
@@ -0,0 +1,19 @@
...
+      case_number: cbv_flow_invitation.case_number,
+      first_name: cbv_flow_invitation.first_name,
+      middle_name: cbv_flow_invitation.middle_name,
+      last_name: cbv_flow_invitation.last_name,
+      agency_id_number: cbv_flow_invitation.agency_id_number,
+      client_id_number: cbv_flow_invitation.client_id_number,
+      snap_application_date: cbv_flow_invitation.snap_application_date,
+      beacon_id: cbv_flow_invitation.beacon_id
+    )
+    cbv_flow_invitation.update!(cbv_client: client)
```
GeorgeCodes19, line 16: Explicitly throwing an exception if `update` fails- enforcing the coupling between `CbvClient` and `CbvFlowInvitation`

tdooner:
Looks good, thanks for fixing the bug.

PR #373, opened 2024-11-25
FFS-1555: Various copy updates

millerti:
## Changes
Two small fixes to copy:
1. On the Caseworker Invite page, the wording about the date now differs by state.
2. On the Client PDF for MA, we now specify DTA where we didn't before.

## Testing
1. In MA, NYC, and Sandbox, navigate to the Caseworker Invite page and make sure that this sentence has been updated as proposed in the ticket:
image

2. Access the Income Verification Report PDF and verify that the agency's short name is now used instead of saying "your SNAP agency":
image


app/config/locales/en.yml
```diff
@@ -82,7 +82,10 @@ en:
...
+        description_html:
+          ma: "<p>Provide some details about the client so we can send them a link to verify their pay information. We'll request the past %{pay_income_days} days of income, based on today's date.</p><p>If multiple members of the household earn income, send an invite link to each household member.</p>"
+          nyc: "<p>Provide some details about the client so we can send them a link to verify their pay information. We'll request the past %{pay_income_days} days of income, based on the client's application/recertification date.</p><p>If multiple members of the household earn income, send an invite link to each household member.</p>"
+          sandbox: "<p>Provide some details about the client so we can send them a link to verify their pay information. We'll request the past %{pay_income_days} days of income, based on the client's application/recertification date.</p><p>If multiple members of the household earn income, send an invite link to each household member.</p>"
```
tdooner, line 88: Side note, but I think for these site-specific strings where we have only one customization, we should use the "default" key along with whichever site required the customization. E.g. in this case

```suggestion
        description_html:
          default: "<p>Provide some details about the client so we can send them a link to verify their pay information. We'll request the past %{pay_income_days} days of income, based on the client's application/recertification date.</p><p>If multiple members of the household earn income, send an invite link to each household member.</p>"
          ma: "<p>Provide some details about the client so we can send them a link to verify their pay information. We'll request the past %{pay_income_days} days of income, based on today's date.</p><p>If multiple members of the household earn income, send an invite link to each household member.</p>"

```

app/config/locales/en.yml
```diff
@@ -341,7 +344,10 @@ en:
             client_report_information: Client and Report Information
             date_created: Date the income report was created
             date_range: Date range for the income report
-            description: We have collected your income information with your consent and sent it to your SNAP agency. Any additional income that you weren't able to add to this report should be shared separately with your SNAP agency.
+            description:
+              ma: We have collected your income information with your consent and sent it to %{agency_acronym}. Any additional income that you weren't able to add to this report should be shared separately with %{agency_acronym}.
+              nyc: We have collected your income information with your consent and sent it to %{agency_acronym}. Any additional income that you weren't able to add to this report should be shared separately with %{agency_acronym}.
+              sandbox: We have collected your income information with your consent and sent it to your SNAP agency. Any additional income that you weren't able to add to this report should be shared separately with your SNAP agency.
```
tdooner, line 350: I'm trying to spot the difference here - is it just "your SNAP agency" vs "%{agency_acronym}"? Even sandbox has an agency acronym, so you could just make the string take that variable in all cases. It would read a bit weird in sandbox ("shared separately with CBV") but that's ok.

The more strings we keep as non-site-specific, the happier our translation process will be (and the easier it is to maintain), so slight bias towards keeping this non-site-specific if possible. (We can still use the variable in a non-site-specific string, to be clear)

app/app/views/cbv/summaries/show.pdf.erb
```diff
@@ -26,7 +26,7 @@
     </div>
   </div>
 <% else %>
-  <p><%= t(".pdf.client.description") %></p>
+  <p><%= t(".pdf.client.description", agency_acronym: current_site.agency_short_name) %></p>
```
tdooner, line 29: Whoops:
image

Either take my suggestion below to keep it a non-site-specific translation, or:

```suggestion
  <p><%= site_translation(".pdf.client.description", agency_acronym: current_site.agency_short_name) %></p>
```

tdooner (changes requested):

app/config/locales/en.yml
millerti: I think you're right, taking this suggestion.

tdooner (approved):
Looks good, thanks for fixing:

image

PR #383, opened 2024-12-04
2113: Redirect to root if staff portal is disabled

allthesignals:
## Ticket

Resolves [FFS-2113](https://jiraent.cms.gov/browse/FFS-2113).

## Changes
Redirects to the root with a flash if the staff protal is disabled.

## Context for reviewers
Adds an field, `staff_portal_enabled`, to the site config.

Adds the hook in the base controller for the caseworker controllers.

## Acceptance testing

image
image

### DIFF
--- a/app/spec/controllers/caseworker/entries_controller_spec.rb
+++ b/app/spec/controllers/caseworker/entries_controller_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe Caseworker::EntriesController do
       it "should show ma specific copy with a link to /sso/ma" do
         agency_short_name = site_config["ma"].agency_short_name
         get :index, params: { site_id: "ma" }
-        expect(response).not_to be_successful
+        expect(response).to redirect_to(root_url)
       end
     end
 


app/config/site-config.yml
```diff
@@ -10,6 +10,7 @@
   transmission_method: shared_email
   transmission_method_configuration:
     email: <%= ENV['NYC_HRA_EMAIL'] %>
+  staff_portal_enabled: false
```
tdooner, line 13: We want to disable the one for Massachusetts, not NYC.

app/db/schema.rb
```diff
@@ -70,8 +70,8 @@
     t.string "site_id"
     t.string "confirmation_code"
     t.datetime "transmitted_at"
-    t.datetime "redacted_at"
     t.datetime "consented_to_authorized_use_at"
+    t.datetime "redacted_at"
```
tdooner, line 74: Might be worth running `bin/rails db:reset` locally on the main branch, to make sure you don't continue to be plagued by this column being out of order.

app/config/site-config.yml
tdooner: Glad we got it resolved, sorry for the confusion.

app/db/schema.rb
tdooner: Were you going to remove this change from the PR?

app/config/site-config.yml
```diff
@@ -31,6 +32,7 @@
   transmission_method_configuration:
     bucket: <%= ENV['MA_DTA_S3_BUCKET'] %>
     public_key: <%= ENV['MA_DTA_S3_PUBLIC_KEY'].inspect %>
+  staff_portal_enabled: true
```
tdooner, line 35: "shut it down" says Makaela: https://nava.slack.com/archives/C06FC5TPAR3/p1733336416266519?thread_ts=1733330841.072599&cid=C06FC5TPAR3

(MA only)

app/app/controllers/caseworker/entries_controller.rb
```diff
@@ -1,5 +1,15 @@
...
     @current_site = current_site
   end
+
+  def redirect_if_disabled
+    unless current_site.staff_portal_enabled
+      redirect_to root_url, flash: {
+        slim_alert: { message: "Staff portal disabled", type: "error" }
+      }
+    end
+  end
```
tdooner, line 14: I would have expected this logic to be in the base controller so it would disable the other pages within the portal as well. But, ok, I suppose this is fine since the session timeout it so short that there's no way there are open sessions still on the caseworker dashboard.

app/app/controllers/caseworker/entries_controller.rb
```diff
@@ -1,5 +1,15 @@
...
+  before_action :redirect_if_disabled
+
   def index
     @current_site = current_site
   end
+
+  def redirect_if_disabled
+    unless current_site.staff_portal_enabled
+      redirect_to root_url, flash: {
+        slim_alert: { message: "Staff portal disabled", type: "error" }
```
tdooner, line 11: Let's use i18n for this kind of stuff, even if we don't need to translate it, just so all our product strings are in one place.

tdooner:
Also, please choose an acceptance testing method.

app/spec/controllers/caseworker/entries_controller_spec.rb
```diff
@@ -27,5 +24,17 @@
         expect(unescaped_body).to include("Log in with your LAN ID")
       end
     end
+
+    context "when state is disabled" do
+      before do
+        stub_site_config_value("sandbox", "staff_portal_enabled", false)
+      end
+
+      it "should show redirect to the root page" do
```
tdooner, line 33: nit: https://www.betterspecs.org/#should

e.g.
```suggestion
      it "redirects to the root page" do
```

app/spec/controllers/caseworker/entries_controller_spec.rb
```diff
@@ -27,5 +24,17 @@
...
+        agency_short_name = site_config["sandbox"].agency_short_name
+        get :index, params: { site_id: "sandbox" }
+        expect(response).not_to be_successful
```
tdooner, line 36: nit: A better expectation would be something like

```suggestion
        expect(response).to redirect_to(root_url)
```

app/spec/support/test_helpers.rb
```diff
@@ -88,4 +88,9 @@ def stub_identities(account_id = SecureRandom.uuid)
       }
     end
   end
+
+  def stub_site_config_value(site, key, value)
+    site_config = Rails.application.config.sites[site]
+    allow(site_config).to receive(key.to_sym).and_return(value)
+  end
```
tdooner, line 95: Nice, thanks for making this helper.

tdooner (approved):

PR #384, opened 2024-12-04
FFS-2014 - [bugfix] Updating Backfills task for invalid entries

GeorgeCodes19:
## Ticket

Resolves [FFS-2104](https://jiraent.cms.gov/browse/FFS-2104).

## Changes

- Replaced `CbvFlowInvitation.update` -> `CbvFlowInvitation.update_column`
- Moved the throw error clause to the `create` method of the `CbvClient`

## Context for reviewers
Environments other than the test environment may have `CbvFlowInvitation` records which may contain fields that fail validation. This includes redacted data and stale `snap_application_date` (past the agency's acceptable date). This PR aims to fulfill backfilling data by first creating the `CbvClient` and using the `update_column` derived from ActiveRecord.

## Acceptance testing


app/spec/lib/backfills_spec.rb
```diff
@@ -7,14 +7,55 @@
...
+          snap_application_date: 31.days.ago.to_date
+        })
+        invitation.save(validate: false)
+        invitation.redact!
+        invitation
+      end
+
+      let(:valid_cbv_flow_invitation) do
+        invitation = create(:cbv_flow_invitation)
+        invitation.redact!
```
tdooner, line 40: Aren't redacted invitations actually invalid, due to the date being wrong? (so this is a misnomer - it should be named should be something like `redacted_cbv_flow_invitation`)

It seems like there were two issues (the backfill failing with both redacted invitations, as well as invalid non-redacted invitations), you could test both independently by removing the `redact!` from the other test case.

tdooner (approved):
Seems like a good fix, but, of course, one suggestion for making the test a bit more clear.

PR #386, opened 2024-12-11
FFS-2211: guardrails for nil pinwheel_event

millerti:
## Changes
Updated synchronizations redirect logic to fail gracefully if we can't find a pinwheel_account
Updated turbo-rails so we incorporate a fix that will help our webserver not crash when this happens

## Context for reviewers
This is part of a response to a downtime event we encountered on 12/5. More details in the Jira ticket.

## Verification Instructions
0. Go through your regular workflow, but pay attention when you get to the synchronizations step (spinning icons when loading from pinwheel). Copy the URL from this step for later.
1. Verify the workflow succeeds. There should be no regression in behavior.
2. Paste the URL you copied but change the ID in the URL to something slightly different (and likely to not be real). When you navigate to this page, you should see our nice "the sync failed" screen. Previously, you would have seen an ugly Rails application error.

## Acceptance testing


app/app/controllers/cbv/synchronizations_controller.rb
```diff
@@ -11,12 +11,10 @@ def update
...
-    if @pinwheel_account && @pinwheel_account.has_fully_synced?
-      if @pinwheel_account.has_required_data?
-        redirect_to cbv_flow_payment_details_path(user: { account_id: @pinwheel_account.pinwheel_account_id })
-      else
-        redirect_to cbv_flow_synchronization_failures_path
-      end
+    if @pinwheel_account && @pinwheel_account.has_fully_synced? && @pinwheel_account.has_required_data?
+      redirect_to cbv_flow_payment_details_path(user: { account_id: @pinwheel_account.pinwheel_account_id })
+    else
+      redirect_to cbv_flow_synchronization_failures_path
```
tdooner, line 17: As discussed, I don't think this will work because `@pinwheel_account.has_fully_synced?` will be false to begin with and we don't want to immediately redirect the user to the failure path.

app/Gemfile.lock
GeorgeCodes19, line 1: are these changes ancillary to the actual issue or are they addressing another issue? One that's perhaps related to outdated dependencies. If so they're related to dependencies then maybe we should separate this into a separate PR.

app/app/controllers/cbv/synchronizations_controller.rb
```diff
@@ -11,7 +11,11 @@ def update
   private
 
   def redirect_if_sync_finished
-    if @pinwheel_account && @pinwheel_account.has_fully_synced?
+    if @pinwheel_account.nil?
+      redirect_to cbv_flow_synchronization_failures_path and return
```
tdooner, line 15: `and return` is a terrible rubyism that I'd like us to avoid.

Reason being:
1. I would like it to be very obvious when methods are returning
2. This is hard to see because it's at the end of a line
3. It also relies on the truthiness of `redirect_to` in order to return. That's weird!
4. It uses the `and` keyword instead of the `&&` operator. Although this is [the most controversial ruby style preference of all time](https://github.com/rubocop/ruby-style-guide#and-or-flow), it is undeniable that there can be weird bugs related to the low precedence of `and` and `or`. Best to just avoid them.

tl;dr: no reason not to just:
```suggestion
      return redirect_to cbv_flow_synchronization_failures_path
```

app/Gemfile.lock
millerti: This is directly related to the crash that we observed in prod. Do you still think we should separate it? 

Other issue: idk why activejob got deleted. I didn't do that on purpose. I think it's a residual change from merging main a few times but idk why that would be. I will re-do my turbo-rails change and commit before merge regardless, depending on your answer to the above question.

GeorgeCodes19: > This is directly related to the crash that we observed in prod. Do you still think we should separate it?
> 
> Other issue: idk why activejob got deleted. I didn't do that on purpose. I think it's a residual change from merging main a few times but idk why that would be. I will re-do my turbo-rails change and commit before merge regardless, depending on your answer to the above question.

Ahh okay, I think it makes sense to include in this PR then.

PR #399, opened 2024-12-17
FFS-2162: Set up mixpanel and use it to track page views

tdooner:
## Ticket

Resolves [FFS-2162](https://jiraent.cms.gov/browse/FFS-2162).

## Changes

* **Add MixpanelEventTracker proof of concept**

## Context for reviewers

This is just an initial implementation. Plans for this PR:
- [x] Add environment variable to terraform

Plans for a subsequent PR that finishes the NewRelic -> Mixpanel transition:
- [ ] Call `mixpanel.identify` at some point to associate the distinct_id with the user's device id
- [x] Use `maybe_later` gem to move event tracking to happen after the body is sent to the client (for performance)
- [x] Audit event names to try and ensure consistency of "[subject][action][object]", e.g. "CaseworkerLoggedIn" or "ApplicantDownloadedIncomePDF" where possible.
- [ ] Try and simplify the metric tracking somehow, possibly by moving the tracking methods onto a separate module

Setup instructions:
1. Engineers will need to get the MIXPANEL_TOKEN value from the shared 1password .env.development.local file.
2. Engineers will need to create Mixpanel accounts.

## Acceptance testing

### DIFF
--- a/app/lib/mixpanel_event_tracker.rb
+++ b/app/lib/mixpanel_event_tracker.rb
@@ -1,0 +1,39 @@
+class MixpanelEventTracker
+  def self.for_request(request)
+    url_params = request.params.slice("site_id", "locale")
+
+    # Set these attributes as default (global) across all the tracked events.
+    new(
+      "$device_id" => request.session.id,
+      "ip" => request.ip,
+      "cbv_flow_id" => request.session[:cbv_flow_id],
+      "site_id" => url_params["site_id"],
+      "locale" => url_params["locale"],
+      "user_agent" => request.headers["User-Agent"],
+    )
+  end
+
+  def initialize(default_attributes)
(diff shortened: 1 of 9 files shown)


app/lib/mixpanel_event_tracker.rb
```diff
@@ -0,0 +1,39 @@
+class MixpanelEventTracker
+  def self.for_request(request)
+    url_params = request.params.slice("site_id", "locale")
+
+    # Set these attributes as default (global) across all the tracked events.
+    new(
```
GeorgeCodes19, line 6: Nice - I'm not sure why we didn't implement something like this for the NR event tracker.

GeorgeCodes19 (approved):

PR #400, opened 2024-12-18
FFS-1809: Disable multiple clicking on send invitation button

YvetteWhiteUSDS:
## Ticket
Resolves [FFS-1809](https://jiraent.cms.gov/browse/FFS-1809).

### Overview of the bug
When a user is presented with the invitation page, if the user clicks on the send button multiple times, then multiple duplicate invitations will be generated and sent to the invitee. The purpose of this fix is to make sure that only one invite is generated.

### Summary of changes
Note: This work went through several iterations of approaches. This final approach is deemed most ideal as it maintains the use of the Turbo framework and allows for component reuse. 

1. app/app/views/caseworker/cbv_flow_invitations/new.html.erb - Wrap the invitation form in a turbo form so that form field errors will be rendered. 
2. app/app/helpers/application_helper.rb - Create a custom "form_with" that will assist with the proper rendering of page errors in the header area and default the builder to UswdsFormBuilder.
3. app/app/controllers/caseworker/cbv_flow_invitations_controller.rb - Fixed a couple bugs that appeared when page errors (i.e., validation errors) were present. First, used `flash.now` instead of `flash` when setting the page's errors. The former will make sure the messages are shown on the page when no redirect occurs. Second, the page returns a 4XX status as opposed to a 2XX status. This fixes the error rendering at the top of the page.

## Acceptance testing

#### Validation errors are rendering properly
Screenshot 2024-12-23 at 7 08 13 PM

#### Screenshot of disabled button
Screenshot 2024-12-18 at 11 12 58 AM

#### Query of before and after an invitation was generated after the fix
Created an invitation for "Jimmy J". The first query is prior to invitation creation and the second is after.

Screenshot 2024-12-18 at 10 25 22 AM

#### For context, prior to changes, I was able to generate multiple invitations by continuously clicking the send button
Screenshot 2024-12-18 at 11 15 04 AM

### DIFF
--- a/app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
+++ b/app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
@@ -33,10 +33,10 @@ class Caseworker::CbvFlowInvitationsController < Caseworker::BaseController
       error_messages = @cbv_flow_invitation.errors.messages.values.flatten.map { |msg| "<li>#{msg}</li>" }.join
       error_messages = "<ul>#{error_messages}</ul>"
 
-      flash[:alert_heading] = error_header
-      flash[:alert] = error_messages.html_safe
+      flash.now[:alert_heading] = error_header
+      flash.now[:alert] = error_messages.html_safe
 
-      return render :new
+      return render :new, status: :unprocessable_entity
     end
 
     # hydrate the cbv_client with the invitation if there are no cbv_flow_invitation errors
--- a/app/app/helpers/application_helper.rb
+++ b/app/app/helpers/application_helper.rb
@@ -73,4 +73,14 @@ module ApplicationHelper
   def coalesce_to_completed(status)
     [ :unsupported, :failed ].include?(status) ? :completed : status
   end
+
+  def uswds_form_with(model: false, scope: nil, url: nil, format: nil, **options, &block)
+    options[:builder] = UswdsFormBuilder
+    options[:data] ||= {}
+    options[:data][:turbo_frame] = "_top"
+
+    turbo_frame_tag(model) do
+      form_with(model: model, scope: scope, url: url, format: format, **options, &block)
+    end
+  end
 end
--- a/app/app/views/caseworker/cbv_flow_invitations/new.html.erb
+++ b/app/app/views/caseworker/cbv_flow_invitations/new.html.erb
@@ -4,7 +4,7 @@
 <div class="usa-prose">
   <%= site_translation(".description_html", pay_income_days: current_site.pay_income_days) %>
 </div>
-<%= form_with(model: @cbv_flow_invitation, data: { turbo: false }, url: invitations_path, builder: UswdsFormBuilder) do |f| %>
+<%= uswds_form_with(model: @cbv_flow_invitation, url: invitations_path) do |f| %>
   <%= render partial: "caseworker/cbv_flow_invitations/#{@cbv_flow_invitation.site_id}", locals: { f: f } %>
   <%= f.submit t(".form.submit") %>
 <% end %>
--- a/app/docker-compose.yml
+++ b/app/docker-compose.yml
@@ -27,7 +27,7 @@ services:
         condition: service_healthy
       redis:
         condition: service_healthy
-    env_file: .env.local.bak
+    env_file: .env.development.local
     environment:
       - DOCKERIZED=true
       - DB_HOST=postgres
--- a/app/Gemfile
+++ b/app/Gemfile
@@ -71,7 +71,6 @@ gem "stackprof"
 gem "rexml", "~> 3.3.9"
 gem "gpgme", "~> 2.0", ">= 2.0.12"
 gem "pdf-reader", "~> 2.12.0"
-gem "skylight"
 
 group :development, :test do
   gem "brakeman", "~> 5.2"
--- a/app/Gemfile.lock
+++ b/app/Gemfile.lock
@@ -449,8 +449,6 @@ GEM
       connection_pool (>= 2.2.5, < 3)
       rack (~> 2.0)
       redis (>= 4.5.0, < 5)
-    skylight (6.0.4)
-      activesupport (>= 5.2.0)
     smart_properties (1.17.0)
     snaky_hash (2.0.1)
       hashie
@@ -572,7 +570,6 @@ DEPENDENCIES
   rubocop-rspec
   selenium-webdriver
   sidekiq (~> 6.4)
-  skylight
   sprockets-rails
   stackprof
   standard (~> 1.7)
--- a/app/app/controllers/health_check_controller.rb
+++ b/app/app/controllers/health_check_controller.rb
@@ -1,5 +1,7 @@
 class HealthCheckController < ActionController::Base
   def ok
+    # keep the database connection alive
+    ActiveRecord::Base.connection.execute("SELECT 1")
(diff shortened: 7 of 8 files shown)


app/app/javascript/controllers/caseworker/cbv_flow_invitations/invitation.js
```diff
@@ -0,0 +1,8 @@
+import { Controller } from "@hotwired/stimulus"
+export default class extends Controller {
+    static targets = ["submitButton"]
+
+    disableSubmit() {
+        this.submitButtonTarget.disabled = true;
+    }
+}
```
YvetteWhiteUSDS, line 8: A few thoughts here:
1. I didn't generate (using `bin/rails generate controller`)this controller; instead, I just created a file. By doing this, did I miss connections that get generated behind the scenes?
2. What do you think about the function name `disableSubmit`? Any standards, guidelines or preferences?
3. Is the creation of this controller overkill for a button event? 
4. I called this controller `invitation.js` and placed it into `controllers/caseworker/cbv_flow_invitations`.
    * It could be used for other form submit buttons if we wanted it to and, if so, then it shouldn't reside here. (And the name should change.)
    * Is `invitation.js` clear or is there a different naming pattern you use? What's the convention?

app/app/views/caseworker/cbv_flow_invitations/new.html.erb
```diff
@@ -4,7 +4,7 @@
 <div class="usa-prose">
   <%= site_translation(".description_html", pay_income_days: current_site.pay_income_days) %>
 </div>
-<%= form_with(model: @cbv_flow_invitation, data: { turbo: false }, url: invitations_path, builder: UswdsFormBuilder) do |f| %>
+<%= form_with(model: @cbv_flow_invitation, data: { controller: "invitation" }, url: invitations_path, builder: UswdsFormBuilder) do |f| %>
   <%= render partial: "caseworker/cbv_flow_invitations/#{@cbv_flow_invitation.site_id}", locals: { f: f } %>
-  <%= f.submit t(".form.submit") %>
+  <%= f.submit t(".form.submit"), data: { action: "invitation#disableSubmit" } %>
 <% end %>
```
YvetteWhiteUSDS, line 10: I removed the `turbo: false` and put in the controller I created. Thoughts on this approach?

app/docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
         condition: service_healthy
       redis:
         condition: service_healthy
-    env_file: .env.local.bak
+    env_file: .env.development.local
     environment:
       - DOCKERIZED=true
       - DB_HOST=postgres
```
YvetteWhiteUSDS, line 33: Random change. I kept it in here because it seems like most of the team doesn't use the compose. But I heard that `.env.development.local` is more standard so I changed it.

app/app/javascript/controllers/caseworker/cbv_flow_invitations/invitation.js
tdooner: 1. Nope, it's all good. The `bin/rails generate controller` is actually for backend controllers (those in app/controllers), not the Javascript controllers. (It's confusing!) The generator for backend controllers does scaffold a number of other files (mostly tests), but nothing that couldn't be done manually.
2. Seems like a fine method name to me.
3. Yeah, kind of. When you're back let's pair to discuss.
4. I see what you're going for, it could be a generic "ClickOnceButton" component or something like that. I'm a fan of not extracting something until you know it's a pattern (i.e. needed in two places).

app/app/views/caseworker/cbv_flow_invitations/new.html.erb
tdooner: It's complicated - let's chat.

tdooner: Since I'm not sure I'm going to connect with you today, here's what's complicated (we can still catch up if you want).

By removing `turbo: false`, you converted the form into a "Turbo form" which is sent via JS `fetch` through Rails's [Turbo JS framework](https://turbo.hotwired.dev/handbook/introduction#turbo-drive%3A-navigate-within-a-persistent-process). This appeared to work locally because Turbo actually already [has support for disabling the submit button](https://turbo.hotwired.dev/reference/attributes#automatically-added-attributes) when it is enabled for its forms. This is what I was suggesting in Slack.

However, it is not so simple as enabling Turbo on this because of the reason we initially added the `turbo: false`: validation errors. Try this with on your branch, and you'll see the problem: the validation errors don't render. (There should be validation errors on each field as well as an alert at the top of the page summarizing all the errors.)

To get the validation errors working again, we need to take advantage of Turbo frames by surrounding the entire page in a `<turbo-frame>` so it gets replaced. The syntax that worked for me for this was:

```ruby
  <%= turbo_frame_tag @cbv_flow_invitation, autoscroll: true do %>
```

around the contents of the `app/views/caseworker/cbv_flow_invitations/new.html.erb` view.

So, in total, what we should do to get this PR merged:
1. Remove your `invitation` controller, its instantiation, and the action on the submit button
2. Surround the `new` view in the turbo_frame_tag as described
3. We will need to figure out what to do about the top-of-page alert that contains a summary of all the validation errors, since that won't work properly when the turbo frame for the form reloads. I think perhaps we should move the alert rendering to be with the form. Probably best to pair on this.

Sorry this became complicated. I thought it was going to be easy but this is one of those things where this is a good thing to do nonetheless because it will help us get towards standard form styling and functionality in a simpler way.

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
```diff
@@ -33,10 +33,10 @@ def create
       error_messages = @cbv_flow_invitation.errors.messages.values.flatten.map { |msg| "<li>#{msg}</li>" }.join
       error_messages = "<ul>#{error_messages}</ul>"
 
-      flash[:alert_heading] = error_header
-      flash[:alert] = error_messages.html_safe
+      flash.now[:alert_heading] = error_header
+      flash.now[:alert] = error_messages.html_safe
```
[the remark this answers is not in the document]
tdooner: > Was the flash not appearing?

The opposite, actually, the flash was appearing on this page *and* whatever next page the user went to.

The form has weird behavior currently: it POSTs to `/sandbox/invitations`, which adds the validation errors via `flash`, but then doesn't take the user to a subsequent page (rather, it renders `:new` view). So it leaves the user on a page they can't refresh (`/sandbox/invitations`) with the `flash` in the cookie for the next request, e.g. if they hit the back button/refresh.

So by using `flash.now` it makes the flash error available to the current view but not afterwards.

app/app/helpers/application_helper.rb
```diff
@@ -73,4 +73,14 @@ def feedback_form_url
...
   end
+
+  def uswds_form_with(model: false, scope: nil, url: nil, format: nil, **options, &block)
+    options[:builder] = UswdsFormBuilder
+    options[:data] ||= {}
+    options[:data][:turbo_frame] = "_top"
+
+    turbo_frame_tag(model) do
+      form_with(model: model, scope: scope, url: url, format: format, **options, &block)
+    end
```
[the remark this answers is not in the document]
tdooner: Yeah, Turbo with `target="_top"` seems to work pretty well in general, so we could probably use it on all the other pages.

PR #401, opened 2024-12-18
FFS-2175: Adding DB query to /health endpoint

GeorgeCodes19:
## Ticket

Relates to [FFS-2175](https://jiraent.cms.gov/browse/FFS-2175).

## Acceptance testing


tdooner (approved):
Yep, that'll do it!

(Just for future record: it's generally not a good idea to test other systems in a healthcheck because if there's a database outage, then your web servers would all go down simultaneously. So, we can remove this in the future when we have enough traffic that we don't need to keep the database connection alive.)

PR #412, opened 2025-01-08
FFS-2291 - Implement Mixpanel events

millerti:
## Changes
1. Events now file to Mixpanel
2. Events still file to NewRelic
3. Some events have new names in Mixpanel per [this map](https://confluenceent.cms.gov/pages/viewpage.action?spaceKey=SFIV&title=CBV+New+Relic+Events)

## Testing
0. Add the MIXPANEL_TOKEN from 1Password into your env.development.local file
0. Satisfy yourself that going through the workflow logs events to both Mixpanel and NewRelic
0. Convince yourself that only the new Mixpanel event names are being logged to Mixpanel
0. Convince yourself that only the old NewRelic event names are being logged to NewRelic
0. Verify that the NewRelic dashboards still appear to work
0. Convince yourself that, when a new invitation_id is present, Mixpanel identifies events for that as being done by a new user

## Acceptance testing


app/lib/mixpanel_event_tracker.rb
```diff
@@ -0,0 +1,42 @@
...
+  def self.for_request(request)
+    new
+  end
+
+  def initialize
+    @tracker = Mixpanel::Tracker.new(ENV["MIXPANEL_TOKEN"], MixpanelErrorHandler.new)
+  end
+
+  def track(event_type, request, attributes = {})
+    start_time = Time.now
```
tdooner, line 11: Let's move `start_time` to within the MaybeLater block so it actually references the time it took to do the mixpanel bits.

app/spec/controllers/cbv/summaries_controller_spec.rb
```diff
@@ -229,20 +228,8 @@
...
-            invitation_id: cbv_flow_invitation.id,
-            account_count: 1,
-            paystub_count: 1,
-            account_count_with_additional_information: 0,
-            flow_started_seconds_ago: flow_started_seconds_ago,
-            language: "en".to_sym
-          })
-        end
+        # Note that we are not testing events here because doing so requires use of expect_any_instance_of,
+        # which does not play nice since there are multiple instances of the event logger.
```
tdooner, line 232: Could you make a follow-up story (it can go deep in the backlog) to see if we can figure out how to get this test back? It would be reassuring to have a test of this particular one, because it's kind of the most important event we have (the ultimate "conversion" type of thing).

app/spec/lib/new_relic_event_tracker_spec.rb
```diff
@@ -4,15 +4,23 @@
...
+    let(:request) { nil }
     let(:attributes) { { key: 'value' } }
+    let(:tracker) { described_class.for_request(nil) }
 
     before do
-      allow(NewRelicEventTracker).to receive(:track).and_call_original
+      # Make MaybeLater execute immediately instead of after the request, which doesn't exist
+      allow(MaybeLater).to receive(:run) do |&block|
+        block.call
+      end
```
tdooner, line 15: We probably should do this in the `spec_helper.rb` so it occurs everywhere rather than just the tests in this file.

tdooner (approved):
No blockers but some suggestions for future work that maybe we could add as tech debt tickets later.

app/app/controllers/api/pinwheel_controller.rb
```diff
@@ -13,6 +13,20 @@ class Api::PinwheelController < ApplicationController
...
+    "PinwheelShowProviderConfirmationPage" => "ApplicantViewedPinwheelProviderConfirmation",
+    "PinwheelShowLoginPage" => "ApplicantViewedPinwheelLoginPage",
+    "PinwheelAttemptLogin" => "ApplicantAttemptedPinwheelLogin",
+    "PinwheelSuccess" => "ApplicantSucceededWithPinwheelLogin",
+    "PinwheelError" => "ApplicantEncounteredPinwheelError",
+    "PinwheelShowDefaultProviderSearch" => "ApplicantViewedPinwheelDefaultProviderSearch",
+    "PinwheelAttemptClose" => "ApplicantAttemptedClosingPinwheelModal",
+    "PinwheelCloseModal" => "ApplicantClosedPinwheelModal",
+    "PinwheelAccountSyncFinished" => "ApplicantFinishedPinwheelSync"
+  }
```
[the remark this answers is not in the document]
millerti: Yeah I think I'd like to punt on this for now, esp because I don't fully understand the suggestion.

app/lib/generic_event_tracker.rb
```diff
@@ -0,0 +1,52 @@
...
+      defaults
+    )
+  end
+
+  def initialize(trackers, default_attributes)
+    @trackers = trackers
+    @default_attributes = default_attributes
+  end
+
+  def track(event_type, request, attributes = {})
```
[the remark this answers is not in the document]
millerti: Yeah, this is a good suggestion, but I definitely want to punt on this so I have a hope of getting this out the door before my extended outage.

app/spec/lib/new_relic_event_tracker_spec.rb
millerti: I don't think so, because in all the other tests, we completely stub out the trackers (there's no way for the MaybeLater stuff to get called). It's only in this test that that's not true.

app/spec/controllers/cbv/summaries_controller_spec.rb
millerti: Created https://jiraent.cms.gov/browse/FFS-2389

app/app/controllers/cbv/base_controller.rb
```diff
@@ -86,31 +86,29 @@ def prevent_back_after_complete
...
-      path: request.path
-    })
-  rescue => ex
-    Rails.logger.error "Unable to track NewRelic event (CbvPageView): #{ex}"
+    begin
+      event_logger.track("CbvPageView", request, {
+        cbv_flow_id: @cbv_flow.id,
+        invitation_id: @cbv_flow.cbv_flow_invitation_id,
+        site_id: @cbv_flow.site_id,
+        path: request.path
```
allthesignals, line 94: Note, I believe this is breaking some views, specifically expired links.

The error message is being swallowed and the user is redirected to a flash error message, which is why we didn't catch it.

PR #417, opened 2025-01-22
FFS-2295: Help modal

GeorgeCodes19:
## Ticket

Resolves [FFS-2295](https://jiraent.cms.gov/browse/FFS-2295).

- [x] Update help topic(s) content
- [x] Add E2E tests
- [x] Add `/help` route to use in iframe
- [x] Add `Help` link in top nav that triggers the Help Modal
- [x] Ensuring that clicking the options link in the help banner triggers the Help Modal
- [x] Ensuring that the "Go Back" button works within the Help Modal
- [x] Ensuring that closing the Pinwheel modal triggers the Help Banner to appear

## Acceptance testing


app/spec/rails_helper.rb
```diff
@@ -10,7 +10,13 @@
 require "view_component/system_test_helpers"
 
 require "capybara/rspec"
+require "rack/handler/puma"
 Capybara.default_driver = Capybara.javascript_driver
+Capybara.server = :puma
+Capybara.register_server :puma do |app, port, host|
```
GeorgeCodes19, line 16: I couldn't get Capybara to automatically register puma as a server so I manually did it.

app/app/components/help_link_component.html.erb
GeorgeCodes19, line 1: I opted to use a component here since there's now a `Help` nav item in the main nav as well as a link in the Help banner.

app/app/components/help_modal_component.html.erb
```diff
@@ -0,0 +1,29 @@
+<div
+  class="usa-modal maxw-none-important"
+  id="help-modal"
+  aria-labelledby="help-modal-heading"
+  aria-describedby="help-modal-description"
+>
+  <div class="usa-modal__content">
+    <div class="usa-modal__main padding-0 margin-0 maxw-none-important">
+      <div class="usa-modal__body padding-0">
+        <iframe
```
GeorgeCodes19, line 10: Note that I've resorted to an iframe here rather coding a bunch of JS. I figured this would more closely resemble our current approach and that we can add custom Mixpanel/New Relic events using Rails helpers.

app/app/components/help_topic_content_component.html.erb
GeorgeCodes19, line 1: Rather than creating explicit views for each help topic I opted to rely on the `en.yml` entries and this component since the help topic content share a consistent layout.

app/app/components/help_modal_component.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize(open: false, help_path: nil)
+    @open = open
+    @help_path = help_path
+  end
+
+  private
+
+  def modal_id
+    "help-modal"
+  end
```
allthesignals, line 11: I don't think this gets used anywhere

app/app/components/help_modal_component.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def open?
+    @open
+  end
```
allthesignals, line 15: I don't think this gets used anywhere

app/app/components/help_modal_component.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def help_path
+    @help_path
+  end
```
allthesignals, line 19: I wouldn't wrap this in a method like this because it's somewhat superfluous.

app/app/components/help_link_component.rb
```diff
@@ -0,0 +1,16 @@
...
+    @text = text
+  end
+
+  def before_render
+    @text ||= t("help.alert.help_options")
+  end
+
+  private
+
+  attr_reader :class_name, :text
```
GeorgeCodes19, line 15: Move to top of class?

app/app/components/help_link_component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class HelpLinkComponent < ViewComponent::Base
+  def initialize(class_name: nil, text: nil)
+    @class_name = class_name
+    @text = text
+  end
+
+  def before_render
+    @text ||= t("help.alert.help_options")
```
GeorgeCodes19, line 10: Pass this in, rather than setting the default in the component

app/app/components/help_link_component.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
```
allthesignals, line 11: Im' not sure what this is for yet maybe vestigial? I think it's confusing way to set a default

app/app/components/help_link_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= link_to text, "#help-modal", 
+  class: class_name,
+  aria: { controls: "help-modal" }, 
+  data: { open_modal: true } %>
```
allthesignals, line 4: This could be a partial probably

app/app/components/help_alert_component.html.erb
GeorgeCodes19, line 1: This can be a partial since it's only used once.

app/app/components/help_alert_component.rb
allthesignals, line 1: I think this can be a partial too

app/app/components/help_modal_component.rb
allthesignals, line 1: I would see if you can get away with a partial for this too :)

app/app/components/help_alert_component.html.erb
allthesignals, line 1: I woudl investigate whether you could do this within the existing flash error/alert system :)

app/app/views/layouts/application.html.erb
```diff
@@ -18,6 +18,8 @@
       <%= render "application/header" %>
       <main id="main-content">
         <section class="grid-container usa-section">
+          <%= render HelpAlertComponent.new(visible: params[:help] == "true", help_path: cbv_flow_help_index_path(locale: I18n.locale)) %>
```
GeorgeCodes19, line 21: Rather than using HelpAlertComponent we can read the help params within the help controller and use `flash` instead. This would mean that we don't introduce new patterns.

app/app/views/layouts/application.html.erb
```diff
@@ -54,5 +56,7 @@
       </main>
       <%= render "application/footer" %>
     </div>
+
+    <%= render HelpModalComponent.new(open: params[:help_modal] == "true", help_path: cbv_flow_help_index_path(locale: I18n.locale)) %>
```
allthesignals, line 60: This is here because this modal can always be accesse and the design calls for accessing the help modal anywhere in the app

app/app/components/help_topic_content_component.html.erb
```diff
@@ -0,0 +1,19 @@
+<p><%= t("help.show.#{topic}.intro") if t("help.show.#{topic}.intro", default: '').present? %></p>
+
+<% if t("help.show.#{topic}.security_message", default: '').present? %>
+  <p><%= t("help.show.#{topic}.security_message") %></p>
+<% end %>
+
+<ol>
+  <% steps = t("help.show.#{topic}", default: {}).select { |k, _| k.to_s.start_with?('step') } %>
+  <% steps.each do |step, _| %>
```
allthesignals, line 9: I would not use this `_` thing :)

app/app/assets/stylesheets/cbv.scss
```diff
@@ -132,3 +132,37 @@ html {
...
+/**
+ * Help Modal
+ */
+#help-iframe {
+  height: 68vh;
+  
+  /* Target mobile devices in portrait mode */
+  @include at-media-max('mobile-lg') {
+    @media (orientation: portrait) {
+      height: 73vh;
```
GeorgeCodes19, line 145: Consult Sam about the iframe height. Currently using `vh`

app/app/components/help_modal_component.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      type="button"
+      class="usa-button usa-modal__close"
+      aria-label="Close this window"
+      data-close-modal>
+      <svg class="usa-icon" aria-hidden="true" focusable="false" role="img">
+        <use xlink:href="<%= asset_path("@uswds/uswds/dist/img/sprite.svg#close") %>"></use>
+      </svg>
+    </button>
+  </div>
+</div>
```
tdooner, line 29: Hmm, would have been interesting to try and make a `UswdsModal` component that took arbitrary contents (i.e. the help `<iframe>`). Then it would be reusable if we needed another modal elsewhere. Don't worry about it for now, though, just a thought. We could generalize it later if we need another modal.

app/config/initializers/content_security_policy.rb
```diff
@@ -7,25 +7,24 @@
...
     policy.object_src :none
     policy.script_src :self, "https://js-agent.newrelic.com", "https://*.nr-data.net", "https://cdn.getpinwheel.com"
     policy.connect_src :self, "https://*.nr-data.net"
     policy.worker_src :self, "blob:"
-    policy.frame_src :self, "https://cdn.getpinwheel.com"
+    policy.style_src :self
     policy.style_src_elem :self
-    # Allow inline styles as we have some on the payment_details page:
-    policy.style_src_attr "'unsafe-inline'"
+    policy.style_src_attr :self
```
tdooner, line 21: This will break the payment_details page rendering, since there are some inline styles on that page (see the comment on line 20):

image

(The section headers should have blue backgrounds.)

I'm going to make a ticket to remove the inline styles since they shouldn't be necessary. Until then, we'll have to keep this as `"'unsafe-inline'"`.

app/config/initializers/content_security_policy.rb
```diff
@@ -7,25 +7,24 @@
 Rails.application.configure do
   config.content_security_policy do |policy|
     policy.default_src :self
-    policy.font_src :self
+    policy.font_src :self, :data
```
tdooner, line 10: Not sure I understand why this change is necessary - did you change anything having to do with fonts?

app/app/views/layouts/application.html.erb
tdooner: To match the other param name:

```suggestion
    <%= render HelpModalComponent.new(open: params[:help] == "true", help_path: cbv_flow_help_index_path(locale: I18n.locale)) %>
```

tdooner (commented):
Looking good. Only blocker I see is the CSP policy and the visual regression on the payment_details page.

tdooner:
@GeorgeCodes19 Also, I just discovered weird behavior if the session expires:

image

app/app/javascript/utilities/help.js
allthesignals, line 1: Hmm, I don't understand why we need all this API stuff. If it's for sending events, doesn't MixPanel have an [SDK for sending events](https://docs.mixpanel.com/docs/tracking-methods/sdks/javascript)?

I can definitely see something like this API warranted in the future, though, but it's overkill when it seems MixPanel has a similar feature to what NR offered...

app/app/views/help/_help_link.html.erb
allthesignals, line 1: Great!!

app/app/views/help/_help_link.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= link_to text, "#help-modal", {
+  aria: { controls: "help-modal" },
+  data: { 
+    open_modal: true, 
+    source: source 
+  }
+}.compact
```
allthesignals, line 7: I guess it's nice to compact this in case `source` isn't provided. Seems we always pass the param though.

app/app/javascript/utilities/help.js
allthesignals: Okay I think I get it. But it seems like we have _two_ of these `trackUserAction` things. Is there some reason we can't re-use the one already defined in Pinwheel.js? Can we pull that out and make it more generic? 

I think my broader point on leveraging the MixPanel SDK for this kind of thing still stands but I see that it's already been implemented.

config/routes.rb
allthesignals, line 1: Wait was this accidentally created?

GeorgeCodes19: yup! good catch

tdooner (approved):
Looks good to me, pending the one last issue I noticed (the feedback form URL)

app/app/assets/stylesheets/cbv.scss
```diff
@@ -132,3 +132,50 @@ html {
...
+  }
+
+  // Larger screens such as desktop
+  @include at-media(desktop) {
+    height: 660px;
+  }
+
+  html {
+    overflow: hidden;
+  }
```
allthesignals, line 160: Does this do anything? I don't think iframe contents can inherit like this.

app/config/locales/en.yml
```diff
@@ -384,6 +384,99 @@ en:
...
+      select_prompt: 'Select an option below:'
+      title: Need help connecting to your payroll provider?
+      username: I don't know my username
+    show:
+      company_id:
+        intro: 'Try the following steps:'
+        step1:
+          search_keywords: Search your email for keywords like "payroll," "[Payroll Provider Name]," or "company ID."
+          title: Check your welcome materials
+          welcome_materials: Your company ID is often in your onboarding documents, welcome email, or payroll setup info.
```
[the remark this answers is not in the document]
allthesignals: I'm sad this feedback didn't make it in!

app/app/controllers/api/help_controller.rb
allthesignals, line 1: Eager for this redundant controller to be refactored away in follow up work

allthesignals (approved):

app/app/views/help/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+
+    <%= render "help_topic_link", topic: "username", text: t("help.index.username") %>
+    <%= render "help_topic_link", topic: "password", text: t("help.index.password") %>
+    <%= render "help_topic_link", topic: "company-id", text: t("help.index.company_id") %>
+    <%= render "help_topic_link", topic: "employer", text: t("help.index.employer") %>
+    <%= render "help_topic_link", topic: "provider", text: t("help.index.provider") %>
+    <%= render "help_topic_link", topic: "credentials", text: t("help.index.credentials") %>
+
+    <% if current_site && current_site.caseworker_feedback_form.present? %>
+      <%= link_to current_site.caseworker_feedback_form, class: "usa-button margin-bottom-1 height-5 text-left display-block", target: "_blank" do %>
```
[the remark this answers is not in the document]
GeorgeCodes19: Thanks for catching that! Fixed.

app/app/controllers/api/help_controller.rb
GeorgeCodes19: Yup, it's actually going to be addressed in the follow up ticket that I'll start on today.

app/config/locales/en.yml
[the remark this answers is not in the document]
GeorgeCodes19: I've updated the PR to include this fix!

app/app/assets/stylesheets/cbv.scss
GeorgeCodes19: It does NOTHING. I was experimenting with some things and accidentally left it in there. The solution here was to set the `scrolling` property on the iframe itself to `false`. This approached worked, but it's actually deprecated.

PR #424, opened 2025-01-29
FFS-2401: Add bare bones invitations controller

allthesignals:
## Ticket

Resolves [FFS-2401](https://jiraent.cms.gov/browse/FFS-2401).

## Changes
Adds basic invitations controller, returns new invitation with working URL.

## Context for reviewers
Recycles the cbv_caseworker/invitations controller. Currently accepts a user_id but will later work by looking for an API_KEY in the headers, inferring a service account user, and using that record internally instead.

## Acceptance testing


app/app/controllers/api/invitations_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      :email_address,
+      :snap_application_date,
+      :agency_id_number,
+      :beacon_id,
+    )
+  end
+
+  def site_id
+    params[:site_id]
+  end
```
allthesignals, line 65: Is there a more future-proof way of managing these different params? could they be inferred from the CbvApplicant model?

app/app/controllers/api/invitations_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    # this is an old refactor
+    @cbv_client = CbvClient.create_from_invitation(@cbv_flow_invitation)
+
+    render json: { **@cbv_flow_invitation.as_json, url: @cbv_flow_invitation.to_url }, status: :created
+  end
+
+  # todo: remove this shim, replace with real user
+  def _service_account_user_shim
+    User.new(email: "service@account.com", site_id: site_id)
+  end
```
allthesignals, line 22: Shimmed the user for now but will figure out real users in a bit

app/spec/controllers/api/invitations_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    it "creates an invitation" do
+      post :create, params: {
+        site_id: :sandbox,
+        cbv_flow_invitation: valid_params
+      }
+
+      expect(response).to have_http_status(:created)
+      expect(JSON.parse(response.body).keys).to include("url")
```
allthesignals, line 16: Is there a better "more real" way to test this? I have some patterns from past projects on testing API endpoints that i'll dig into:

app/app/controllers/api/invitations_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    # hydrate the cbv_client with the invitation if there are no cbv_flow_invitation errors
+    # this is an old refactor
+    @cbv_client = CbvClient.create_from_invitation(@cbv_flow_invitation)
+
+    render json: { **@cbv_flow_invitation.as_json, url: @cbv_flow_invitation.to_url }, status: :created
+  end
+
+  # todo: replace with inference via API_KEY
+  def service_account_user
+    User.find_by(id: cbv_flow_invitation_params[:user_id])
```
allthesignals, line 24: note that find_by will return nil if it can't find the record, which is handled in the create method. it's too bad this can't be handled by rails model validation features, but if we don't validate this before creating an invitation, it could throw a lower-level parsing error from active record that i believe would surface as a 500 rather than a more helpful validation error message...

app/app/controllers/api/invitations_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    @cbv_flow_invitation = CbvInvitationService.new(event_logger).invite(cbv_flow_invitation_params, service_account_user)
+
+    if @cbv_flow_invitation.errors.any?
+      return render json: @cbv_flow_invitation.errors, status: :unprocessable_entity
+    end
+
+    # hydrate the cbv_client with the invitation if there are no cbv_flow_invitation errors
+    # this is an old refactor
+    @cbv_client = CbvClient.create_from_invitation(@cbv_flow_invitation)
```
allthesignals, line 17: This should be addressed in that refactor follow up ticket: https://jiraent.cms.gov/browse/FFS-2110

allthesignals, line 17: This should be addressed in that refactor follow up ticket: https://jiraent.cms.gov/browse/FFS-2110

app/app/controllers/api/invitations_controller.rb
```diff
@@ -0,0 +1,48 @@
+class Api::InvitationsController < ApplicationController
+  skip_forgery_protection
+
+  def create
+    if service_account_user.nil?
+      return render json: { error: "User not found" }, status: :unprocessable_entity
+    end
+
+    @cbv_flow_invitation = CbvInvitationService.new(event_logger).invite(cbv_flow_invitation_params, service_account_user)
```
tdooner, line 9: See https://github.com/DSACMS/iv-cbv-payroll/pull/425, which you can merge/rebase on if you want, in order to add `delivery_method: nil` here.

app/app/controllers/api/invitations_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    if @cbv_flow_invitation.errors.any?
+      return render json: @cbv_flow_invitation.errors, status: :unprocessable_entity
+    end
+
+    # hydrate the cbv_client with the invitation if there are no cbv_flow_invitation errors
+    # this is an old refactor
+    @cbv_client = CbvClient.create_from_invitation(@cbv_flow_invitation)
+
+    render json: { **@cbv_flow_invitation.as_json, url: @cbv_flow_invitation.to_url }, status: :created
```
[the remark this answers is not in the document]
tdooner: Oh man, I forgot about that aspect of API management. I've used serializer libraries in the past ([this one](https://github.com/rails-api/active_model_serializers) back before it became unmaintained), and they are very helpful, but I wonder if we can get away without them for now due to the simplicity of this API? Maybe once we have multiple versions or something, then we bring in a library?

app/app/controllers/api/invitations_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      :client_id_number,
+      :case_number,
+      :email_address,
+      :snap_application_date,
+      :agency_id_number,
+      :beacon_id,
+      :user_id,
+      :site_id
+    )
+  end
```
tdooner, line 43: Would be nice if our API library could keep track of which fields were valid (synced with the documentation), and then ensure that only those fields are added to the `params` object.

tdooner (approved):
seems legit

app/app/controllers/api/invitations_controller.rb
allthesignals: @tdooner yeah, I am not happy with param management here. I'd like to think the future CbvApplicant model would be able to provide information about client-specific required fields. Previously, we included that logic here in the controller.

I think this controller code here is way less sophisticated than what was built for the caseworker invitations view, but I wanted to keep it simpler and less restricted for now until the dust settled on CbvApplicant and client agency metadata.

allthesignals: Note for this ticket here: https://jiraent.cms.gov/browse/FFS-2404

app/app/controllers/api/invitations_controller.rb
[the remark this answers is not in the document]
allthesignals: Ha, I've used that one in the past, too. There are so many of them, even [specific implementations of JSON standards](https://github.com/stas/jsonapi.rb). 

Definitely agree we can get away with avoiding it for now.

PR #425, opened 2025-01-29
FFS-2402: Add delivery_method parameter to invite method

tdooner:
## Ticket

Resolves [FFS-2402](https://jiraent.cms.gov/browse/FFS-2402).

## Changes

**FFS-2402: Add delivery_method parameter to invite method**

## Acceptance testing

### DIFF
--- a/app/spec/services/cbv_invitation_service_spec.rb
+++ b/app/spec/services/cbv_invitation_service_spec.rb
@@ -1,0 +1,105 @@
+require "rails_helper"
+
+RSpec.describe CbvInvitationService, type: :service do
+  let(:event_logger) { instance_double('GenericEventTracker') }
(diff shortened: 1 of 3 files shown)


app/spec/services/cbv_invitation_service_spec.rb
allthesignals, line 1: Awesome, thank you!

allthesignals (approved):
> In addition to this, I removed the exception catching from the service
> itself, since the controller that calls the service also catches
> exceptions.

Yeah, I find this to all be a little awkward. Validation errors, for example, are managed separately through checking whether the model reports any, but mailer errors are managed in this begin-rescue... not sure whether there's a better pattern or not. Would want to think about it more for the API controller at least.

PR #431, opened 2025-01-31
FFS-2403: First pass at token-based authentication

allthesignals:
## Ticket

Resolves [FFS-2403](https://jiraent.cms.gov/browse/FFS-2403).

## Changes
Adds API Access Tokens, associates with a User, provides some methods to create and find/validate a user by token. Protects the invitations api with token auth.

## Context for reviewers
WIP, needs more testing and some other, smaller requirements.

## Acceptance testing


app/app/models/api_access_token.rb
```diff
@@ -0,0 +1,9 @@
+class ApiAccessToken < ApplicationRecord
+  belongs_to :user
+
+  has_secure_password :access_token, validations: false
+
+  before_create do
+    self.access_token = SecureRandom.urlsafe_base64
+  end
```
allthesignals, line 8: @tdooner I'm not sure if this is right... but seems okay, I guess. `has_secure_password` makes available the `authenticate_***` method and that's what I used to validate.

app/app/models/user.rb
```diff
@@ -3,4 +3,14 @@ class User < ApplicationRecord
...
+
+  has_many :api_access_tokens, dependent: :destroy
+
+  def create_api_access_token
+    ApiAccessToken.create(user: self).access_token
+  end
+
+  def self.find_by_valid_access_token(token)
+    ApiAccessToken.all.find { |t| t.authenticate_access_token(token) }&.user
+  end
```
allthesignals, line 15: I'm not sure what others think but this method was a bit hard to name because it both finds a user by a given token through validating that token against _all tokens_. Which seems slow. I'm not sure how else token lookup is supposed to work.

allthesignals: Like maybe you could do something fancier in SQL, i don't know.

app/spec/controllers/api/invitations_controller_spec.rb
```diff
@@ -3,19 +3,23 @@
...
         beacon_id: "ABC123",
-        agency_id_number: "7890120",
-        user_id: service_account_user.id
+        agency_id_number: "7890120"
       )
     end
 
+    before do
+      request.headers["Authorization"] = "Bearer #{api_access_token.access_token}"
+    end
```
allthesignals, line 21: I wish there was a nicer helper for this kind of thing

app/spec/factories/api_access_token.rb
```diff
@@ -0,0 +1,4 @@
+FactoryBot.define do
+  factory :api_access_token do
+  end
+end
```
allthesignals, line 4: Feel like I could just put this into the User factor to generate tokens as part of a trait

app/.env
```diff
@@ -11,3 +11,7 @@ MA_PINWHEEL_ENVIRONMENT=sandbox
 SANDBOX_PINWHEEL_ENVIRONMENT=sandbox
 MAINTENANCE_MODE=false
 MA_WEEKLY_REPORT_RECIPIENTS=test@email.com
+
+ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY=primary
+ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY=deterministic
+ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT=derivationsalt
```
tdooner, line 17: It would be good to include the terraform to create these in AWS, otherwise I think the app will crash when it tries to start up lacking these values?

app/app/models/api_access_token.rb
```diff
@@ -0,0 +1,9 @@
+class ApiAccessToken < ApplicationRecord
+  belongs_to :user
+
+  encrypts :access_token, deterministic: true
+
+  before_create do
+    self.access_token = SecureRandom.urlsafe_base64
```
tdooner, line 7: nit: Can we use `SecureRandom.alphanumeric(32)` instead? Just a bit of an aesthetic preference because `base64` can include dashes/underscores, which feel a bit unusual to have in API keys. Also, the default of 16 bytes turns into a 22-character-long API key after it's encoded, which feels a bit short. See what I mean?

image

app/app/models/user.rb
```diff
@@ -3,4 +3,16 @@ class User < ApplicationRecord
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
   devise :rememberable, :trackable, :timeoutable, :omniauthable
+
+  has_many :api_access_tokens, dependent: :destroy
+
+  def create_api_access_token
+    ApiAccessToken.create(user: self).access_token
+  end
```
tdooner, line 11: Not sure why we have this method? We could instead do this anywhere we needed to create and return an access_token:

```
  user.api_access_tokens.create
```

Also this method is only used in test files, where we could just do `create(:user, :with_access_token)` as you defined here.

tdooner (approved):
Looks good but we will need the terraform before deploying or else I'd bet the app won't start up.

app/.env
allthesignals: Oh crap, you're right. Thank you!

app/app/models/api_access_token.rb
allthesignals: Yes! They look weird now, I'll change it.

allthesignals:
Ugh I misnamed something, fixing

PR #432, opened 2025-01-31
FFS-2326: Encapsulate Pinwheel responses into data objects

tdooner:
## Ticket
Ready for review 2/3.

Todo list:
- [x] Do the same conversion for all the other Pinwheel API endpoints
- [x] Test the flow in sandbox and development environments

Resolves [FFS-2326](https://jiraent.cms.gov/browse/FFS-2326).

## Changes

**Upgrade Rubocop, remove standard**
> Standard gem is not being actively used as a linter, and it's holding
> back Rubocop's version. Let's remove it.

**Use `activeresource` gem for employment Pinwheel response**
> This will allow us to access their properties using dots, rather than
> having objects whose keys are a mixture of symbols and strings.

> This first one for the /employments endpoint is a proof of concept and
> I'll be following up with the rest in subsequent commits.

## Acceptance testing


app/app/views/cbv/summaries/show.pdf.erb
```diff
@@ -92,11 +92,12 @@
       <%= table.with_data_point(:client_full_name, summary[:identity]["full_name"]) %>
     <% end %>
     <% if summary[:has_employment_data] %>
-      <%= table.with_data_point(:employer_phone, summary.dig(:employment, "employer_phone_number", "value")) %>
-      <%= table.with_data_point(:employer_address, summary.dig(:employment, "employer_address", "raw")) %>
-      <%= table.with_data_point(:employment_status, summary[:employment]["status"]) %>
-      <%= table.with_data_point(:employment_start_date, summary[:employment]["start_date"]) %>
-      <%= table.with_data_point(:employment_end_date, summary[:employment]["termination_date"]) %>
+      <% employment = summary[:employment] %>
+      <%= table.with_data_point(:employer_phone, employment.employer_phone_number.value) %>
```
allthesignals, line 96: Oh cool, so it handles nested stuff okay

app/app/services/pinwheel_service.rb
```diff
@@ -150,7 +165,9 @@ def fetch_paystubs(account_id:, **params)
   end
 
   def fetch_employment(account_id:)
-    @http.get(build_url("#{ACCOUNTS_ENDPOINT}/#{account_id}/employment")).body
+    json = @http.get(build_url("#{ACCOUNTS_ENDPOINT}/#{account_id}/employment")).body
+
+    Employment.new(json["data"], environment: @environment)
```
allthesignals, line 170: Okay, I see, we're not having ARes issue requests. Makes sense why we have to do that stuff at the top.

app/app/services/pinwheel_service.rb
```diff
@@ -3,6 +3,21 @@
...
+    def initialize(*params, environment:)
+      # ActiveResource requires us to set the `site` (the API base url) on the
+      # class. Since Pinwheel's API base URL's differ per-environment, let's
+      # set the value just in time for these records to be instantiated.
+      #
+      # This is thread-unsafe, but hopefully since this is just instantiating
+      # an object, the GVL will prevent any other thread from interrupting us.
+      self.class.site = environment[:base_url]
+      super(*params)
+    end
```
allthesignals, line 16: This particuilar [docs section](https://github.com/rails/activeresource?tab=readme-ov-file#configuration-and-usage) mentions a way to set this stuff dynamically:

```
ActiveResource::Base.site = api_site_for(request)
```

Is this different? No idea what `api_site_for` is, though... maybe something specific to the example.

app/app/services/pinwheel_service.rb
```diff
@@ -103,6 +103,53 @@ class PinwheelService
...
+
+      base_hours + overtime_hours
+    end
+
+    def hours_by_earning_category
+      earnings
+        .filter { |e| e.hours && e.hours > 0 }
+        .group_by { |e| e.category }
+        .transform_values { |earnings| earnings.sum { |e| e.hours } }
+    end
```
allthesignals, line 149: Yessssssss

app/app/views/cbv/summaries/show.pdf.erb
tdooner: Yep, it'll recursively make objects for any nested hash.

app/app/services/pinwheel_service.rb
```diff
@@ -103,6 +103,53 @@ class PinwheelService
...
+  # via dot-notation.
+  class ResponseObject < ActiveResource::Base
+    def initialize(*params, environment:)
+      # ActiveResource requires us to set the `site` (the API base url) on the
+      # class. Since Pinwheel's API base URL's differ per-environment, let's
+      # set the value dynamically as these records are instantiated.
+      self.class.site = environment[:base_url]
+      super(*params)
+    end
+  end
```
allthesignals, line 116: I still don't quite understand why we need to set like this. According to [this](https://github.com/rails/activeresource?tab=readme-ov-file#configuration-and-usage):

> Connection settings (site, headers, user, password, bearer_token, proxy) and the connections themselves are store in thread-local variables to make them thread-safe, so you can also set these dynamically, even in a multi-threaded environment, for instance:

```ruby
ActiveResource::Base.site = api_site_for(request)
```

app/app/services/pinwheel_service.rb
```diff
@@ -103,6 +103,53 @@ class PinwheelService
...
+      base_hours + overtime_hours
+    end
+
+    def hours_by_earning_category
+      earnings
+        .filter { |e| e.hours && e.hours > 0 }
+        .group_by { |e| e.category }
+        .transform_values { |earnings| earnings.sum { |e| e.hours } }
+    end
+  end
```
allthesignals, line 150: Would it make sense at some point to have these as separate model definition files? I guess the others don't need it because they're empty!

app/app/services/pinwheel_service.rb
```diff
@@ -146,19 +193,26 @@ def fetch_accounts(end_user_id:)
   end
 
   def fetch_paystubs(account_id:, **params)
-    @http.get(build_url("#{ACCOUNTS_ENDPOINT}/#{account_id}/paystubs"), params).body
+    json = @http.get(build_url("#{ACCOUNTS_ENDPOINT}/#{account_id}/paystubs"), params).body
+    json["data"].map { |paystub_json| Paystub.new(paystub_json, environment: @environment) }
```
allthesignals, line 197: I think it makes sense to leave the actual work of issuing HTTP requests to our existing Faraday implementation, but I am curious about what it'd look like to leverage ARes's APIs for this (would this look like `Account.find(account_id).paystubs`?)

app/spec/support/test_helpers.rb
```diff
@@ -6,26 +6,32 @@ def stub_environment_variable(variable, value, &block)
...
+        gross_pay_ytd: 1_000,
         deductions: [],
-        hours_by_earning_category: { salary: 80 }
+        earnings: []
       }
+
+      PinwheelService::Paystub.new(
+        json,
+        environment: PinwheelService::ENVIRONMENTS[:sandbox]
+      )
```
allthesignals, line 28: Wonderful. So now we're not doing this thing where we're mocking data in the post-processed form...

allthesignals (approved):
image

app/app/services/pinwheel_service.rb
tdooner: Yeah, I considered using activeresource for the requests as well, but I didn't want there to be two different ways that we were making these requests (and also I didn't want to refactor the entire pinwheel integration in one go).

app/app/services/pinwheel_service.rb
tdooner: Yeah, that's basically what this code does. It has to be dynamic because we support all pinwheel environments at the same time. I did see in ActiveResource documentation that they store each `site` value in a Thread-local variable, so that makes my thread safety concern go away.

app/app/services/pinwheel_service.rb
tdooner: Agreed, yeah, breaking these out into some kind of file structure is a natural move whenever we have to add more to these classes. I imagine we will once we need these data objects to provide the same interface between Pinwheel and another aggregator.

app/spec/support/test_helpers.rb
tdooner: Yeah. Maybe these test stubs can be simplified, actually, now that we're doing the same thing in multiple places.

app/app/services/pinwheel_service.rb
allthesignals: Just my preference that that does feel less weird than overriding init, but I'm not married to it

PR #436, opened 2025-02-04
FFS-2403: Add service account promotion rake task

allthesignals:
## Ticket
[FFS-2403](https://jiraent.cms.gov/browse/FFS-2403)

## Changes
Adds a rake task to promote certain users to service account users and return their API keys

## Context for reviewers
We don't have a way to do this in any deployed environment, even through rails console, afaict!

## Acceptance testing

### DIFF
--- a/app/lib/tasks/users.rake
+++ b/app/lib/tasks/users.rake
@@ -1,5 +1,5 @@
 namespace :users do
-  desc "Redact data that is older than our retention policy"
+  desc "Promote an existing account to service account status, returning an access token"
   task promote_to_service_account: :environment do
     user = User.find(ENV["id"])
 
@@ -10,6 +10,7 @@ namespace :users do
     end
   end
 
+  desc "Demote an existing account, deleting all associated access tokens"
   task demote_service_account: :environment do
     user = User.find(ENV["id"])
 
--- a/app/lib/tasks/users.rake
+++ b/app/lib/tasks/users.rake
@@ -1,0 +1,22 @@
+namespace :users do
+  desc "Redact data that is older than our retention policy"
+  task promote_to_service_account: :environment do
+    user = User.find(ENV["id"])
+
+    if user && user.update(is_service_account: true)
+      token = user.api_access_tokens.create
+
+      puts "User #{ENV["id"]} (#{user.email_address}) is now a service account: #{token.access_token}"
+    end
+  end
+
+  task demote_service_account: :environment do
+    user = User.find(ENV["id"])
+
+    if user && user.update(is_service_account: false)
+      user.api_access_tokens.destroy_all
+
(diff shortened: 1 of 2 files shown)


app/lib/tasks/users.rake
```diff
@@ -0,0 +1,22 @@
+namespace :users do
+  desc "Redact data that is older than our retention policy"
```
allthesignals, line 2: Crap need to change this

app/lib/tasks/users.rake
```diff
@@ -0,0 +1,23 @@
+namespace :users do
+  desc "Promote an existing account to service account status, returning an access token"
+  task promote_to_service_account: :environment do
+    user = User.find(ENV["id"])
```
tdooner, line 4: Fun fact, you can actually pass arguments to Rake tasks. The syntax is bonkers, though, so I think the way you did it is better.

The syntax is something like this:

```ruby
task :promote_to_service_account, [:user_id] => :environment do |_t, args|
  user = User.find(args[:user_id])
end
```

and then you have to call the rake task like:

```
bin/rails "users:promote_to_service_account[user_id=123]"
```

app/lib/tasks/users.rake
```diff
@@ -0,0 +1,23 @@
...
+      puts "User #{ENV["id"]} (#{user.email}) is now a service account: #{token.access_token}"
+    end
+  end
+
+  desc "Demote an existing account, deleting all associated access tokens"
+  task demote_service_account: :environment do
+    user = User.find(ENV["id"])
+
+    if user && user.update(is_service_account: false)
+      user.api_access_tokens.destroy_all
```
tdooner, line 18: Uhhh, are we doing soft deletion for these? Probably should just set `deleted_at` here. Or, we should remove the `deleted_at` column.

tdooner (approved):

app/lib/tasks/users.rake
allthesignals: Yeah, I tried that and thought the syntax was too weird. If you dont put double quotes it breaks in zsh lol

app/lib/tasks/users.rake
allthesignals: Oh duh. Sorry. Yep, you're right. Wondering if we should somehow protect against this kind of action.

allthesignals:
@tdooner I think I have a memory leak with the tests here. I'll try re-running all the jobs but seems fishy to me. I think the test db isn't being destroyed... or I'm doing something else wrong.

app/lib/tasks/users.rake
allthesignals: another thing about that syntax... is that an implicit hash map as the second argument? 

```ruby
{ [:user_id] => :environment }
```

PR #437, opened 2025-02-05
2401: Use different syntax

allthesignals:
## Ticket
N/A

## Changes
Simplifies the syntax (sort of) so that it's easier to run these rake commands against a deployed environment through our infra platform tooling

## Context for reviewers
Instead of
```
bin/run-command --environment-variables '[{ "name" : "id", "value" : "1" }]' app env '["bin/rails", "users:promote_to_service_account"]'
```

It's
```
`bin/run-command app env '["bin/rails", "users:promote_to_service_account[123]"]'`
```

## Acceptance testing

### DIFF
--- a/app/spec/lib/users_rake_spec.rb
+++ b/app/spec/lib/users_rake_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe "users.rake" do
   describe "promotes and demotes user service accounts" do
     it "promotes a user to a service account" do
       user = create(:user)
-      Rake::Task['users:promote_to_service_account'].invoke(user.id.to_s)
+      Rake::Task['users:promote_to_service_account'].execute(user.id.to_s)
       user.reload
       expect(user.is_service_account).to eq(true)
       expect(user.api_access_tokens.count).to eq(1)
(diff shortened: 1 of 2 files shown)


app/spec/lib/users_rake_spec.rb
```diff
@@ -9,17 +9,15 @@
...
       user.reload
       expect(user.is_service_account).to eq(true)
       expect(user.api_access_tokens.count).to eq(1)
     end
 
     it "demotes a service account user and removes all access tokens" do
       user = create(:user, :with_access_token, is_service_account: true)
-      ENV["id"] = user.id.to_s
-      Rake::Task['users:demote_service_account'].execute
+      Rake::Task['users:demote_service_account'].invoke(user.id.to_s)
```
tdooner, line 20: Do these need to pass in the arguments as keywords (like `.invoke(user_id: user.id.to_s)`)?

tdooner (approved):

app/spec/lib/users_rake_spec.rb
allthesignals: nope they're positional when i tested it...

allthesignals:
Huh, these tests are failing again. flaky.

PR #438, opened 2025-02-06
FFS-2110: Update flow to read/write indexing data to CbvApplicant

tdooner:
## Ticket

Resolves [FFS-2110](https://jiraent.cms.gov/browse/FFS-2110).

## Changes
**Rename CbvClient -> CbvApplicant**
**Write CbvApplicant fields directly from Caseworker form**
* Move all validations and logic for indexing fields onto CbvApplicant.
* Use `accepts_nested_attributes_for` and `fields_for` to forward <form>
  inputs directly into that object
* Add `site_id` to CbvApplicant since it will be a pain to go backwards
  to either the CbvFlow or CbvFlowInvitation to get the site_id in all
  of the places it's used.
* Write all the fields to CbvFlowInvitation as well, just for
  safekeeping until we drop them from there.

Test updates:
* Update factory for CbvFlow and CbvFlowInvitation to also create a
  cbv_applicant. Tests can override the fields of the applicant by
  passing in `cbv_applicant_attributes` to the factory when creating
  either the flow or the invitation.

## Context for reviewers

Uhhh, this is a pretty big refactor. Planning to get some more eyes on this for smoke testing before merging.

TODO list (as of 2/6):
- [x] Associate the CbvApplicant to CbvFlow when beginning the flow from invitation
- [x] Update Data retention service to redact CbvApplicant properly (and its dependencies)
- [x] Un-pend the test in Data Retention service
- [x] Make the `has_one` associations to cbv_flow and cbv_invitation -> `has_many`
- [x] Remove case_number field copied onto CbvFlow
- [x] Make sure the db/schema.rb is accurate - I think I need to rename the indexes as well.
- [x] See if I can remove `CbvApplicant#set_site_id` since it should be provided by whatever creates the applicant?
- [x] Rebase and handle merge conflicts
- [x] Add field to analytics
- [x] Actually go through the flow and make sure CbvApplicant is being created properly
- [x] Make story to limit `site_id` to the authenticated API user in Api::InvitaitonsController
- [x] Make story to remove requirement that we have an EmailAddress

## Acceptance testing


app/app/controllers/api/invitations_controller.rb
```diff
@@ -4,13 +4,18 @@ class Api::InvitationsController < ApplicationController
...
+      .invite(cbv_flow_invitation_params, @current_user, delivery_method: nil)
 
-    if @cbv_flow_invitation.errors.any?
-      return render json: @cbv_flow_invitation.errors, status: :unprocessable_entity
+    errors = @cbv_flow_invitation.errors
+    if errors.delete(:cbv_applicant)
+      errors.merge!(@cbv_flow_invitation.cbv_applicant.errors)
+    end
+    if errors.any?
+      return render json: errors, status: :unprocessable_entity
```
allthesignals, line 15: my hope is that our future tooling around these kinds of attributes will make this unnecessary!

app/app/controllers/api/invitations_controller.rb
```diff
@@ -21,19 +26,27 @@ def create
...
+        :first_name,
+        :middle_name,
+        :last_name,
+        :site_id,
+        :client_id_number,
+        :case_number,
+        :snap_application_date,
+        :agency_id_number,
+        :beacon_id
+      ]
```
allthesignals, line 49: ditto, hoping this is the sort of thing we can pull from a source of truth (the future attributes table work)

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
allthesignals, line 1: i appreciate extending the refactor to this controller as well _but_ are we even continuing caseworker portals? i assume we'd want to?

(stray thought: would be good to have this frontend consume our api controller somehow.)

app/app/controllers/api/invitations_controller.rb
tdooner: unfortunately, even if this is a json column or something, we will probably still need to do this trickery to hoist the errors on the dependent object `cbv_applicant` up onto the `cbv_flow_invitation` object that we're actually validating

app/app/models/cbv_applicant.rb
```diff
@@ -0,0 +1,133 @@
...
+    validate :ma_snap_application_date_not_in_future
+  end
+
+  # NYC specific validations
+  with_options(if: :nyc_site?) do
+    validates :case_number, format: { with: NYC_CASE_NUMBER_REGEX, message: :invalid_format }
+    validates :client_id_number, format: { with: NYC_CLIENT_ID_REGEX, message: :invalid_format }
+    validate :nyc_snap_application_date_not_more_than_30_days_ago
+    validate :nyc_snap_application_date_not_in_future
+  end
```
allthesignals, line 41: future approach will make this nicer! will love having specific class definitions for these (according to that blog post you linked to)

app/app/models/cbv_flow.rb
```diff
@@ -20,7 +20,7 @@ def complete?
   def self.create_from_invitation(cbv_flow_invitation)
     create(
       cbv_flow_invitation: cbv_flow_invitation,
-      case_number: cbv_flow_invitation.case_number,
+      case_number: cbv_flow_invitation.cbv_applicant.case_number,
       site_id: cbv_flow_invitation.site_id
     )
   end
```
allthesignals, line 26: Will we still need this class method if it should now happen in the CbvApplicant?

app/app/controllers/api/invitations_controller.rb
allthesignals: aw mannn even witht [this](https://cmsgov.slack.com/archives/C066XS5543S/p1738866442857529?thread_ts=1738863147.598119&cid=C066XS5543S)?

allthesignals: otherwise it'd be helpful to have this kind of stuff on the model somehow

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
```diff
@@ -14,23 +14,24 @@ def new
...
       Rails.logger.error("Error inviting applicant: #{e.message}")
       flash[:alert] = t(".invite_failed",
-                        email_address: cbv_flow_invitation_params[:email_address],
+                        email_address: invitation_params[:email_address],
                         error_message: e.message)
       return redirect_to caseworker_dashboard_path(site_id: params[:site_id])
     end
 
     if @cbv_flow_invitation.errors.any?
+      @cbv_flow_invitation.errors.delete(:cbv_applicant)
```
allthesignals, line 33: Interesting. During an error scenario, because it's a separate model, we have to delete it. I'm surprised there isn't a way to configure this as part of validations? Could you say:

```ruby
accepts_nested_attributes_for :asdf, :reject_if => :something_bad
```

app/app/controllers/help_controller.rb
```diff
@@ -12,16 +12,19 @@ def show
...
+        cbv_applicant_id: cbv_flow&.cbv_applicant_id,
+        cbv_flow_id: session[:cbv_flow_id],
+        invitation_id: cbv_flow&.cbv_flow_invitation_id,
+        site_id: current_site&.id,
+        flow_started_seconds_ago: cbv_flow ? (Time.now - cbv_flow.created_at).to_i : nil,
+        language: I18n.locale
+      })
+    rescue => ex
+      Rails.logger.error "Unable to track event (ApplicantViewedHelpTopic): #{ex}"
+    end
```
allthesignals, line 27: Wait, I don't understand why. Is it bad form to use that approach to rescues (on the method)?

app/app/models/cbv_applicant.rb
```diff
@@ -0,0 +1,140 @@
...
+    return if case_number.blank?
+    case_number.upcase!
+    if case_number.length == 9
+      self.case_number = "000#{case_number}"
+    end
+  end
+
+  def paystubs_query_begins_at
+    PAYSTUB_REPORT_RANGE.before(snap_application_date)
+  end
```
allthesignals, line 139: I can't really tell from the diff but this is all copied oer from cbv_invitation, right?

app/app/models/cbv_applicant.rb
```diff
@@ -0,0 +1,140 @@
...
+    validate :ma_snap_application_date_not_in_future
+  end
+
+  # NYC specific validations
+  with_options(if: :nyc_site?) do
+    validates :case_number, format: { with: NYC_CASE_NUMBER_REGEX, message: :invalid_format }
+    validates :client_id_number, format: { with: NYC_CLIENT_ID_REGEX, message: :invalid_format }
+    validate :nyc_snap_application_date_not_more_than_30_days_ago
+    validate :nyc_snap_application_date_not_in_future
+  end
```
allthesignals, line 40: Just noting this is slightly different from how it was written in cbv_flow_invitation... seems like nicer syntax for the same thing?

app/app/services/data_retention_service.rb
```diff
@@ -18,7 +18,10 @@ def redact_invitations
       .unstarted
       .unredacted
       .find_each do |cbv_flow_invitation|
-        cbv_flow_invitation.redact! if Time.now.after?(cbv_flow_invitation.expires_at + REDACT_UNUSED_INVITATIONS_AFTER)
+        next unless Time.now.after?(cbv_flow_invitation.expires_at + REDACT_UNUSED_INVITATIONS_AFTER)
+
+        cbv_flow_invitation.redact!
+        cbv_flow_invitation.cbv_applicant&.redact!
```
allthesignals, line 24: Aaaaah. Need a feature in redactable, like `has_redactable_fields (hi: string), dependent: :redact` or something

allthesignals (approved):
Amazing!! My only complaint is that we will continue to write those attributes to cbv_flow_invitation... should I make a ticket to remove that?

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
tdooner: You probably saw but [I asked about whether we should remove caseworker portals](https://cmsgov.slack.com/archives/C08BBUALWQN/p1738868606093559) prompted by this comment, and it seems like the answer is that we don't have plans to remove them.

app/app/models/cbv_applicant.rb
tdooner: Yeah, all were copied without modification. (I should have left a comment pointing that out, oops.)

app/app/controllers/help_controller.rb
tdooner: Rescue on a method is fine in general, but I didn't like it in this case because it might accidentally rescue exceptions for the other parts of the method that aren't related to event sending. It would make debugging the situation harder because we wouldn't get the actual exception reported to NewRelic.

app/app/models/cbv_applicant.rb
tdooner: Yeah, I tend to like using Rails's `with_options` to group together similar validation rules. The `with_options` helper will add whatever you pass it as an options hash to all the functions within. Helpful to make sure we don't forget to add the NYC conditional for any NYC validation rules.

app/app/services/data_retention_service.rb
tdooner: Yeah, was thinking about having the `Redactable` module redact associations, similar to how you had it. Would be good for a future change?

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
tdooner: This is only an issue because we iterate over the whole `@cbv_flow_invitation.errors` array to generate the error alert message at the top of the page (e.g. "5 Errors: [list of errors]").

With `cbv_applicant` in the `.errors` list, the count one-too-many and there is an error message for `cbv_applicant` that isn't meant to be user-facing.

tdooner:
> My only complaint is that we will continue to write those attributes to cbv_flow_invitation... should I make a ticket to remove that?

@allthesignals Already created - https://jiraent.cms.gov/browse/FFS-2463 !

PR #439, opened 2025-02-06
Fix flaky rake tests by always loading Rake tasks once

tdooner:
## Ticket

No ticket - flaky tests.

## Changes

* Fix flaky rake tests by always loading Rake tasks once

## Context for reviewers

If `Rails.application.load_tasks` is called multiple times, than any
invocation of a Rake task will actually happen multiple times. This is
problematic for non-idempotent Rake tasks (like
`users:promote_to_service_account`).

Let's prevent this issue by just loading the tasks in the
"rails_helper.rb" file so they're always available.

## Acceptance testing

### DIFF
--- a/app/spec/lib/backfills_spec.rb
+++ b/app/spec/lib/backfills_spec.rb
@@ -1,12 +1,7 @@
 require "rails_helper"
-require "rake"
 
 RSpec.describe "backfills.rake" do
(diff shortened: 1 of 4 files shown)


allthesignals (approved):
Amazing, thank you! Just tested by duplicating tests in my users spec file.

PR #442, opened 2025-02-07
FFS-2336: First sketch of multi-provider search

allthesignals:
## Ticket

Resolves [FFS-2336](https://jiraent.cms.gov/browse/FFS-2336).

## Changes
Integrates Argyle, implements a common multi-search interface, used in employer search

## Context for reviewers
Very rough draft of the provider search class and the approach to supporting multiple providers. There are nuances around SiteConfig, and how the provider (specifically pinwheel) is set up and accessed in controllers that I haven't figured out yet. 

## Acceptance testing


app/app/services/provider_search_service.rb
```diff
@@ -0,0 +1,35 @@
+class ProviderSearchService
+  PROVIDER_RESULT = Struct.new(:provider_name, :provider_options, :name, :logo_url, keyword_init: true)
+
+  def initialize(client_agency_id)
+    set_pinwheel(client_agency_id)
+    set_argyle(client_agency_id)
+  end
```
allthesignals, line 7: Need a way more elegant approach to this, not sure yet

allthesignals: These services should implement a common interface somehow but I'm not sure what that'd be yet.

app/app/services/provider_search_service.rb
```diff
@@ -0,0 +1,35 @@
...
+  def search(query = "")
+    [
+      *@argyle.items(query)["results"].map do |result|
+        PROVIDER_RESULT.new(provider_name: :argyle, provider_options: { response_type: result["kind"], provider_id: result["id"] }, name: result["name"], logo_url: result["logo_url"])
+      end,
+      *@pinwheel.fetch_items(q: query)["data"].map do |result|
+        PROVIDER_RESULT.new(provider_name: :pinwheel, provider_options: { response_type: result["response_type"], provider_id: result["id"] }, name: result["name"], logo_url: result["logo_url"])
+      end
+    ]
+  end
```
allthesignals, line 18: This is pretty bad because it doesn't deal with error failures or how to handle pagination!

app/app/controllers/cbv/employer_searches_controller.rb
```diff
@@ -20,6 +20,10 @@ def show
 
   private
 
+  def provider_search(query = "")
+    ProviderSearchService.new(@cbv_flow.site_id).search(query)
+  end
```
allthesignals, line 25: What are the memory implications of instantiating this every time? It's only really used for looking up some values specific to the partner agency configuration...

allthesignals: I guess it's only used in once place so it doesn't really matter right now

app/app/services/provider_search_service.rb
```diff
@@ -0,0 +1,68 @@
...
+          provider_options: {
+            response_type: result["kind"],
+            provider_id: result["id"]
+          },
+          name: result["name"],
+          logo_url: result["logo_url"]
+        )
+      end
+    end
+  end
```
allthesignals, line 67: Now after doing all of this I wonder if it makes sense just to have the services themselves assume a common interface. I don't know if you'd enforce it. There's no reason why `PROVIDER_RESULT` can't be used in the service classes themselves? 🤔

app/app/controllers/cbv/employer_searches_controller.rb
tdooner: Memory implications are probably not too bad, since GC should clear it out automatically after the method runs. There is a tiny performance hit in allocating and then GCing it, but whatever, not worth worrying about.

tdooner:
Looking good. What changes were you planning to make in response to reviewing it at Eng Time earlier today?

tdooner (approved):

PR #444, opened 2025-02-11
FFS-2337: Multi-Provider Support in Javascript Front-end code

jeffcatania-usds:
## Ticket

Resolves [FFS-2337](https://jiraent.cms.gov/browse/FFS-2337).

## Changes

* Creates new "Adapter" classes to encapsulate income data provider sdk's
   * IncomeDataAdapter = an abstract class that is used to wrap a provider SDK
   * PinwheelIncomeDataAdapter = extends IncomeDataAdapter to implement Pinwheel SDK
   * createIncomeDataAdapter() =  a factory method for selecting the right IncomeDataAdapter. Currently, only Pinwheel is allowed.
* Replaces Pinwheel logic inside employer_search controller with the generic IncomeDataAdapter
* Assumes a new "data-provider" attribute on the HTML tag that is used to select the provider.  Currently this should be hardcoded "pinwheel"
* Introduce global mocks into the test_setup.js file for API and load-script to reduce complexity of testing

## Context for reviewers
* This PR is primarily written in Typescript to take advantage of its Object-Oriented features

## Acceptance testing


app/app/javascript/adapters/PinwheelIncomeDataAdapter.ts
```diff
@@ -0,0 +1,100 @@
...
+        this.Pinwheel = await new Promise((resolve, reject) => {
+            loadScript('https://cdn.getpinwheel.com/pinwheel-v3.0.js', (err, script) => {
+                if (err) {
+                    reject(err);
+                } else {
+                    resolve(Pinwheel);
+                }
+            });
+        })
+    }
```
allthesignals, line 23: So, we don't have linting for this, but we've been using space-based tabs...

app/app/javascript/controllers/cbv/employer_search.js
```diff
@@ -14,14 +12,29 @@ export default class extends Controller {
     cbvFlowId: Number
   }
 
-  pinwheel = loadPinwheel();
+  initialize() {
+    const { responseType, id, name, isDefaultOption, providerName } = this.element.dataset;
+    const IncomeDataAdapter = createIncomeDataAdapter(providerName);
```
allthesignals, line 17: I see, this is the factory pattern. But it's a factory that produces instances of adapters :P

app/app/javascript/controllers/cbv/employer_search.js
```diff
@@ -14,14 +12,29 @@ export default class extends Controller {
...
+      requestData: {
+        responseType,
+        id,
+        isDefaultOption,
+        providerName,
+        name
+      },
+      onSuccess: this.onSuccess.bind(this),
+      onExit: this.onExit.bind(this)
+    })
```
allthesignals, line 29: fab

app/app/javascript/utilities/api.js
```diff
@@ -15,5 +15,4 @@ export const fetchToken = (response_type, id, locale) => {
     method: 'post',
     body: JSON.stringify({ response_type, id, locale }),
   })
-};
-
+};
```
allthesignals, line 18: Seeing stuff like this... is this your IDE removing eols?

app/app/javascript/utilities/getDocumentLocale.js
```diff
@@ -0,0 +1,7 @@
+export const getDocumentLocale = () => {
+  const docLocale = document.documentElement.lang;
+  if (docLocale) return docLocale;
+  // Extract locale from URL path (e.g., /en/cbv/employer_search)
+  const pathMatch = window.location.pathname.match(/^\/([a-z]{2})\//i);
+  return pathMatch ? pathMatch[1] : 'en';
+};
```
allthesignals, line 7: Yeah, seems like we switch to space-based tabbing.

app/app/javascript/adapters/PinwheelIncomeDataAdapter.ts
allthesignals: I'd say the rest of the code base is space-based and only uses 2

app/app/javascript/adapters/IncomeDataAdapter.ts
allthesignals, line 1: amazing

app/app/javascript/adapters/IncomeDataAdapter.ts
```diff
@@ -0,0 +1,43 @@
...
+    name: string;
+}
+export abstract class IncomeDataAdapter {
+    requestData: RequestData;
+    successCallback?: Function;
+    exitCallback?: Function;
+
+    abstract load();
+    abstract open(): void;
+    abstract onEvent(eventName: string, eventPayload: any): void;
```
allthesignals, line 22: Whoa, cool! Was not aware of this feature.

allthesignals: Are these TS features or ES6?

allthesignals (commented):
Sorry, I have more comments coming, just haven't had time to dive in

app/app/javascript/adapters/IncomeDataAdapter.ts
jeffcatania-usds: I think its TS only

app/app/javascript/adapters/PinwheelIncomeDataAdapter.ts
jeffcatania-usds: George is working on a fix for Javascript linting. we haven't had anything in our codebase to enforce.

app/app/javascript/adapters/PinwheelIncomeDataAdapter.ts
```diff
@@ -0,0 +1,102 @@
+import { fetchToken, trackUserAction } from "../utilities/api";
+import loadScript from 'load-script';
+import { getDocumentLocale } from "../utilities/getDocumentLocale";
+import { IncomeDataAdapter } from "./IncomeDataAdapter";
+
+declare global {
+    var Pinwheel: any;
```
GeorgeCodes19, line 7: We should probably add a TS rule to not use `any` since it pretty much defeats the purpose of using the type system. Instead we should opt for `unknown` since it'll force us to do some type of type-casting or conditional logic before handling the data. Otherwise let's explicitly define the `Pinwheel` type/interface here.

app/app/javascript/controllers/cbv/employer_search.js
```diff
@@ -1,7 +1,5 @@
 import { Controller } from "@hotwired/stimulus"
-import { loadPinwheel, initializePinwheel } from "../../utilities/pinwheel"
-import { fetchToken } from '../../utilities/api';
-import { trackUserAction } from '../../utilities/api';
+import { createIncomeDataAdapter } from "../../utilities/createIncomeDataAdapter";
```
GeorgeCodes19, line 2: I just realized that there's no `tsconfig` file in this project. Adding a tsconfig file and updating the path alias config ([similar to the vitest.config.js](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/vitest.config.js)) means that we won't have to do this directory traversal while importing.

i.e. `@js/utilities/createIncomeDataAdapter`

app/app/javascript/adapters/PinwheelIncomeDataAdapter.ts
```diff
@@ -0,0 +1,102 @@
...
+                if (err) {
+                    reject(err);
+                } else {
+                    resolve(Pinwheel);
+                }
+            });
+        })
+    }
+    
+    async open() {
```
GeorgeCodes19, line 25: It took me a while to figure out the context of `open`. Would a more appropriate name be `openModal`?

app/spec/javascript/controllers/employer_search.spec.js
```diff
@@ -0,0 +1,159 @@
+import { vi, describe, beforeEach, it, expect } from 'vitest'
+import EmployerSearchController from '@js/controllers/cbv/employer_search'
+import { fetchToken, trackUserAction } from '@js/utilities/api';
+import loadScript from "load-script";
+import { mockPinwheel, mockPinwheelAuthToken } from '@test/fixtures/pinwheel.fixture';
+import { fetchToken, trackUserAction } from '@js/utilities/api';
```
GeorgeCodes19, line 6: Duplicate imports. Line 3 above is also importing the same thing

app/spec/javascript/controllers/employer_search.spec.js
```diff
@@ -0,0 +1,159 @@
...
+
+describe('EmployerSearchController', () => {
+    let stimulusElement;
+
+    beforeEach(() => {
+        stimulusElement = document.getElementById('employer-search-button')
```
GeorgeCodes19, line 12: Is there a reason why we're getting a button by id and then immediately overriding it with a new button?

app/spec/javascript/test_setup.js
```diff
@@ -24,14 +30,35 @@ window.matchMedia = vi.fn().mockImplementation(query => ({
...
+      ...apiModule,
+      trackUserAction: vi.fn((eventName, eventPayload) => Promise.resolve()),
+      fetchToken: vi.fn(() => Promise.resolve(mockPinwheelAuthToken)),
+  }
+})
+
 // Reset all mocks before each test
 beforeEach(() => {
+  // Set up Stimulus
+  window.Stimulus = Application.start();
```
GeorgeCodes19, line 52: Nice. Prevents memory leaks

GeorgeCodes19 (changes requested):
There's a few optimizations that we can make as well as resolve some things like duplicate imports and proper TS configuration. We should definitely do away with using the `any` type.

app/app/javascript/adapters/PinwheelIncomeDataAdapter.ts
allthesignals: I wonder if a better name for these adapters would be something like `PinwheelModalAdapter` or something. Then it would be okay to keep it as `open`.

jeffcatania-usds: used matt's suggested name

app/app/javascript/controllers/cbv/employer_search.js
jeffcatania-usds: Good catch! that line was legacy code.

app/spec/javascript/controllers/employer_search.spec.js
jeffcatania-usds: removed

app/app/javascript/adapters/PinwheelIncomeDataAdapter.ts
jeffcatania-usds: fixed

app/app/javascript/adapters/ModalAdapter.ts
```diff
@@ -0,0 +1,54 @@
...
+    onExit?: Function;
+}
+
+interface RequestData {
+    responseType: string;
+    id: string;
+    isDefaultOption: boolean;
+    providerName: string;
+    name: string;
+}
```
tdooner, line 14: Is there a best practice on putting types in the same file vs having a `.d.ts` that contains them?

app/app/javascript/adapters/PinwheelModalAdapter.ts
```diff
@@ -0,0 +1,98 @@
...
+                onEvent: this.onEvent.bind(this),
+            });
+        }
+    }
+
+
+    async onSuccess(eventPayload: {
+        accountId: string;
+        platformId: string;
+    }) {
```
tdooner, line 48: Hmm, maybe we should make a `PinwheelEventPayload` type?

app/app/javascript/controllers/cbv/employer_search.js
```diff
@@ -14,14 +12,21 @@ export default class extends Controller {
     cbvFlowId: Number
   }
 
-  pinwheel = loadPinwheel();
+  initialize() {
+    const { providerName } = this.element.dataset;
+    const ModalAdapter = createModalAdapter(providerName);
```
tdooner, line 17: This reads a bit weird to me since `ModalAdapter` here is actually going to be `PinwheelModalAdapter`, not the `ModalAdapter` base class (as the name would imply).

Not sure what the idiomatic JS approach is for this, but I could imagine either:
```
const modalClass = createModalAdapter(providerName)
this.adapter = new modalClass()
```
or 
```
// createModalAdapter is updated to return an instance of the class, then:
this.adapter = createModalAdapter(providerName)
```

app/app/javascript/utilities/createModalAdapter.ts
```diff
@@ -0,0 +1,9 @@
+import PinwheelModalAdapter from "../adapters/PinwheelModalAdapter.js";
+
+
+export const createModalAdapter = (providerName: string) => {
+    switch (providerName) {
+        default:
```
tdooner, line 6: Could we make this "pinwheel" instead of `default`? And raise some kind of JS error if the `providerName` is unsupported?

app/spec/javascript/adapters/PinwheelIncomeDataAdapter.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        mockPinwheel();
+        adapter = new PinwheelModalAdapter()
+        adapter.init(pinwheelModalAdapterArgs)
+        triggers = await adapter.open()
+    })
+    afterEach(() => {
+    })
+
+    describe('open', () => {
+        it('should call track user action', async () => {
```
tdooner, line 37: nit: Not a fan of `should` in test names, personally, since they just waste space. I prefer, e.g.,

```javascript
it('calls track user action', ...)
```

app/spec/javascript/adapters/PinwheelIncomeDataAdapter.spec.js
```diff
@@ -0,0 +1,107 @@
...
+            expect(trackUserAction.mock.calls[1][0]).toBe('PinwheelError')
+            expect(trackUserAction.mock.calls[1][1]['type']).toBe('error-type')
+        })
+        it('should log exit event', async () => {
+            await triggers.triggerEvent("exit")
+            expect(trackUserAction).toHaveBeenCalledTimes(2)
+            expect(trackUserAction.mock.calls[1][0]).toBe('PinwheelCloseModal')
+        })
+    })
+})
```
tdooner, line 107: This is awesome we have tests like this now!

tdooner (approved):
I didn't review the tests very thoroughly, and may have missed something since I'm not as experienced at JS setup, but, with those caveats, it looks good to me. No blockers, but some suggestions to think about.

Do we want to run the tests automatically in CI? (I think we probably should!) Perhaps we can make a follow-up ticket for that.

app/app/javascript/adapters/ModalAdapter.ts
jeffcatania-usds: i can't comment on your overall PR comment - but javascript tests are already ru in CI.  "[Vitest Javascript Unit Tests / Vitest](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/13422081838/job/37496850561?pr=444)"

app/spec/javascript/adapters/PinwheelIncomeDataAdapter.spec.js
jeffcatania-usds: fixed

app/app/javascript/controllers/cbv/employer_search.js
jeffcatania-usds: Good call.  I used your second option.  This weirdness was due to my initial solution was to pass args into the modalClass, but that got refactored away. So simplest solution here is best.

app/app/javascript/adapters/ModalAdapter.ts
jeffcatania-usds: i moved this to ModalAdapter.types.ts

app/app/javascript/adapters/PinwheelModalAdapter.ts
jeffcatania-usds: I grabbed Pinwheel's actual type file.  Using it's LinkResult type

app/app/javascript/utilities/createModalAdapter.ts
jeffcatania-usds: I don't see any other error throws in the JS code. What should the default behavior be?  Redirect to /500.html?  Will we have language specific implementations of this?  Does something need to be logged to the track user action endpoint?

tdooner: Not sure the best error behavior, since we don't have any frontend error-tracking at all right now. `new Error(...)` would be fine with me for now, but I'm not sure whether there's a better pattern.

PR #450, opened 2025-02-12
Patching vulnerabilities | esbuild, net-imap

GeorgeCodes19:
## Ticket

## Acceptance testing

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -131,7 +131,9 @@ RUN apt-get update -qq && \
 # Remove base rexml for GHSA-4xqq-m2hx-25v8
 RUN gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 debug && \
     gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 rss && \
-    gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 rexml
+    gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 rexml && \
+    gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 net-imap || true && \
+    gem install net-imap -v 0.4.19
 
 # Install custom db migrate script
 COPY bin/db-migrate /usr/bin/
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -128,19 +128,16 @@ RUN apt-get update -qq && \
     ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws && \
     rm -rf ./awscli-bundle awscli-bundle.zip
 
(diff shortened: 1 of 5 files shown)


app/Dockerfile
```diff
@@ -130,7 +131,9 @@ RUN apt-get update -qq && \
 # Remove base rexml for GHSA-4xqq-m2hx-25v8
 RUN gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 debug && \
     gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 rss && \
-    gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 rexml
+    gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 rexml && \
+    gem uninstall -i /usr/local/lib/ruby/gems/3.3.0 net-imap || true && \
+    gem install net-imap -v 0.4.19
```
tdooner, line 136: I wonder if it works without the `gem install` here?

app/Gemfile
```diff
@@ -72,6 +72,7 @@ gem "stackprof"
 gem "rexml", "~> 3.3.9"
 gem "gpgme", "~> 2.0", ">= 2.0.12"
 gem "pdf-reader", "~> 2.12.0"
+gem "net-imap", "~> 0.4.19"  # Fixing CVE-2025-25186
```
tdooner, line 75: Probably shouldn't need to do this - the updated version in the lock file should be sufficient.

tdooner (approved):

app/Gemfile
GeorgeCodes19: Okay interesting. So the `net-imap` entry here doesn't result in the `Gemfile.lock` being updated? I thought the lock file was dependent on the contents of the Gemfile.

app/Dockerfile
GeorgeCodes19: I think we can do away with all of this.

app/Gemfile
GeorgeCodes19: Gotcha! Yeah, let's pin this version.

tdooner: `bundle update` will update every gem in the lockfile to the most recent version that satisfies all constraints. Since there wasn't previously a constraint for this gem, I would have expected `bundle update net-imap` to have upgraded its version to 0.4.19 in the lockfile. (Sometimes, there are constraints in the dependencies of other gems that may prevent bundler from doing an upgrade like that, though.)

Also, side note, but when upgrading for security reasons, we should use `>=` rather than `~>` since `~> 0.4.19` means "allow versions 0.4.x". We don't have the need to actually prevent an upgrade to 0.5 of this gem, hence `>= 0.4.19` being slightly preferable.

PR #451, opened 2025-02-13
Split CbvApplicant into three STI subclasses

tdooner:
## Ticket

Related to [FFS-2456](https://jiraent.cms.gov/browse/FFS-2456). Maybe not completely solving the ultimate purpose of that ticket, but I think this will go far enough towards solving it that we can probably close that ticket out until we hit more pain from these agency-specific configurations.

## Changes

**Split CbvApplicant into three Single-Table Inheritance (STI) subclasses**

These subclasses contain the complexity for validating MA/NYC/Sandbox
sites. The base class contains shared functionality. The underlying
`cbv_applicants` table still contains all the columns for all the
subclasses :-/.

## Context for reviewers

[Discussed via Slack](https://cmsgov.slack.com/archives/C066XS5543S/p1739469255741429) and it seems like we agree on this approach. So this is mergeable now (2/14 ❤️ ).

## Acceptance testing


GeorgeCodes19 (approved):
Looks good to me. I'm a big fan of this approach. 

I can see how things can get a bit confusing with using helper methods such as a `agency_translation` while having multiple language file entries 

```
 pdf:
          agency_header_name:
            ma: Massachusetts Department of Transitional Assistance
            nyc: NYC Human Resources Administration
            sandbox: CBV Test Agency
```

But also having the STI convention

```
cbv_applicant/ma:
          attributes:
            agency_id_number:
              blank: Enter a valid agency ID number.
              invalid_format: Agency ID number must be 7 digits.
            beacon_id:
              invalid_format: Your WELID must be 6 characters. It can only include letters and numbers.
            snap_application_date:
              invalid_date: Enter today's date or the date you contacted the client. This date must be today or in the past year.
        cbv_applicant/nyc:
```

app/config/locales/en.yml
```diff
@@ -4,24 +4,41 @@ en:
...
+        cbv_applicant/ma:
           attributes:
             agency_id_number:
               blank: Enter a valid agency ID number.
               invalid_format: Agency ID number must be 7 digits.
             beacon_id:
               invalid_format: Your WELID must be 6 characters. It can only include letters and numbers.
+            snap_application_date:
+              invalid_date: Enter today's date or the date you contacted the client. This date must be today or in the past year.
+        cbv_applicant/nyc:
```
allthesignals, line 23: my MDR sense is that this is "scary," but IDk why — it's unusual. What does it work?

allthesignals (approved):

app/config/locales/en.yml
```diff
@@ -3,31 +3,31 @@ en:
...
     errors:
       models:
+        cbv_applicant:
+          attributes:
+            first_name:
+              blank: Enter the client's first name.
+            last_name:
+              blank: Enter the client's last name.
+            snap_application_date:
+              invalid_date: Enter today's date or the date you contacted the client.
```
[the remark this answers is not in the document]
tdooner: Yep, it's the default used when an STI subclass doesn't override it (as `cbv_applicant/ma` and `cbv_applicant/nyc` both do.)

app/config/locales/en.yml
tdooner: The Rails i18n gem is mysterious and important. In this case, it's because the [validation error message lookup logic](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/error.rb#L78-L81) is aware of potential STI ancestors and tries multiple different keys until it finds one that matches.

allthesignals: ahhh. i see. the inheriting model serializes its name with the slash. fascinating thx

tdooner:
r.e. @GeorgeCodes19 
> I can see how things can get a bit confusing with using helper methods such as a agency_translation while having multiple language file entries

Yeah, this is a second pattern for have agency-specific translation overrides. However, we already had this pattern because the `snap_application_date` helper had different keys for different agencies. I don't love having a second pattern for this, but at least it does match the agencies' attributes pretty cleanly.

PR #453, opened 2025-02-13
Adding pre-commit hook for linting

GeorgeCodes19:
## Ticket

Resolves [FFS-2488](https://jiraent.cms.gov/browse/FFS-2488).

## Changes

- Updated pre-commit hook to utilize the package.json `lint` command e.g. `npm run lint`
- Improved error handling for commit message formatting with ticket number
- Swapped out Ruby `standard` library in favor of `rubocop` for linting

## Acceptance testing

- [X] No acceptance testing needed

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -20,14 +20,7 @@ if [ $linter_status -ne 0 ]; then
 fi
 
 echo "Running JavaScript/TypeScript formatter..."
-# Get all staged .js, .jsx, .ts, and .tsx files from specific directories
-js_files=$(git diff --cached --name-only --diff-filter=ACMR "app/app/javascript/**/*.js" "app/app/javascript/**/*.jsx" "app/app/javascript/**/*.ts" "app/app/javascript/**/*.tsx" "app/spec/javascript/**/*.js" "app/spec/javascript/**/*.jsx" "app/spec/javascript/**/*.ts" "app/spec/javascript/**/*.tsx" | sed 's| |\\ |g')
-if [ -n "$js_files" ]; then
-    # Format the files
-    cd app && npx prettier --write $js_files
-    # Add the formatted files back to staging
-    git add $js_files
-fi
+cd app && npm run format
 
 echo "Running Terraform formatter"
 files=$(git diff --cached --name-only terraform)
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -10,14 +10,23 @@
 # $ git commit --no-verify
 # $ git commit -n
 
+# Check if there are any staged changes
+if git diff --cached --quiet; then
+    echo "No changes to commit"
+    exit 0
+fi
+
 echo "Running linter..."
-bundle exec rake standard
+(cd app && bundle exec rubocop --autocorrect-all)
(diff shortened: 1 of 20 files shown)


.githooks/pre-commit
```diff
@@ -10,15 +10,27 @@
...
 linter_status=$?
 
 if [ $linter_status -ne 0 ]; then
-    echo "Fix above before committing. Run 'git commit -n' to bypass linter."
+    echo "Fix remaining linting issues before committing. Run 'git commit -n' to bypass linter."
     exit 1
 fi
 
+# Stage the files that were auto-corrected by RuboCop
+git add app/**/*.rb
```
tdooner, line 29: This is unexpected to me, and I'd guess will cause people to inadvertently commit files. Also, this script should really make sure it's running in the right directory, ideally by doing something like:

```bash
cd "$(git rev-parse --show-toplevel)"
```

at the top of the script. (This will cd to the top-level of the git repo.)

app/app/javascript/application.js
```diff
@@ -20,8 +20,8 @@ document.addEventListener("turbo:load", () => {
     const behavior = components[key]
     behavior.on(target)
   })
-});
+})
 
-document.addEventListener("turbo:frame-render", () => {
-  initialLoad = true;
-});
+document.addEventListener('turbo:frame-render', () => {
```
tdooner, line 25: Hmm, can we just standardize on double-quotes in JS for consistency with Ruby? Is there a technical reason to use single quotes?

tdooner (approved):
Seems fine to me, in principle, since people have to opt into it by symlinking the git hook script.

Personally I like the [pre-commit framework](https://pre-commit.com/) which has a lot of wrapping logic to this stuff which makes it faster and easier to deal with. So I'll probably keep https://jiraent.cms.gov/browse/FFS-2488 open to track that work.

app/app/javascript/application.js
GeorgeCodes19: No technical reason. By convention I'm used to single quotes and only using double quotes for strings that require interpolation. It's simply a practice that I've used for years now.

allthesignals: I think string interpolation in JS uses tildes, right?

allthesignals (commented):
I kind of wish that was broken out a little more. There are a few things going on:

1. first run of lint rules against JS files which makes a lot of changes
2. introduction of linter tooling (prettier)
3. introduction of pre-commit hook

~i would first get it to lint as part of the CI pipeline over this pre-commit thing.~ Oh it already does, duh! Eh, I'd definitely split the pre-commit thing out so we can have some more discussion on it because the linted-up javascript is valuable.

app/app/javascript/application.js
GeorgeCodes19: Yeah, JS uses tildes. In other languages that differs (PHP for example uses double quotes)- it's just convention that I'm used to, but not married to. I don't mind setting  `"singleQuote": false`.

jeffcatania-usds:
I'm going to close this again in favor of https://github.com/DSACMS/iv-cbv-payroll/pull/488

PR #455, opened 2025-02-14
FFS-2336: Move paystub, identity, employment, income into a response object

allthesignals:
## Ticket

Follow up from https://github.com/DSACMS/iv-cbv-payroll/pull/442 to migrate existing "response object" definitions to /models, leveraging Struct over Active Resource.

## Context for reviewers
We needed a way to model API responses so we have explicit definitions for those things.

Our first attempt was in PinwheelService, where we leveraged ActiveResource (gem) to model these responses. We found this got us most of the way there with dot-notation accessor style and some clarity around the concepts. But they didn't help document the expected domain used throughout CBV.

This change moves the Paystub object into a ResponseObject that leverages Struct to enforce required fields. We define a `from_pinwheel` method to help map Pinwheel's values into our domain modeling

That said, our "domain modeling" here so far is just however Pinwheel does it. When Argyle comes in, we'll basically be twisting it to fit in to how we've deferred to Pinwheel's design. We should find a middle ground that makes the most sense to CBV.

## Acceptance testing

### DIFF
--- a/app/app/models/response_objects/paystub.rb
+++ b/app/app/models/response_objects/paystub.rb
@@ -1,0 +1,79 @@
+module ResponseObjects
+  Paystub = Struct.new(
+    :account_id,
+    :gross_pay_amount,
+    :net_pay_amount,
+    :gross_pay_ytd,
+    :pay_period_start,
+    :pay_period_end,
+    :pay_date,
+    :earnings,
+    :deductions,
+
+    keyword_init: true
+  ) do
+    def self.from_pinwheel(response_body)
+      new(
+        account_id: response_body["account_id"],
+        gross_pay_amount: response_body["gross_pay_amount"],
+        net_pay_amount: response_body["net_pay_amount"],
+        gross_pay_ytd: response_body["gross_pay_ytd"],
+        pay_period_start: response_body["pay_period_start"],
+        pay_period_end: response_body["pay_period_end"],
+        pay_date: response_body["pay_date"],
+        earnings: response_body["earnings"].map do |earning|
+          PaystubEarning.new(
+            category: earning["category"],
+            hours: earning["hours"],
+          )
+        end,
+        deductions: response_body["deductions"].map do |deduction|
+          PaystubDeduction.new(
+            category: deduction["category"],
+            amount: deduction["amount"],
+          )
+        end,
+      )
+    end
+
+    alias_attribute :start, :pay_period_start
+    alias_attribute :end, :pay_period_end
+
+    def hours
+      base_hours = earnings
+        .filter { |e| e.category != "overtime" }
+        .map { |e| e.hours }
+        .compact
+        .max
+      return unless base_hours
+
+      # Add overtime hours to the base hours, because they tend to be additional
+      # work beyond the other entries. (As opposed to category="premium", which
+      # often duplicates other earnings' hours.)
+      #
+      # See FFS-1773.
+      overtime_hours = earnings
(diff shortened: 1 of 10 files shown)


app/app/models/response_objects/paystub.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      base_hours + overtime_hours
+    end
+
+    def hours_by_earning_category
+      earnings
+        .filter { |e| e.hours && e.hours > 0 }
+        .group_by { |e| e.category }
+        .transform_values { |earnings| earnings.sum { |e| e.hours } }
+    end
```
allthesignals, line 67: I wonder if we will need to do these types of calculations in the `from_*` class methods soon...

app/app/models/response_objects/paystub.rb
```diff
@@ -0,0 +1,79 @@
...
+  Paystub = Struct.new(
+    :account_id,
+    :gross_pay_amount,
+    :net_pay_amount,
+    :gross_pay_ytd,
+    :pay_period_start,
+    :pay_period_end,
+    :pay_date,
+    :earnings,
+    :deductions,
```
tdooner, line 11: Maybe from a syntax/readability point of view we could put these in a constant, i.e.

```ruby
PAYSTUB_FIELDS = %i[
  account_id
  gross_pay_amount
  net_pay_amount
  gross_pay_ytd
  pay_period_start
  pay_period_end
  pay_date
  earnings
  deductions
]

Paystub = Struct.new(*PAYSTUB_FIELDS, keyword_init: true) do
  # ...
end
```

app/app/models/response_objects/paystub.rb
tdooner: Good point, I suppose these methods are intended to transform Pinwheel-specific data quirks. I suppose they probably should happen while it's being initialized from pinwheel response data.

tdooner (approved):
Sick

app/app/models/response_objects/paystub.rb
allthesignals: love it

app/app/models/response_objects/paystub.rb
```diff
@@ -0,0 +1,79 @@
...
+          PaystubDeduction.new(
+            category: deduction["category"],
+            amount: deduction["amount"],
+          )
+        end,
+      )
+    end
+
+    alias_attribute :start, :pay_period_start
+    alias_attribute :end, :pay_period_end
```
allthesignals, line 40: I don't see any reason to have this alias 🤷

allthesignals:
@tdooner okay, thanks. I went ahead and extended the refactor to the other objects! I removed ActiveResource entirely.

app/app/models/response_objects/paystub.rb
tdooner: Not sure I understand what you're saying? The naming discrepancy is a difference between the API response and how we use the data in the PDF rendering. The API field is `pay_period_start` and we use it in app/views/cbv/summaries/show.pdf.erb as `payment.start` and `payment.end`. Maybe worth standardizing that so we don't have two ways to refer to the same attributes?

allthesignals: @tdooner Oh I just meant that I'm not sure why we didn't just defer to Pinwheel's names in this case (but did almost everywhere else) and instead arbitrarily named them something else.

With this new pattern, however, we can do whatever we like!

PR #456, opened 2025-02-14
FFS-2397: Improve Mixpanel's user profile data

millerti:
## Changes
Mixpanel user profiles will now have accurate locations associated with them.

## Testing Instructions
0. Create a brand new invite
1. Proceed through the CBV flow enough to file an event 
2. Click on the Distinct ID in Mixpanel for your new event and observe the User Profile section. It should contain your correct location (or the location of your VPN). Previously, it was just wrong (the theory is that it was the location of the AWS server being used).
image

## Acceptance testing


tdooner (approved):
Cool, good stuff. At some point we should change the distinct ID from `invitation-#{id}` to `applicant-#{id}` (based on the CbvApplicant table).

PR #457, opened 2025-02-14
FFS-2336: Better Organize Response Objects

allthesignals:
## Ticket

Follow up from https://github.com/DSACMS/iv-cbv-payroll/pull/442 to prepare our new response objects for an additional provider.

## Changes
Extracts exactly the values we need from Pinwheel's API response in Employment. 

In Paystub, we pre-compute the hours and hours_by_earning_category values during instantiation because these response shapes are novel to Pinwheel.

## Acceptance testing

### DIFF
--- a/app/app/models/response_objects/employment.rb
+++ b/app/app/models/response_objects/employment.rb
@@ -1,0 +1,30 @@
+EMPLOYMENT_FIELDS = %i[
+  account_id
+  employer_name
+  start_date
+  termination_date
+  status
+  employer_phone_number
+  employer_address
+]
+
+module ResponseObjects
+  Employment = Struct.new(*EMPLOYMENT_FIELDS, keyword_init: true) do
+    def self.from_pinwheel(response_body)
+      new(
+        account_id: response_body["account_id"],
+        employer_name: response_body["employer_name"],
+        start_date: response_body["start_date"],
+        termination_date: response_body["termination_date"],
+        status: response_body["status"],
+        employer_phone_number: OpenStruct.new(
+          value: response_body["employer_phone_number"]["value"],
+          type: response_body["employer_phone_number"]["type"],
+        ),
+        employer_address: OpenStruct.new(
+          raw: response_body["employer_address"]["raw"],
+        ),
+      )
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


app/app/models/response_objects/employment.rb
```diff
@@ -17,13 +17,8 @@ def self.from_pinwheel(response_body)
...
         status: response_body["status"],
-        employer_phone_number: OpenStruct.new(
-          value: response_body["employer_phone_number"]["value"],
-          type: response_body["employer_phone_number"]["type"],
-        ),
-        employer_address: OpenStruct.new(
-          raw: response_body["employer_address"]["raw"],
-        ),
+        employer_phone_number: response_body["employer_phone_number"]["value"],
+        employer_address: response_body["employer_address"]["raw"]
```
allthesignals, line 21: We only ever use these deep nested values and don't need to model them

app/app/models/response_objects/paystub.rb
```diff
@@ -6,8 +6,9 @@
   pay_period_start
   pay_period_end
   pay_date
-  earnings
   deductions
+  hours_by_earning_category
+  hours
```
allthesignals, line 11: So, this maybe problematic since we're not modeling the API response anymore and doing calculations later down the road. Instead, we're modeling what a Paystub is _in the context of how it's used in CBV_. This is important for normalizing other providers' data into our implementation... but I could be swayed in some other direction! 

One idea is to use the decorator pattern to define methods like `hours_by_earning_category` that compute downstream of their respective model's instantiation. We'd still model the API response, do some light normalization. 🤷

app/app/models/response_objects/paystub.rb
```diff
@@ -39,31 +36,35 @@ def self.from_pinwheel(response_body)
...
 
-      # Add overtime hours to the base hours, because they tend to be additional
-      # work beyond the other entries. (As opposed to category="premium", which
-      # often duplicates other earnings' hours.)
-      #
-      # See FFS-1773.
-      overtime_hours = earnings
-        .filter { |e| e.category == "overtime" }
-        .sum { |e| e.hours || 0.0 }
+      def hours_from_pinwheel(earnings)
```
allthesignals, line 42: I've renamed these to be pinwheel specific because they're intended to take pinwheel-specific data and munge it into something we actually use...

tdooner (approved):

PR #459, opened 2025-02-18
FFS-2453: Replace Help Modal iframe with Turbo Frame

GeorgeCodes19:
## Ticket

Resolves [FFS-2453](https://jiraent.cms.gov/browse/FFS-2453).

## Changes
- Swapped out the iframe with a turbo frame
- Removed the CSS for the now deprecated iframe
- Removed the JS functionality for refreshing the iframe URL

## Acceptance testing

### DIFF
--- a/app/app/controllers/application_controller.rb
+++ b/app/app/controllers/application_controller.rb
@@ -92,11 +92,9 @@ class ApplicationController < ActionController::Base
   def check_help_param
     if params[:help] == "true"
       help_link = helpers.render(partial: "help/help_link", locals: { text: t("help.alert.help_options"), source: "banner" })
-      flash.merge!(
-        alert: "#{t('help.alert.text_before')} #{help_link}",
-        alert_heading: t("help.alert.heading"),
-        alert_type: "warning"
-      )
+      flash.now[:alert] = "#{t('help.alert.text_before')} #{help_link}"
+      flash.now[:alert_heading] = t("help.alert.heading")
+      flash.now[:alert_type] = "warning"
     end
   end
 end
--- a/app/app/javascript/controllers/help.js
+++ b/app/app/javascript/controllers/help.js
@@ -4,16 +4,17 @@ import { trackUserAction } from "../utilities/api";
 export default class extends Controller {
   static targets = ["content"];
 
+  handleClick(event) {
+    if (event.target.href?.includes("#help-modal")) {
+      trackUserAction("ApplicantOpenedHelpModal", {
+        source: event.target.dataset.source,
+      });
+      // reset the help modal src on mousedown to ensure the help modal src is reset to "/help"
+      document.querySelector("#help_modal_content").src = "/help";
+    }
+  }
+
   connect() {
-    this.handleClick = (event) => {
-      if (event.target.href?.includes("#help-modal")) {
-        trackUserAction("ApplicantOpenedHelpModal", {
-          source: event.target.dataset.source,
-        });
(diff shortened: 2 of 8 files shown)


app/app/controllers/application_controller.rb
```diff
@@ -92,11 +92,9 @@ def redirect_if_maintenance_mode
   def check_help_param
     if params[:help] == "true"
       help_link = helpers.render(partial: "help/help_link", locals: { text: t("help.alert.help_options"), source: "banner" })
-      flash.merge!(
-        alert: "#{t('help.alert.text_before')} #{help_link}",
-        alert_heading: t("help.alert.heading"),
-        alert_type: "warning"
-      )
+      flash.now[:alert] = "#{t('help.alert.text_before')} #{help_link}"
```
GeorgeCodes19, line 95: `flash.now` prevents the "stickiness" of the flash message from page-to-page

app/app/javascript/controllers/help.js
```diff
@@ -11,30 +11,31 @@ export default class extends Controller {
...
+          }
+        }
+      })
+    })
+
+    const modalElement = document.getElementById('help-modal')
+    this.observer.observe(modalElement, {
+      attributes: true,
+      attributeFilter: ['class']
+    })
```
allthesignals, line 30: I'm confused, why is this here? Do we still use an iframe? Using MutationObserver seems pretty complicated for what this is... maybe we can find something wtihin the framework to solve whatever the issue?

GeorgeCodes19: We still need the ability for the modal contents to reset to the help topic selection upon closing. The USWDS modal doesn't do anything special such as emitting an event on close that we can listen to, but it does have CSS classes that change when the modal is visible vs not visible. This doesn't seem like anything crazy or code smell to me. It can be condensed down to around 10 lines of code and doesn't entail introducing more complexity on the Rails side.

I also spent time trying to figure out a backend solution here, but came up empty handed. The Mutation Observer API is pretty much universally adopted by browsers e.g. https://caniuse.com/?search=mutationobserver.

clediggins-usds (approved):
Paired, functionality is correct, looks good.

app/app/javascript/controllers/help.js
```diff
@@ -1,40 +1,29 @@
...
   connect() {
     this.handleClick = (event) => {
       if (event.target.href?.includes("#help-modal")) {
-        trackUserAction("ApplicantOpenedHelpModal", { source: event.target.dataset.source })
+        trackUserAction("ApplicantOpenedHelpModal", {
+          source: event.target.dataset.source,
+        });
       }
-    }
+    };
```
tdooner, line 14: Can this be made an instance method instead of defining it in the `connect` method here? It doesn't even appear to use `this` so I think it could even be defined as a static/class method, but I think defining it as an instance method makes more sense.

tdooner (approved):
Looks and works great!

app/app/javascript/controllers/help.js
```diff
@@ -1,40 +1,29 @@
...
+    
+    // reset the help modal src on mousedown to ensure the help modal src is reset to "/help"
+    document
+      .querySelector('[aria-controls="help-modal"]')
+      .addEventListener("mousedown", () => {
+        document.querySelector("#help_modal_content").src = "/help";
+      });
```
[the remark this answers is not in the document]
GeorgeCodes19: Good catch!

PR #460, opened 2025-02-19
FFS-2492: use applicant ID as distinct ID on Mixpanel

millerti:
## Changes
Use applicant ID as distinct ID on Mixpanel

## Testing Instructions
1. Log a few events
2. Log into Mixpanel and verify that the Distinct ID being used for your new events uses the Applicant ID

## Acceptance testing


app/app/controllers/cbv/base_controller.rb
```diff
@@ -90,6 +90,7 @@ def capture_page_view
       event_logger.track("CbvPageView", request, {
         cbv_flow_id: @cbv_flow.id,
         invitation_id: @cbv_flow.cbv_flow_invitation_id,
+        cbv_applicant_id: @cbv_flow.cbv_applicant_id,
```
tdooner, line 93: How will this work if the vast majority of the mixpanel events don't have the `cbv_applicant_id` field?

millerti: Well, it wouldn't. Maybe I'm misunderstanding how the applicant id works. I assumed that all people going through the cbv flow would have one, invited or not. But if that's not the case then maybe I should fall back to a distinct ID of "invite-#" when applicant ID is missing.

millerti: I've verified all the events have the applicant id

tdooner (approved):

PR #469, opened 2025-02-21
FFS-2377 Extend Session Modal

GeorgeCodes19:
## Ticket

Resolves [FFS-2377](https://jiraent.cms.gov/browse/FFS-2377).

- [x] Create modal for extending session
- [x] Update javascript page to have a timer with a warning threshold 5 mins prior to the application's configured session
- [x] Added base_controller method to check of requests to end session
- [x] Extend session cookie (refresh the current page)

Screenshot 2025-02-21 at 6 03 50 PM

## Context for reviewers

I explored several different approaches to resolving the ticket during this implementation. Most notably using `ActionCables` and websockets to poll the backend for session status and passing user state related props (`cbv_flow_id`, `last_activity_time`, etc.) to Stimulus controllers. This discovery process allowed me to become familiar with different parts of the Rails framework and better understand the pros/cons to each approach. Ultimately I opted for a solution with rather inconsequential modifications to the codebase.

This PR utilizes [Stimulus values](https://stimulus.hotwired.dev/reference/targets) in combination with [Stimulus controllers](https://stimulus.hotwired.dev/reference/controllers).

1. When a webpage loads the session timeout modal view partial is passed two locales - the `cbv_flow_id` from the session and the application configured `session time` e.g. `expire_after`. These two values are available as data values on the Stimulus controller.
2. Upon connecting (on each page load) the Stimulus `session_controller` initiates two timers. One timer is used to trigger the session warning modal via a hidden button that triggers the modal visibility. The other timer will redirect to the `cbv/entry?end_session=true&user_action=false` path if the user does not forcefully continue the current session. Opting to continue the session will simply result in a page reload. I figured this is the path of least resistance.
3. The `end_session` parameter is observed by a `check_session_expiry` before_action on the `cbv/base_controller`. The `check_session_expiry` is responsible for clearing the `cbv_flow` from the session, tracking a timeout event, and then redirecting the user to the entry page.

## Acceptance testing


app/app/javascript/controllers/cbv/sessions/timeout_controller.js
allthesignals, line 1: Nice, this makes sense

app/app/controllers/cbv/base_controller.rb
```diff
@@ -99,6 +99,16 @@ def capture_page_view
     end
   end
 
+  def check_session_expiry
```
GeorgeCodes19, line 102: Ehh... might be able to clean this up some, but it works fine.

app/app/javascript/controllers/session_controller.js
```diff
@@ -0,0 +1,50 @@
...
+    "modal",
+  ];
+
+  timeoutValue;
+  cbvFlowId;
+
+  connect() {
+    console.log("Session controller connected");
+
+    this.cbvFlowId = parseInt(this.modalTarget.dataset.itemCbvFlowIdParam, 10);
```
GeorgeCodes19, line 15: I'm going to clean this up. We no longer need these instance variables. I'm pretty sure that `this.modalTarget.dataset` can be referenced directly instead of configuring an instance variable. There's only one element utilizing this stimulus controller- the timeout modal element itself.

app/app/javascript/controllers/session_controller.js
```diff
@@ -0,0 +1,50 @@
...
+
+    // Set timer for the actual session expiration
+    this.expirationTimer = setTimeout(() => {
+      const locale = getDocumentLocale();
+      window.location = `/${locale}/cbv/entry?end_session=true&user_action=false`;
+    }, expirationDelay);
+  }
+
+  showWarning() {
+    if(!this.cbvFlowId) {
```
GeorgeCodes19, line 38: If there isn't a cbv_flow in the session then there's no need to show the modal.

app/app/views/cbv/sessions/_timeout_modal.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      <div class="usa-prose">
+        <p id="cbv-session-timeout-description">
+          <%= t("session_timeout.modal.description") %>
+        </p>
+      </div>
+      <div class="usa-modal__footer">
+        <ul class="usa-button-group">
+          <li class="usa-button-group__item">
+            <button type="button"
+                    onclick="location.reload()"
```
GeorgeCodes19, line 24: I figured just simply refreshing the page would suffice since all data is stored in the session and database. Is this an acceptable user experience? To simply refresh the current page?

app/app/views/layouts/application.html.erb
```diff
@@ -54,8 +54,11 @@
         </section>
       </main>
       <%= render "application/footer" %>
+      <%= render partial: "cbv/sessions/timeout_modal", locals: {
+        cbv_flow_id: session[:cbv_flow_id],
+        timeout: Rails.application.config.session_options[:expire_after]
+      } %>
```
tdooner, line 60: Perhaps we can only render the partial when there is a `session[:cbv_flow_id]`, rather than passing its id through to the modal?

app/app/views/cbv/sessions/_timeout_modal.html.erb
tdooner: Hmm, I don't think it will be acceptable. My experience on other sites with those modals is that they just do something behind the scenes to extend the session. Maybe we can make a simple "session extension" API endpoint (or blank page fetched via Turbo?) that does nothing except refreshing the cookie automatically due to it being requested.

app/app/controllers/cbv/base_controller.rb
```diff
@@ -99,6 +99,16 @@ def capture_page_view
...
+    if params[:end_session] == "true"
+      session[:cbv_flow_id] = nil
+      if params[:user_action] == "false"
+        track_timeout_event
+      end
+      redirect_to root_url
+    end
+  end
```
tdooner, line 110: Let's make a Cbv::SessionsController we can use for this functionality. It really should be a `POST`, or better yet `DELETE`, to prevent CSRF.

app/app/assets/stylesheets/cbv.scss
```diff
@@ -161,3 +161,11 @@ html {
     }
   }
 }
+
+/*
+* Session timeout modal
+*/
+.invisible {
+    visibility: hidden;
```
tdooner, line 169: Is there a reason it needs to be `visibility: hidden` vs `display: none`?

app/app/views/cbv/sessions/_timeout_modal.html.erb
```diff
@@ -0,0 +1,51 @@
+<div class="usa-modal"
+     data-controller="session"
+     data-session-target="modal"
+     data-item-cbv-flow-id-param="<%= cbv_flow_id %>"
+     data-item-timeout-param="<%= timeout %>"
+     data-force-action
+     id="session-timeout-modal"
+     aria-labelledby="session-timeout-heading"
+     aria-describedby="session-timeout-description">
```
tdooner, line 9: Good idea to include these (I don't know how/whether these modals will work with screen readers, which would be good for us to try at some point). But, the IDs here don't match the elements. They should be:

```suggestion
     aria-labelledby="session-heading"
     aria-describedby="cbv-session-timeout-description">
```

(or maybe update these to have a consistent prefix?)

tdooner (commented):
Looking good so far, but I suppose there are some nuanced behaviors here we didn't specify clearly at the beginning of the ticket.

app/app/views/layouts/application.html.erb
GeorgeCodes19: Yeah, good catch. I was too far zoomed in and missed this simple optimization.

app/app/assets/stylesheets/cbv.scss
GeorgeCodes19: I think `display:none` makes an element unclickable? I think that was an issue that I ran into early on, but now I'm curious to know.

app/app/views/cbv/sessions/_timeout_modal.html.erb
GeorgeCodes19: Thanks, good catch! I'll just update the prefixes to be consistent.

app/app/controllers/cbv/base_controller.rb
GeorgeCodes19: Hmmm... I'm seeing overlap with the [resets_controller](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/app/controllers/cbv/resets_controller.rb). Maybe we add an `end` method to the `SessionController` and do away with the `resets_controller.rb` file?

tdooner (approved):
Looks and works good to me! Let's get this accepted!

No blockers, but would like to see you address the comments related to:
* `display:none`
* Adding the check that the rake task you made isn't running in production
* Addressing comments in the `seed_cbv_user_flow_spec.rb`

app/lib/tasks/seed_cbv_user_flow.rake
```diff
@@ -0,0 +1,198 @@
...
+        user.api_access_tokens.destroy_all
+        user.destroy!
+      end
+      
+      puts "Successfully deleted CBV user flow and all associated records."
+    end
+  end
+
+  desc "Delete all CBV user flows and associated records"
+  task delete_all_cbv_user_flows: :environment do
```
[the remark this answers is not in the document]
GeorgeCodes19: Oh dang. I didn't even mean to commit this!

app/lib/tasks/seed_cbv_user_flow.rake
```diff
@@ -0,0 +1,198 @@
...
+      CbvApplicant.where(id: applicant_ids).destroy_all
+      
+      # Delete API tokens and users
+      ApiAccessToken.where(user_id: user_ids).destroy_all
+      User.where(id: user_ids).destroy_all
+      
+      puts "Successfully deleted #{cbv_flow_count} CBV user flows and all associated records."
+    end
+  end
+end
```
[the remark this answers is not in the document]
GeorgeCodes19: Sorry, referring to my other comment I wasn't even aware that I had committed this. I thought I un-staged it and then pushed the code up. It's simply meant to be a personal tool that I symlink into the project.

app/app/javascript/controllers/cbv/sessions_controller.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  handleModalOpen() {
+    // Find the pinwheel portal if it exists
+    const pinwheelPortal = document.querySelector(".pinwheel-portal");
+    if (pinwheelPortal) {
+      // Store the original z-index
+      this.originalPinwheelZIndex = pinwheelPortal.style.zIndex || getComputedStyle(pinwheelPortal).zIndex;
+      pinwheelPortal.style.zIndex = "0";
+    }
```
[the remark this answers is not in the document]
GeorgeCodes19: Thanks! CSS was a no-op here since the modal types are injected and stylized via JS

PR #470, opened 2025-02-25
FFS-2466: Rename PinwheelAccount model to PayrollAccount

allthesignals:
## Ticket

## Changes
In preparation for making this more generic we should rename this model to reflect that. 

## Context for reviewers
We still call the table `pinwheel_accounts` but I can add a migration to rename it if it helps...

## Acceptance testing


app/app/models/payroll_account.rb
```diff
@@ -1,4 +1,6 @@
-class PinwheelAccount < ApplicationRecord
+class PayrollAccount < ApplicationRecord
+  self.table_name = "pinwheel_accounts"
```
tdooner, line 2: Hmmm, we should rename the table at some point as well to prevent confusion.

tdooner (approved):

app/app/models/payroll_account.rb
allthesignals: I'll just add a migration for it here.

PR #472, opened 2025-02-26
FFS-2466: Rename cbv_flow identifier column for clarity

allthesignals:
## Ticket

## Changes
Renames cbv_flow identifier column so that it's clear it can be used simply to identify the flow

## Context for reviewers
It was confusing how this column was named. We use this column to identify which flow the incoming aggregator account is meant for.

## Acceptance testing


app/db/migrate/20240613154226_add_end_user_id_column_to_cbv_flows.rb
```diff
@@ -1,5 +1,5 @@
 class AddEndUserIdColumnToCbvFlows < ActiveRecord::Migration[7.1]
   def change
-    add_column :cbv_flows, :pinwheel_end_user_id, :uuid, default: 'gen_random_uuid()', null: false
+    add_column :cbv_flows, :end_user_id, :uuid, default: 'gen_random_uuid()', null: false
```
tdooner, line 3: In general, not a good idea to edit migrations that have already been run, since this would break the future migration.

tdooner (approved):
Lgtm, but, can you remove the old migration change before merging?

app/db/migrate/20240613154226_add_end_user_id_column_to_cbv_flows.rb
allthesignals: Oof, that's a mistake. My bad, thanks for catching

PR #473, opened 2025-02-27
2466: STI-ify PayrollAccounts

allthesignals:
## Ticket

## Changes
Convert PayrollAccounts to an STI with a Pinwheel subclass.

## Acceptance testing


app/app/models/payroll_account.rb
```diff
@@ -1,65 +1,19 @@
 class PayrollAccount < ApplicationRecord
+  def self.sti_name
+    # "PayrollAccount::Pinwheel" => "pinwheel"
+    name.demodulize.downcase
+  end
+
+  def self.sti_class_for(type_name)
+    # "pinwheel" => PayrollAccount::Pinwheel
+    PayrollAccount.const_get(type_name.capitalize)
+  end
```
allthesignals, line 10: Not sure this is necessary? Since the behavior is default

tdooner: I guess we can try removing it? It probably is not necessary if we are ok with the "type" column having "payroll_account/pinwheel" as the value rather than just "pinwheel".

tdooner (approved):

PR #474, opened 2025-02-27
FFS-2466: Infer supported jobs from within Pinwheel subclass

allthesignals:
## Ticket

Addresses [FFS-2466](https://jiraent.cms.gov/browse/FFS-2466).

## Changes
This makes the model infer the pinwheel accounts supported jobs through the provided pinwheel_account_id.

## Context for reviewers
We need to do this to make it easier to support multi-provider. The reason for this is that we want to create a payrol_acount earlier in the process, not when the webhook is received. In order to do that, we needed a tidier way to create payroll account objects. 

Before, we would look up the support jobs in the events controller and provide that to the model.

Now, we do that in an after_create hook. This means we can create payroll acounts anywhere without also needing to supply the supported jobs.

## Acceptance testing

### DIFF
--- a/app/app/models/payroll_account/pinwheel.rb
+++ b/app/app/models/payroll_account/pinwheel.rb
@@ -1,4 +1,6 @@
 class PayrollAccount::Pinwheel < PayrollAccount
+  after_create :lookup_supported_jobs
+
   EVENTS_MAP = {
     "employment.added" => :employment_synced_at,
     "income.added" => :income_synced_at,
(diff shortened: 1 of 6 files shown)


app/app/models/payroll_account/pinwheel.rb
```diff
@@ -54,4 +56,19 @@ def event_columns_for(job)
 
     [ error_column, sync_column ]
   end
+
+  def lookup_supported_jobs
+    platform_id = pinwheel.fetch_account(account_id: pinwheel_account_id)["data"]["platform_id"]
+    supported_jobs = pinwheel.fetch_platform(platform_id: platform_id)["data"]["supported_jobs"]
```
tdooner, line 62: (Hmm _thinks to self._ Maybe we should save `platform_id` in the table itself.)

app/app/models/payroll_account/pinwheel.rb
```diff
@@ -54,4 +56,19 @@ def event_columns_for(job)
 
     [ error_column, sync_column ]
   end
+
```
tdooner, line 59: Would be good to make these methods `private`

tdooner (approved):

tdooner:
FWIW, I don't love making API calls to 3rd party services on model creation. Probably not the best practice because it might make testing harder. But it seems ok for now, and we can refactor it later if it becomes annoying.

PR #475, opened 2025-02-27
FFS-2338: /payment_details content updates

millerti:
## Changes
Implemented copy updates on the [/payment_details](https://confluenceent.cms.gov/pages/viewpage.action?pageId=748887616) page.

## Testing Instructions
View the /payment_details page and make sure the copy is sensible and that the updates have been made. The changes are summarized by [this diff](https://confluenceent.cms.gov/pages/diffpagesbyversion.action?pageId=748887616&selectedPageVersions=21&selectedPageVersions=20).

## Acceptance testing

## Screenshots of changes
image
image


app/config/locales/es.yml
```diff
@@ -177,7 +174,6 @@ es:
         pay_period_value: "%{start_date} al %{end_date}"
         payment_hours: "%{amount} horas"
         payments_and_deductions_table_header: Pagos y deducciones
-        subheader: Hemos recopilado sus registros de pago de los últimos 90 días, del %{start_date} al %{end_date}. Si falta información o esta es inexacta, añada un comentario para %{agency_acronym}. Esta información se incluirá en su informe de ingresos.
```
tdooner, line 180: Removing these will cause the `i18n-tasks missing` check to fail, and I see why since I assume this means we want to update these translations to match the newly-translated English strings.

Will you be including these in the next translation request? Maybe as long as they are captured in the next translation request, we should just keep the old untranslated strings here?

tdooner (approved):

app/config/locales/es.yml
millerti: Yeah I wasn't quite sure what to do. I didn't want to forget to translate them later. But I agree this is probably better.

PR #477, opened 2025-03-03
FFS-2494: Implement /summary and /submit pages

millerti:
## Changes
Previously, the income data was shown on the /summary page and included a legal consent checkbox.
Now, the consent checkbox has been moved to the new /submit page.
Additionally, the pdf generation code was moved under the /submit page to keep things Rails-y.

## Testing instructions
0. Go through the normal CBV flow. Now, the /summary page should look like the figma design in the ticket, and the new /submit page should also look like the design.
1. Be on the lookout for missing strings. I had a tough time with wrangling our string tokens and I'm worried I inadvertently deleted something. It would look like blank space where a string should go.
2. Download the PDFs, including the caseworker version, and look for missing strings.
3. The /submit page should drop the new ApplicantAccessedSubmitPage event. It's also added to our funnel in Mixpanel. Make sure you see it!

## Acceptance testing


app/app/views/cbv/submits/show.html.erb
```diff
@@ -0,0 +1,17 @@
+<% title = t(".page_header") %>
+<% content_for :title, title %>
+<div class="grid-col tablet:grid-col-10">
+<h1><%= title %></h1>
+</div>
+
+<div class="usa-prose">
+  <p><%= raw(t(".warning")) %></p>
```
tdooner, line 8: `raw(...)` is a security risk. It's preferable to use a [HTML-safe key](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations) so that we don't inadvertently change it later and introduce a risk of XSS.

So, if you make this `t(".warning_html")` instead, it should work by rendering the HTML properly without the need for `raw`.

app/app/views/cbv/submits/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+</div>
+
+<div class="usa-prose margin-y-5">
+  <h2><%= t(".legal_header") %></h2>
+  <%= form_with model: @cbv_flow, url: cbv_flow_submit_path, html: { class: "usa-form maxw-full" }, builder: UswdsFormBuilder, method: "patch" do |f| %>
+  <%= f.check_box :consent_to_authorized_use, { 'label': raw(t(".consent_checkbox_label", agency_acronym: current_agency.agency_short_name)) } %>
```
tdooner, line 14: Same note about `raw(...)`

app/app/controllers/cbv/summaries_controller.rb
```diff
@@ -221,32 +25,6 @@ def track_accessed_income_summary_event(cbv_flow, payments)
...
-    Zlib::GzipWriter.open(gzipped_path) do |gz|
-      input_tempfile.binmode
-      input_tempfile.rewind
-      gz.write(input_tempfile.read)
-    end
-
-    gzipped_tempfile.rewind
-    gzipped_tempfile
-  ensure
-    input_tempfile.close unless input_tempfile.closed?
```
tdooner, line 250: The next time we have to move this around, we really should just refactor it into its own module/service object.

app/app/helpers/cbv/submit_helper.rb
```diff
@@ -0,0 +1,2 @@
+module Cbv::SubmitHelper
+end
```
tdooner, line 2: This file is not necessary. It doesn't hurt anything either but I generally err on the side of removing unnecessary files so they don't create confusion.

app/spec/helpers/cbv/submit_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+# describe Cbv::SubmitHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe Cbv::SubmitHelper, type: :helper do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
tdooner, line 15: Also unnecessary

app/spec/requests/cbv/submit_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+RSpec.describe "Cbv::Submits", type: :request do
+  describe "GET /index" do
+    pending "add some examples (or delete) #{__FILE__}"
+  end
+end
```
tdooner, line 7: That's the problem with rails generators... they do tend to make some unnecessary stuff.

tdooner (commented):
Wow, big diff. Good work with moving all this stuff around. Looks good to me once you fix the `raw(t(...))` thing.

PR #483, opened 2025-03-04
FFS-2397-3: Experimental: fix Mixpanel location inference

millerti:
## Changes
This is an experimental PR to see if Mixpanel support's suggestion of sending "0" as the IP for our events will fix their location inference. It works locally, but the real test will be if it works in Demo. If this doesn't work in Demo, I will revert this PR.

## Acceptance testing


tdooner (approved):
Do you have a theory as to why "0" will work differently than `nil` (as it was before we tried to update this code)?

Personally if this works, a comment might be helpful so we don't accidentally remove this later (since it does look pretty nonintuitive).

millerti:
This is just what the Mixpanel support person told me to do, so I assume it's some weird "let's just override this specific value with meaning" programming choice. That is, if it works. I agree that it's odd and have added explanatory comments.

PR #487, opened 2025-03-05
No ticket: adding applicant id to user_events controller

millerti:
## Changes
I discovered that the user_events_controller never got the applicant ID added. This is causing an issue where events will not get the right profile for pinwheel and help modal events.
In addition, this fixes an issue where caseworkers would all get the same profile. Now, caseworkers will get a distinct id of "caseworker-" + their user ID.

## Acceptance testing


app/lib/mixpanel_event_tracker.rb
```diff
@@ -8,24 +8,29 @@ def initialize
...
+    end
 
-      @tracker.people.set(distinct_id, tracker_attrs)
+    # For caseworker events, use the "user_id" attribute as the distinct_id
+    # For client events, use the "cbv_applicant_id" attribute as the distinct_id as it currently best
+    # represents the concept of a unique user.
+    user_id = attributes.fetch(:user_id, "")
+    applicant_id = attributes.fetch(:cbv_applicant_id, "")
+    if user_id.present?
+      distinct_id = "caseworker-#{user_id}"
```
tdooner, line 26: Oh, this is a neat idea.

tdooner (approved):

PR #491, opened 2025-03-06
FFS-2340: Encapsulate Events Handling in a New Model

tdooner:
## Ticket

Resolves [FFS-2340](https://jiraent.cms.gov/browse/FFS-2340).

## Changes

**FFS-2340: Create `WebhookEvents` to track sync success/failure**
> We currently use Pinwheel-specific columns on `PayrollAccount` to track
> the status of the sync: income/paystubs/employment/identity
> synced_at/errored_at.

> To accommodate other payroll aggregators, and better record sync
> statuses, let's create a WebhookEvents model to record which webhooks
> we've received for a given PayrollAccount. Success or failure, then, is
> not determined by the columns, but by the presence of the necessary
> WebhookEvents for that payroll aggregator.

## Context for reviewers

See above.

## Acceptance testing


app/app/models/payroll_account.rb
```diff
@@ -11,9 +11,12 @@ def self.sti_class_for(type_name)
 
   belongs_to :cbv_flow
 
-  after_update_commit {
-    I18n.with_locale(cbv_flow.cbv_flow_invitation.language) do
-      broadcast_replace target: self, partial: "cbv/synchronizations/indicators", locals: { pinwheel_account: self }
```
tdooner, line 16: I moved this from the model -> webhooks controller, since it made a bit more sense to me to put it there.

app/app/controllers/webhooks/pinwheel/events_controller.rb
```diff
@@ -1,53 +1,36 @@
...
+  # using a same-length dummy key even if the `end_user_id` does not match a
+  # valid `cbv_flow`.
   DUMMY_API_KEY = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
 
   def create
-    unless @cbv_flow
-      Rails.logger.info "Unable to find CbvFlow for end_user_id: #{params["payload"]["end_user_id"]}"
-      return render json: { status: "ok" }
+    @payroll_account = @cbv_flow.payroll_accounts.find_or_create_by(type: :pinwheel, pinwheel_account_id: params["payload"]["account_id"]) do |new_payroll_account|
+      new_payroll_account.supported_jobs = get_supported_jobs(params["payload"]["platform_id"])
```
tdooner, line 13: This fixes a rare bug where the webhooks are received out-of-order. The previous code required the `account.added` webhook to be the first one we receive. I think we saw once or twice that that webhook was delayed, and therefore the other webhook had no PayrollAccount to be associated with.

GeorgeCodes19: I wonder if this is the bug that I encountered when pairing with @daphnegold. The synchronizations page would immediately fail for some reason unknown to me. Let's see if this fixes that issue.

app/app/models/payroll_account.rb
```diff
@@ -10,10 +10,14 @@ def self.sti_class_for(type_name)
...
 
   belongs_to :cbv_flow
+  has_many :webhook_events
 
-  after_update_commit {
-    I18n.with_locale(cbv_flow.cbv_flow_invitation.language) do
-      broadcast_replace target: self, partial: "cbv/synchronizations/indicators", locals: { pinwheel_account: self }
+  private
+
+  def webhook_event(event_name, event_outcome = nil)
```
GeorgeCodes19, line 17: I was a bit confused on the purpose of this method at first, but after reviewing some other logic in the code I can see that the purpose of this method is to check for the presence of a `WebhookEvent` matching the name and status/outcome. Maybe we can consider adding a verb to this method? i.e. `find_webhook_by_status`. It's a big verbose, but it makes following the flow a bit easier.

app/app/models/payroll_account/pinwheel.rb
```diff
@@ -1,57 +1,37 @@
...
-    (supported_jobs.exclude?("employment") || employment_synced_at.present?) &&
-    (supported_jobs.exclude?("income") || income_synced_at.present?) &&
-    (supported_jobs.exclude?("identity") || identity_synced_at.present?)
+    (supported_jobs.exclude?("paystubs") || webhook_event(JOBS_TO_WEBHOOK_EVENTS["paystubs"]).present?) &&
+      (supported_jobs.exclude?("employment") || webhook_event(JOBS_TO_WEBHOOK_EVENTS["employment"]).present?) &&
+      (supported_jobs.exclude?("income") || webhook_event(JOBS_TO_WEBHOOK_EVENTS["income"]).present?) &&
+      (supported_jobs.exclude?("identity") || webhook_event(JOBS_TO_WEBHOOK_EVENTS["identity"]).present?)
   end
 
   def job_succeeded?(job)
```
GeorgeCodes19, line 18: Nice, this looks a lot better than the previous iteration.

app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -10,18 +10,15 @@
     let(:account_id) { SecureRandom.uuid }
     let(:comment) { "This is a test comment" }
     let(:supported_jobs) { %w[income paystubs employment] }
-    let(:income_errored_at) { nil }
-    let(:paystubs_errored_at) { nil }
-    let(:employment_errored_at) { nil }
+    let(:errored_jobs) { [] }
```
GeorgeCodes19, line 13: Nice- I like this pattern more than explicitly digging up different `*_errored_at` vars

app/spec/factories/payroll_account.rb
```diff
@@ -2,14 +2,28 @@
...
-    income_synced_at { DateTime.now }
-    identity_synced_at { DateTime.now }
     supported_jobs { %w[income paystubs employment identity] }
 
-    trait :with_paystubs_errored do
-      paystubs_errored_at { DateTime.now }
+    transient do
+      # Use this to create a test case where a job has failed to sync for a
+      # PayrollAccount.
+      with_errored_jobs { %w[] }
```
GeorgeCodes19, line 10: For some reason this is my favorite part about the PR. I appreciate a nice test suite.

app/spec/models/payroll_account/pinwheel_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        create_webhook_events(income_success: false, income_errored: true)
+      end
+
+      it "returns false when income is supported but it errored out" do
+        expect(payroll_account.job_succeeded?('income')).to be_falsey
+      end
+    end
+
+    context "when job is unsupported" do
+      let(:supported_jobs) { super() - [ "income" ] }
```
GeorgeCodes19, line 89: What's going on here with `super()`?

GeorgeCodes19 (approved):
Good works. Seems like a lot. I really dig the updates to the test suite.

app/app/controllers/webhooks/pinwheel/events_controller.rb
tdooner: Was just pairing with her - hers was a setup issue (we have a limit of 10 pinwheel webhook subscriptions, and periodically need to garbage collect old engineers out of there). So her sync was failing because pinwheel never registered the webhooks.

app/spec/models/payroll_account/pinwheel_spec.rb
tdooner: Good question. It refers to the `supported` jobs in the enclosing context/describe. Here's an example:

```ruby
describe "how super() works" do
  let(:arguments) { [1, 2] }

  it { expect(arguments).to eq([1, 2]) }    # passes

  context "when there are three arguments" do
    let(:arguments) { super() + [3] }

    it { expect(arguments).to eq([1, 2, 3]) }     # passes
  end
end

tdooner: I try not to over-use it, because lord knows Rspec tests are already hard enough to understand where the variables are coming from even without this. But I did like how readable it is to do `super() - [ "income" ]` to denote the fact that the "income" job is not supported.

app/app/models/payroll_account.rb
tdooner: Yeah that's a good idea. I was waffling on it - brevity vs being explicit. I almost always lean towards "being explicit" and I think you're right that I should have here as well. I'll go with `find_webhook_event`.

tdooner:
As this is a refactor with no user-facing changes to behavior, going to skip acceptance testing. We will be testing this all holistically in Demo as we develop Multi-Aggregator.

PR #495, opened 2025-03-10
Display Pinwheel initialization error alert during employer search

daphnegold:
## Ticket

n/a

## Changes
- Store Pinwheel initialization error during app config
- In dev only, display alert with initialization error during employer search
- Add `arm64-darwin-24` to Gemfile.lock platforms

## Context for reviewers
We have a limit of 10 webhooks with Pinwheel before it will error out and cause issues during employer search. There is a line in the logs, but this PR makes it more obvious to developers that there's an issue from the UI. 

My computer adds `arm64-darwin-24` during `bundle install`.

## Acceptance testing

Screenshot 2025-03-07 at 3 55 56 PM

### DIFF
--- a/app/Gemfile.lock
+++ b/app/Gemfile.lock
@@ -512,6 +512,7 @@ GEM
 PLATFORMS
   aarch64-linux
   arm64-darwin-23
+  arm64-darwin-24
   x86_64-darwin-20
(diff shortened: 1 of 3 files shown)


tdooner (approved):
Daphne and I paired on this!

PR #499, opened 2025-03-11
FFS-2567: Fix datepicker on caseworker invitation page (after validation errors)

daphnegold:
## Ticket

Resolves [FFS-2567](https://jiraent.cms.gov/browse/FFS-2567).

## Acceptance testing

### DIFF
--- a/app/app/javascript/application.js
+++ b/app/app/javascript/application.js
@@ -8,6 +8,14 @@ import "@uswds/uswds"
 import components from "@uswds/uswds/src/js/components"
 let initialLoad = true
 
+const reinitializeUSWDS = () => {
+  const target = document.body
+  Object.keys(components).forEach((key) => {
+    const behavior = components[key]
+    behavior.on(target)
+  })
+}
+
 document.addEventListener("turbo:load", () => {
   if (initialLoad) {
     // initial domready is handled by `import "uswds"` code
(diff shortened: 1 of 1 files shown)


app/app/javascript/application.js
```diff
@@ -8,18 +8,26 @@ import "@uswds/uswds"
 import components from "@uswds/uswds/src/js/components"
 let initialLoad = true
 
+const reinitializeUSWDS = () => {
+  const target = document.body
+  Object.keys(components).forEach((key) => {
+    const behavior = components[key]
```
GeorgeCodes19, line 14: I have no idea how this is working behind the scenes. Can you please add some comments?

app/app/javascript/application.js
```diff
@@ -8,18 +8,26 @@ import "@uswds/uswds"
 import components from "@uswds/uswds/src/js/components"
 let initialLoad = true
 
+const reinitializeUSWDS = () => {
+  const target = document.body
```
GeorgeCodes19, line 12: is the `target` var needed? can we simply use `behavior.on(document.body)`?

GeorgeCodes19 (approved):
I'm glad this works! I suppose that I should spend some time getting familiar with the USWDS source code. A few minor things, but nothing blocking.

app/app/javascript/application.js
daphnegold: It doesn't need to be, but I don't think it matters either way here greatly. The variable is immediately garbage collected and only has functional scope.

PR #502, opened 2025-03-12
FFS-2462 - client_agency_id derived from the current user

GeorgeCodes19:
## Ticket

Resolves [FFS-2462](https://jiraent.cms.gov/browse/FFS-2462).

Merge after FFS2489

## Changes

- [x] Update the `Api::InvitationsController` to derive the `client_agency_id` from the user-supplied authentication token rather than the request.
- [x] Create a test to validate that a user supplied `client_agency_id` does not create an invitation link for an unrecognized or non-permitted `client_agency_id`

## Context for reviewers
For security purposes we should not allow the user to specify the `client_agency_id` in the request as this would enable them to create invitations for agencies outside of their own. The `client_agency_id` shall be derived from the auth bearer token instead.

## Acceptance testing

- [X] No acceptance testing needed


app/spec/controllers/api/invitations_controller_spec.rb
```diff
@@ -25,24 +28,40 @@
         .and change(CbvApplicant, :count).by(1)
 
       expect(response).to have_http_status(:created)
-      expect(JSON.parse(response.body).keys).to include("url")
+      expect(JSON.parse(response.body).keys).to include("tokenized_url")
+    end
+
+    it "creates an invitation with the client_agency_id from the api token" do
```
GeorgeCodes19, line 34: I'm not sure if this is adding any real value. The malicious user could attempt to supply the `client_agency_id` as an arbitrary property anywhere in the request payload. This test specifically tests against the previous API implementation which is not yet public or known.

Maybe a better test would be to assert that the resulting `CbvFlowInvitation` and `CbvApplicant` have a `client_agency_id` that matches that the one associated with an authentication token.

app/app/controllers/api/invitations_controller.rb
```diff
@@ -22,27 +22,29 @@ def create
...
         :middle_name,
         :last_name,
-        :client_agency_id,
         :client_id_number,
         :case_number,
         :snap_application_date,
         :agency_id_number,
-        :beacon_id
+        :beacon_id,
+        :client_agency_id
```
daphnegold, line 45: It's a bit confusing for me that we are removing `client_agency_id` from the params via `delete`, but we are still keeping it in the `params.permit` list.

daphnegold (approved):
This lgtm, and a big security improvement. I made a comment on logic that we could maybe discuss. Approving anyway if this functions as intended. Nice job!

app/spec/controllers/api/invitations_controller_spec.rb
daphnegold: I agree there might be a better test here to make sure security is locked down, but good to have a unit test on the functionality

app/app/controllers/api/invitations_controller.rb
GeorgeCodes19: So the `params.delete(:client_agency_id)` isn't really needed. We can simply do without `:client_agency_id` in the `.permit` call. Deleting and omitting `client_agency_id` essentially do the same thing here- that is, they're not allowing the `client_agency_id` to be permitted in the request payload.

Whenever we call `.merge` we're creating a new `ActionController::Parameters` instance with the `client_agency_id` being derived from the current user rather than the request.

https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-merge

I think this can be written better. I've been augmenting the existing code, but I'll consider rewriting this method entirely since the logic is hard to follow.

daphnegold: I like that refactor, but I defer to you whether it's worth it or if we are okay with it as is, since it works. I do think doing this change would make the code a lot clearer and be less brittle.

PR #509, opened 2025-03-14
FFS-2528: Make CbvApplicants nullible

millerti:
## Changes
We can now make a CbvApplicant without first, last, and snap app date. 

## Context for reviewers
We need to be able to do this before we have any info about the user in the generic link flow. Now, we'll make this object right when the user clicks the generic link, and we will fill out the information in this object as we receive it (in a future ticket). CWs are still prevented from creating an invite for an applicant without a first, last, or snap date. But later we will programmatically create such applications.

## Acceptance testing


app/app/models/response_objects/paystub.rb
```diff
@@ -33,7 +33,6 @@ def self.from_pinwheel(response_body)
       )
     end
 
-
```
tdooner, line 36: Thank you. These were due to me merging the rubocop rule to enforce this at the same time a different PR was open that committed these violations of the rule.

PR #510, opened 2025-03-14
FFS-2540: Generic link

millerti:
## Changes
1. Can now complete CBV flow through the generic link by accessing /cbv/links/sandbox
2. You cannot yet provide any applicant details
3. This also ensures that the snap_application_date is by default today and gets the previous 90 days. [FFS-2558](https://jiraent.cms.gov/browse/FFS-2558)

## Testing instructions
1. Regression test the normal caseworker invitation flow. The default application date will be today. 
2. Make sure modifying the application date also works.
3. Make sure the invitation form validation still works.
4. Verify that the new flow starting at /cbv/links/sandbox works and results in data from the previous 90 days
5. Note that /cbv/links/ma and /cbv/links/nyc will not work because those legacy configurations still have requirements that I did not fix. When we enable a new pilot, we will need to make sure that the generic link flow sets any unique defaults they need at that time.

## Acceptance testing


app/app/models/cbv_applicant.rb
```diff
@@ -46,6 +43,10 @@ def paystubs_query_begins_at
     PAYSTUB_REPORT_RANGE.before(snap_application_date)
   end
 
+  def set_default_snap_application_date
+    self.snap_application_date ||= Time.current - 90.days
```
jeffcatania-usds, line 47: Question for my own understanding: what timezone does our server run in? If an applicant in California accesses the generic link at 11PM PT, would they only get 89 days of data?

app/config/routes.rb
```diff
@@ -26,6 +26,10 @@
     get "/maintenance", to: "maintenance#show"
     # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
 
+    scope "/:client_agency_id", module: :cbv, constraints: { client_agency_id: Regexp.union(Rails.application.config.client_agencies.client_agency_ids) } do
```
jeffcatania-usds, line 29: "Regular Expressions: now you have two problems".  It doesn't look like we use Regexp in our other routes. Is there a more explicit way that we could handle?

jeffcatania-usds (commented):
Took a pass at the explicit code here.  I don't have full system visibility to know exactly where things could break.  I'd recommend moving away from request.path checking in the base_controller and regular expressions in the routes in favor of more explicit routing.

app/app/models/cbv_applicant.rb
millerti: Hmm, not sure. I wonder if Tom has insight here. I really wanted to be able to refer to created_at here instead of Time.current, but since I'm creating the row and because I think that gets set in SQL, I don't think I'll have access to it. I wonder if there's a way to! But yeah, in reference to your question, I bet if the machine running SQL is in a different TZ than here, it could result in an unexpected range.

app/config/routes.rb
millerti: I actually didn't add this regexp! I'd be happy to consider refactoring somehow but I just copied it from further down in the file.

tdooner: Yeah, this is weird. I think it's acceptable, given the pattern below, but really I think we don't need the `constraint` if we validate the value of `client_agency_id` in the controller? Still, a bit unfortunate. Can we think of a good URL for this page that doesn't have it underneath this agency prefix?

Currently it's: **/ma/new**
What about?
* **/cbv/entries/ma**
* (or if we want it to be a different controller) **/cbv/links/ma** or something?

app/app/controllers/cbv/generic_links_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Cbv::GenericLinksController < Cbv::BaseController
+  skip_before_action :set_cbv_flow, :capture_page_view
+
+  def show
+    @cbv_flow = CbvFlow.create_without_invitation(params[:client_agency_id])
```
tdooner, line 5: If we're doing the `:client_agency_id` regex thing in routes, let's add a sanity check before trying to create the `@cbv_flow` to ensure it's a valid agency ID. (As the Regex as currently written will match any string containing "ma", "nyc", or "sandbox")

image

Something like:

```ruby
before_action :ensure_valid_client_agency_id

# ...

def ensure_valid_client_agency_id
  return if agency_config.client_agency_ids.include?(client_agency_id)

  redirect_to root_url, flash: { info: "Invalid link used to begin CBV. Try clicking the link again or contact your local SNAP agency for help." }
end
```

millerti: Fixed both of these. Now, the link will be /cbv/links/sandbox. I still think we need the agency id in the url.

app/config/routes.rb
millerti: Settled on /cbv/links/agency

app/app/views/caseworker/cbv_flow_invitations/_nyc.html.erb
```diff
@@ -1,4 +1,4 @@
-<% cbv_applicant = @cbv_flow_invitation.cbv_applicant || @cbv_flow_invitation.build_cbv_applicant %>
+<% cbv_applicant = @cbv_flow_invitation.cbv_applicant || @cbv_flow_invitation.build_cbv_applicant() %>
```
jeffcatania-usds, line 1: Is the requirement to only update the NYC flow and not MA?

app/config/locales/en.yml
```diff
@@ -242,6 +242,7 @@ en:
         step3: Submit your payment information.
         step3_description: We’ll automatically send it to %{agency_acronym}. This will help you prove your income and can help %{agency_acronym} get to a decision faster.
         subheader_html: 'We’ll send your payment records from your employer(s) to %{agency_acronym} to verify your income. <strong>Most people complete this process in under 5 minutes.</strong> Just follow these steps:'
+    error_invalid_link: Invalid link used to begin CBV. Try clicking the link again or contact your local SNAP agency for help.
```
jeffcatania-usds, line 245: Why do we assume SNAP in this text? It doesn't look like this error message is program specific.

jeffcatania-usds (approved):
Approved pending my comments. My big question is why this PR is so specific to NYC and SNAP.

app/app/controllers/cbv/entries_controller.rb
```diff
@@ -1,5 +1,7 @@
 class Cbv::EntriesController < Cbv::BaseController
   def show
+    @current_agency = current_agency
```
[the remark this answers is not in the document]
millerti: My bad -- this line isn't even necessary anymore.

app/app/views/caseworker/cbv_flow_invitations/_nyc.html.erb
millerti: No, and I think this is just a remnant of an earlier change that I discarded.

app/config/locales/en.yml
millerti: Good catch. Do you think "contact your caseworker for help" is appropriate?

PR #511, opened 2025-03-14
FFS-2585 Fix Expired Invitations crash

jeffcatania-usds:
## Ticket

Resolves [FFS-2585](https://jiraent.cms.gov/browse/FFS-2585).

## Background
This bug likely affects production, and the user would get the wrong error page.

image

When I attempt to access a CBV flow at an expired token, I see this crash page. http://localhost:3000/en/cbv/entry?token={token} 

The CBV Entry page was calling :set_cbv_flow (via BaseController's before_action methods) and redirecting to the expired_invitations_controller.  The problem is that expired_invitations_controller was then calling capture_page_view but it doesn't have a cbv_flow to log there and it crashes.

## Changes
* Skip :capture_page_view in the expired_invitations_controller to prevent crash

## Acceptance testing


millerti (approved):
Looks good. I think this is the right solution because we still log another event in this case. Great find!

PR #517, opened 2025-03-20
FFS-2540-2: /applicant_information page

millerti:
## Changes
Added configurable fields to the /applicant_information page. This data also now shows up on the /summary page. Note that this info does NOT yet show up on the PDF reports -- it will be added later.

## Testing instructions
1. Verify that the generic link workflow starting at /cbv/links/sandbox works as you'd expect. There are 3 required fields (first, last, case number) and one optional field (middle name) on the sandbox's /applicant_information page.
1. Verify that the /applicant_information page does not show up when all required fields have data in them
1. Verify that it DOES show up if you click the Edit button on the /summary page, even if all required fields are fulfilled
1. Verify that the invitation work flow does NOT show the /applicant_information page and the /summary page never shows the applicant information table
1. Satisfy yourself that the strategy I've implemented for adding new customizations of the /applicant_information page is sane.
1. Verify that the events described in the ticket are added, being logged, and documented on our [events page](https://confluenceent.cms.gov/display/SFIV/CBV+Events) (ApplicantAccessedInformationPage, ApplicantEncounteredInformationPageError, ApplicantClickedEditInformationLink)

## Instructions for customizing the /applicant_information page
When adding a new state customization, you must:
1. Add info for that customization in client_agency_config.yml under required_applicant_attrs. Format is field_name: [true if required, false if optional]. field_name must match the field name on the form and in the translation tokens.
3. Add form layout in /applicant_informations view
4. Add string translations for each field under cbv.applicant_informations.[client_agency_id].fields.[field_name]. You must include both the .blank and .prompt strings, where blank is an error message for being blank and prompt is the field title.

## Acceptance testing


app/app/views/cbv/summaries/show.html.erb
```diff
@@ -53,4 +53,45 @@
   <%= get_comment_by_account_id(account_id).dig("comment") != "" ? get_comment_by_account_id(account_id).dig("comment") : t("shared.not_applicable") %>
   <% end %>
 
+<% if !@cbv_flow.cbv_flow_invitation.present? %>
```
tdooner, line 56: The opposite of `present?` is `blank?`, so you could do:

```suggestion
<% if @cbv_flow.cbv_flow_invitation.blank? %>
```

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -53,4 +53,45 @@
...
+            <svg class="usa-icon usa-icon--size-1" aria-hidden="true" role="img">
+              <use
+                xlink:href="<%= asset_path("@uswds/uswds/dist/img/sprite.svg#edit") %>"
+              ></use>
+            </svg>
+            Edit
+          </a>
+        </h4>
+      </th>
+          </tr>
```
tdooner, line 82: Weird indenting on the `<tr>`/`</tr>`

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -53,4 +53,45 @@
...
+  </thead>
+  <tbody>
+    <% Rails.application.config.client_agencies[@cbv_flow.client_agency_id].required_applicant_attrs.each do |item| %>
```
tdooner, line 85:
```suggestion
    <% current_agency.applicant_attributes.each do |item| %>
```

app/config/locales/es.yml
```diff
@@ -166,16 +170,10 @@ es:
...
-        consent_checkbox_label_html: Check this box to confirm that the information provided by you is true and complete to the best of your knowledge. You agree to inform %{agency_acronym} of any income not reflected in this report or any discrepancies found in the information gathered with this tool. You understand that providing accurate and complete information is your responsibility, and any false or omitted information may have legal consequences. For more information on your rights and responsibilities regarding private information you share with %{agency_acronym} please refer to the Rights and Responsibilities you signed at application, found on the %{agency_acronym} website.<br><br>By sending this report, you authorize its use for income verification by %{agency_acronym} and authorized personnel.
-        legal_header: Legal agreement
-        none_found: We didn't find any payments from this employer in the past 90 days.
-        none_found_confirmed: We've confirmed that there are no payments from this employer in the past 90 days. This happens when the client hasn't received income from this job during that time.
-        page_header: Submit your income report
```
tdooner, line 178: There are a lot more changes in the locale files than I was expecting to see... For example, why were these (and so many other strings) removed? Aren't they still being used in the submits controller?

(This also applies to the change made to `config/i18n-tasks.yml` - I'm not sure why that was necessary here?)

app/spec/controllers/cbv/applicant_informations_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
+require "rails_helper"
+
+RSpec.describe Cbv::ApplicantInformationsController, type: :controller do
+  describe "#show" do
+    let(:current_time) { Time.now }
+    let(:cbv_flow) do
+      create(:cbv_flow,
+        created_at: current_time
```
tdooner, line 8: This isn't necessary, is you don't seem to change the value of `current_time` in the tests, and it will default to `Time.now` anyway.

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -53,4 +53,45 @@
...
+      <th class="padding-3" style="background-color: #d9e8f6; text-align: right;">
+        <h4 class="margin-0">
+          <a class="usa-link" href="<%= cbv_flow_applicant_information_path(force_show: true) %>">
+            <svg class="usa-icon usa-icon--size-1" aria-hidden="true" role="img">
+              <use
+                xlink:href="<%= asset_path("@uswds/uswds/dist/img/sprite.svg#edit") %>"
+              ></use>
+            </svg>
+            Edit
+          </a>
```
tdooner, line 79: Generally better to use `link_to` to let Rails make the link so we can use Rails' options if we want to make the link have special additional behavior, later. We can also use `image_tag` to output the SVG tag:

```suggestion
          <%= link_to cbv_flow_applicant_information_path, class: "usa-link" do %>
            <%= image_tag "@uswds/uswds/dist/img/usa-icons/edit.svg", role: "img", alt: "" %>
            Edit
          <% end %>
```

app/app/controllers/cbv/base_controller.rb
```diff
@@ -134,4 +149,14 @@ def track_invitation_clicked_event(invitation, cbv_flow)
...
+
+  def load_applicant_attrs
+    if !@cbv_flow.present?
+      @applicant_attrs = []
+      @required_applicant_attrs = []
+    else
+      @applicant_attrs = Rails.application.config.client_agencies[@cbv_flow.client_agency_id].required_applicant_attrs.compact.keys.map(&:to_sym)
+      @required_applicant_attrs = Rails.application.config.client_agencies[@cbv_flow.client_agency_id].required_applicant_attrs.select { |_, required| required }.keys.map(&:to_sym)
+    end
+  end
```
tdooner, line 161: These variables are akin to global variables because they will be set in every Cbv:: controller (as they are all subclasses of `Cbv::BaseController`). We should be judicious about how many variables we set in the base controller to reduce confusion and poorly factored code. As it seems like these variables are only used in the Cbv::ApplicantInformationsController, let's move this method there.

Also, we have a helper method to get the current agency config. It's `current_agency`.

Its name also could be better. We use `set_*` to indicate setting controller instance variables. And I don't love the abbreviation. How about `set_applicant_attributes`?

app/spec/controllers/cbv/applicant_informations_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      session[:cbv_flow_id] = cbv_flow.id
+    end
+
+    context "when rendering views" do
+      render_views
+
+      it "renders the sandbox fields" do
+        get :show
+
+        puts "response.status: #{response.status}"
```
tdooner, line 22:
```suggestion
```

app/spec/controllers/cbv/applicant_informations_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+              case_number: "" # required
+            }
+          }
+        }
+
+        expect(response).to have_http_status(:unprocessable_entity)
+        expect(response.body).not_to include(I18n.t(".cbv.applicant_informations.sandbox.fields.first_name.blank"))
+        expect(response.body).not_to include(I18n.t(".cbv.applicant_informations.sandbox.fields.middle_name.blank"))
+        expect(response.body).to include(I18n.t(".cbv.applicant_informations.sandbox.fields.last_name.blank"))
+        expect(response.body).to include(I18n.t(".cbv.applicant_informations.sandbox.fields.case_number.blank"))
```
tdooner, line 46: The beginning dot isn't doing anything in here. This is because a dot (relative I18n key) is [a nice feature](https://guides.rubyonrails.org/i18n.html#lazy-lookup) of the view helper method `translate` (aka `t`). It knows which view or controller you're in and prepends it to the key before calling `I18n.t`. But since we're here in tests, calling `I18n.t` directly, whatever raw key we pass in will be used directly.

```suggestion
        expect(response.body).not_to include(I18n.t("cbv.applicant_informations.sandbox.fields.first_name.blank"))
        expect(response.body).not_to include(I18n.t("cbv.applicant_informations.sandbox.fields.middle_name.blank"))
        expect(response.body).to include(I18n.t("cbv.applicant_informations.sandbox.fields.last_name.blank"))
        expect(response.body).to include(I18n.t("cbv.applicant_informations.sandbox.fields.case_number.blank"))
```

tdooner (commented):
Overall looks good, but would like to see a bit of code reorganization for code quality reasons:
- [x] Field validation logic into `CbvApplicant` model
- [ ] Redirection logic moved from `next_add_jobs_path` into `next_path` in AddJobsController
- [ ] `load_applicant_attrs` moved into its relevant controller
- [ ] Renaming these fields to just `applicant_attributes`
- [ ] A good understanding of why the I18n changes were so expansive/possibly unrelated

app/app/controllers/caseworker/cbv_flow_invitations_controller.rb
```diff
@@ -29,7 +29,7 @@ def create
       @cbv_flow_invitation.errors.delete(:cbv_applicant)
 
       error_count = @cbv_flow_invitation.errors.size
-      error_header = "#{helpers.pluralize(error_count, 'error')} occurred"
+      error_header = "#{helpers.pluralize(error_count, 'error')} occurred" # TODO: is this properly tokenized?
```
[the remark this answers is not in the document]
millerti: Fixed in the applicant_information file, which should be translated. Thanks!

app/app/views/cbv/applicant_informations/_sandbox.html.erb
```diff
@@ -0,0 +1,7 @@
+<% cbv_applicant = @cbv_flow.cbv_applicant %>
+<%= f.fields_for :cbv_applicant, cbv_applicant do |f2| %>
+  <%= f2.text_field :first_name, label: t(".fields.first_name.prompt") %>
+  <%= f2.text_field :middle_name, label: t(".fields.middle_name.prompt") %>
+  <%= f2.text_field :last_name, label: t(".fields.last_name.prompt") %>
+  <%= f2.text_field :case_number, label: t(".fields.case_number.prompt") %>
```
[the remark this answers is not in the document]
millerti: [Created!](https://jiraent.cms.gov/browse/FFS-2641)

app/config/locales/es.yml
millerti: I'm removing these English strings in the es file and overriding the linter warning in i18n-tasks because it's our process: "Implement English strings only. In cases where you modify an existing English string, **delete the corresponding Spanish string**."

This makes it easy to find missing strings in translation tickets (we can just run i18n-tasks missing).

app/lib/client_agency_config.rb
```diff
@@ -57,6 +58,7 @@ def initialize(yaml)
       @staff_portal_enabled = yaml["staff_portal_enabled"]
       @sso = yaml["sso"]
       @weekly_report = yaml["weekly_report"]
+      @required_applicant_attrs = yaml["required_applicant_attrs"] || {}
```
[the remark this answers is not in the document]
millerti: I am going to lean on the side of YAGNI on this right now because this ticket is so late.

app/app/views/cbv/summaries/show.html.erb
millerti: Unfortunately one of our linters is insisting on this indentation.

tdooner (approved):
Whew, lgtm. One little bit of weirdness with the error flash message on the "add jobs" page - its i18n key and how it's being rendered. Other than that, :shipit: .

app/spec/mailers/weekly_report_mailer_spec.rb
```diff
@@ -43,6 +43,10 @@
     travel_to(now)
   end
 
+  after do
+    travel_back
+  end
```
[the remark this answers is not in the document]
millerti: I did NOT see this fix the weird issues I was seeing with the weekly report mailer spec, but I agree that this was missing and is appropriate to include.

app/app/models/cbv_applicant.rb
```diff
@@ -1,5 +1,8 @@
 class CbvApplicant < ApplicationRecord
   after_initialize :set_snap_application_date, if: :new_record?
+  after_initialize :set_applicant_attributes
+  attr_reader :applicant_attributes, :required_applicant_attributes
```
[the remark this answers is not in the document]
millerti: Both are used outside of the class in the controller. We use the non-required version to allow params, then we use the required version to check for missing info.

app/config/i18n-tasks.yml
```diff
@@ -143,7 +143,7 @@ ignore_missing:
     - "pages.{error_404,error_500}.*"
     - "cbv.error_invalid_link" # FFS-2540
     - "cbv.applicant_informations.*" # FFS-2540
-    - "cbv.add_jobs.next_path.notice_no_answer" # FFS-2540
+    - "cbv.add_jobs.create.notice_no_answer" # FFS-2540
```
tdooner, line 146: Is this necessary now that the key is `cbv.add_jobs.next_path.notice_no_answer`?

tdooner (approved):

PR #523, opened 2025-03-28
FFS-2615: Form validation errors not applied to individual fields in CBV flow invitation form

daphnegold:
## Ticket

Resolves [FFS-2615](https://jiraent.cms.gov/browse/FFS-2615).

## Changes
Validation on the First Name and Last Name fields is back!

## Context for reviewers
Fixed form validation errors for name fields. Used the same pattern as the radio buttons with `<input>`. 

Considered adding a helper to our USWDS UswdsFormBuilder that could handle nested errors, but it got overly complicated and specific to solve this problem. This solution was more work in the view, but followed current conventions and was less of a 'surprise' to just keep the logic with our forms.

This regressed when we needed to display errors for nested fields, which broke the standard error display pattern. For nested error handling in our forms, we need to manually implement the error display.

## Acceptance testing

  ### Behold, validation errors
Screenshot 2025-03-28 at 3 24 58 PM

### DIFF
--- a/app/app/views/caseworker/cbv_flow_invitations/_ma.html.erb
+++ b/app/app/views/caseworker/cbv_flow_invitations/_ma.html.erb
@@ -1,10 +1,32 @@
 <% cbv_applicant = @cbv_flow_invitation.cbv_applicant || @cbv_flow_invitation.build_cbv_applicant(snap_application_date: Date.today) %>
 <%= f.fields_for :cbv_applicant, cbv_applicant do |f2| %>
-  <%= f2.text_field :first_name, label: t(".invite.first_name") %>
+  <div class="usa-form-group <%= "usa-form-group--error" if @cbv_flow_invitation.errors[:'cbv_applicant.first_name'].present? %>">
+    <legend class="usa-legend"><%= t(".invite.first_name") %></legend>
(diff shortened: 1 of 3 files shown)


GeorgeCodes19 (approved):
Overall this seems good. I like that it's explicit and declarative down to the specific field, but there seems to be a lot of code duplication. 

Might be a good area for us to practice some DRY and/or perhaps use our component system? That may involve creating a new directory and file e.g. `app/components/forms/input.html.erb`

GeorgeCodes19 (approved):
Overall this seems good. I like that it's explicit and declarative down to the specific field, but there seems to be a lot of code duplication. 

Might be a good area for us to practice some DRY and/or perhaps use our component system? That may involve creating a new directory and file e.g. `app/components/forms/input.html.erb`

daphnegold:
> Might be a good area for us to practice some DRY and/or perhaps use our component system? That may involve creating a new directory and file e.g. `app/components/forms/input.html.erb`

I think since it's just the two fields here and two of the views (`_nyc and _ma) aren't even used and being removed, it's okay to have it like this with a bit of duplication for now and not overoptimize. If we end up reusing bits more places, let's make a component for sure. Perhaps separate out each part of the view into reusable components. Great suggestion, thanks!

PR #524, opened 2025-03-29
Add client agency config and invitation sending for AZ

tdooner:
## Ticket

Resolves [FFS-2632](https://jiraent.cms.gov/browse/FFS-2632).

## Changes

**Add client agency config for AZ**

**Temporarily add OAuth configuration for AZ DES caseworker portal**
> Although we don't intend for caseworkers to log into AZ DES in the pilot
> itself, allowing ourselves to log into the AZ caseworker portal will
> facilitate our testing until we build some easy way for non-Engineers to
> test with the Tokenization API.

**Create CbvApplicant::AzDes to allow invitation creation**
> This model allows for the custom validation logic necessary to create an
> invitation for AZ DES. These CbvApplicants are a bit different than the
> others, since they don't validate first_name, last_name, or
> snap_application_date.

> One other tweaks was necessary to update the `sti_class_for`
> implementation to properly convert "az_des" to "AzDes" (previously it
> converted it only to "Az_des").

**Remove underscores from confirmation numbers**
> The AZ DES agency name has an underscore in it, but we shouldn't have
> that underscore in the generated confirmation number.

> Before - AZ_DES0001234
> After - AZDES0001234

**Fix `rails users:promote_to_service_account`**
> Not sure how this worked before. I also provided example invocations
> as comments because passing in arguments to rake tasks is a dark art.

## Acceptance testing

### DIFF
--- a/app/.env
+++ b/app/.env
@@ -3,6 +3,7 @@
 # ##############################################################################
 NYC_PINWHEEL_ENVIRONMENT=sandbox
 MA_PINWHEEL_ENVIRONMENT=sandbox
+AZ_DES_PINWHEEL_ENVIRONMENT=sandbox
 SANDBOX_PINWHEEL_ENVIRONMENT=sandbox
 MAINTENANCE_MODE=false
 MA_WEEKLY_REPORT_RECIPIENTS=test@email.com
--- a/app/app/controllers/cbv/submits_controller.rb
+++ b/app/app/controllers/cbv/submits_controller.rb
@@ -210,7 +210,7 @@ class Cbv::SubmitsController < Cbv::BaseController
 
   def generate_confirmation_code(prefix = nil)
     [
-      prefix,
+      prefix.gsub("_", ""),
       (Time.now.to_i % 36 ** 3).to_s(36).tr("OISB", "0158").rjust(3, "0"),
       @cbv_flow.id.to_s.rjust(4, "0")
     ].compact.join.upcase
--- a/app/app/models/cbv_applicant.rb
+++ b/app/app/models/cbv_applicant.rb
@@ -16,8 +16,8 @@ class CbvApplicant < ApplicationRecord
   end
 
   def self.sti_class_for(type_name)
-    # "ma" => CbvApplicant::Ma
-    CbvApplicant.const_get(type_name.capitalize)
+    # "az_des" => CbvApplicant::AzDes
+    CbvApplicant.const_get(type_name.camelize)
   end
 
   PAYSTUB_REPORT_RANGE = 90.days
--- a/app/app/views/caseworker/cbv_flow_invitations/_az_des.html.erb
+++ b/app/app/views/caseworker/cbv_flow_invitations/_az_des.html.erb
@@ -1,0 +1,29 @@
+<% cbv_applicant = @cbv_flow_invitation.cbv_applicant ||
+     @cbv_flow_invitation.build_cbv_applicant(client_agency_id: @cbv_flow_invitation.client_agency_id) %>
+<%= f.fields_for :cbv_applicant, cbv_applicant do |f2| %>
+  <%= f2.text_field :case_number, label: t(".invite.case_number") %>
+
+  <%= f.email_field :email_address, label: t(".invite.email_address") %>
+
+  <div class="usa-form-group <%= "usa-form-group--error" if f.field_error(:language).present? %>">
+    <legend class="usa-legend"><%= t(".invite.language_label") %></legend>
+    <%= f.field_error :language %>
+    <div class="usa-radio-group">
+      <% language_options.each_with_index do |(value, label), index| %>
+        <div class="usa-radio usa-radio--tile">
+          <input
+            class="usa-radio__input usa-radio__input--tile"
+            id="language-<%= value %>-<%= index %>"
+            type="radio"
+            name="<%= f.object_name %>[language]"
+            value="<%= value %>"
+            <%= 'checked="checked"' if f.object.language == value %>
+          >
+          <label class="usa-radio__label" for="language-<%= value %>-<%= index %>">
+            <%= label %>
+          </label>
+        </div>
+      <% end %>
+    </div>
+  </div>
+<% end %>
--- a/app/app/views/caseworker/entries/_az_des_signin_button.html.erb
+++ b/app/app/views/caseworker/entries/_az_des_signin_button.html.erb
@@ -1,0 +1,1 @@
+<%= button_to agency_translation("caseworker.entries.index.continue_to_login"), user_az_des_omniauth_authorize_path, data: { turbo: false }, class: "usa-button" %>
--- a/app/config/client-agency-config.yml
+++ b/app/config/client-agency-config.yml
@@ -23,6 +23,7 @@
   invitation_valid_days: 14
   weekly_report:
     recipient: <%= ENV['NYC_HRA_EMAIL'] %>
+
 - id: ma
   agency_name: Department of Transitional Assistance
   agency_short_name: DTA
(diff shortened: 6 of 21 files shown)


app/app/controllers/cbv/submits_controller.rb
```diff
@@ -210,7 +210,7 @@ def track_transmitted_event(cbv_flow, payments)
 
   def generate_confirmation_code(prefix = nil)
     [
-      prefix,
+      prefix.gsub("_", ""),
```
jeffcatania-usds, line 213: Can I understand this requirement more? Was the removal of underscores a request from AZ or are we implementing this so that our system does not break?  What happens in the future if we have an agency client that has underscores?

app/app/models/cbv_applicant.rb
```diff
@@ -16,8 +16,8 @@ def self.sti_name
   end
 
   def self.sti_class_for(type_name)
-    # "ma" => CbvApplicant::Ma
-    CbvApplicant.const_get(type_name.capitalize)
+    # "az_des" => CbvApplicant::AzDes
```
jeffcatania-usds, line 19: My initial reaction is that we need to have a more generic solution for this because we hardcoded for the MA use case before and now we hard code for the AZ use case. But it looks like the generic solution would be camelcase here. Maybe add in comments that this still works for Ma or other agency clients.

app/app/views/caseworker/cbv_flow_invitations/_az_des.html.erb
```diff
@@ -0,0 +1,29 @@
+<% cbv_applicant = @cbv_flow_invitation.cbv_applicant ||
+     @cbv_flow_invitation.build_cbv_applicant(client_agency_id: @cbv_flow_invitation.client_agency_id) %>
```
jeffcatania-usds, line 2: In the MA view, we build_cbv_applicant with the snap_application_date set as today.  Why does this not apply here?

jeffcatania-usds: As a side note because this is the first time i've read this part of the codebase - is it best practices to build the cbv applicant in the view? I'd imagine that should be in the controller.

app/app/views/caseworker/cbv_flow_invitations/_az_des.html.erb
```diff
@@ -0,0 +1,29 @@
...
+<%= f.fields_for :cbv_applicant, cbv_applicant do |f2| %>
+  <%= f2.text_field :case_number, label: t(".invite.case_number") %>
```
jeffcatania-usds, line 4: Is the AZ value called "Case Number"? or do they refer to it as something different?

app/config/client-agency-config.yml
```diff
@@ -50,6 +51,38 @@
...
+  agency_name: Department of Economic Security
+  agency_short_name: DES
+  agency_contact_website: https://des.az.gov/services/basic-needs/food-assistance/contact-des
+  # No caseworker feedback form since they will only be interacting with the PDFs
+  # caseworker_feedback_form: null
+  pinwheel:
+    environment: <%= ENV["AZ_DES_PINWHEEL_ENVIRONMENT"] %>
+  argyle:
+    environment: <%= ENV["AZ_DES_ARGYLE_ENVIRONMENT"] %>
+  # TODO: this will be SFTP:
```
jeffcatania-usds, line 65: Is there a ticket for this? Can you put the ticket in the comment?

app/config/client-agency-config.yml
```diff
@@ -50,6 +51,38 @@
...
+  transmission_method: shared_email
+  transmission_method_configuration:
+    email: <%= ENV['SLACK_TEST_EMAIL'] %>
+  # TODO: Disable staff portal before pilot launch
```
jeffcatania-usds, line 69: As above - is there at ticket for this?

app/config/locales/en.yml
```diff
@@ -70,6 +74,11 @@ en:
         ma: Verify your income to renew your SNAP benefits
   caseworker:
     cbv_flow_invitations:
+      az_des:
```
jeffcatania-usds, line 77: Do we only support english for cbv flow invitations page?

app/config/locales/en.yml
```diff
@@ -11,6 +11,10 @@ en:
               blank: Enter the client's last name.
             snap_application_date:
               invalid_date: Enter today's date or the date you contacted the client.
+        cbv_applicant/az_des:
```
jeffcatania-usds, line 14: Did not see this on es? Do we not support?

app/app/views/caseworker/cbv_flow_invitations/_az_des.html.erb
```diff
@@ -0,0 +1,29 @@
+<% cbv_applicant = @cbv_flow_invitation.cbv_applicant ||
```
jeffcatania-usds, line 1: Can we attach a screenshot of what this page looks like on the PR description?

app/app/views/caseworker/entries/_az_des_signin_button.html.erb
```diff
@@ -0,0 +1 @@
+<%= button_to agency_translation("caseworker.entries.index.continue_to_login"), user_az_des_omniauth_authorize_path, data: { turbo: false }, class: "usa-button" %>
```
jeffcatania-usds, line 1: Can we attach a screenshot of what this button looks like on the PR description?

app/spec/factories/cbv_applicant.rb
```diff
@@ -35,5 +35,17 @@
...
+
+    trait :az_des do
+      client_agency_id { "az_des" }
+      first_name { nil }
+      middle_name { nil }
+      last_name { nil }
+
+      case_number do
+        # TODO: Determine actual AZ DES case number format.
+        8.times.map { rand(10) }.join
```
jeffcatania-usds, line 47: should this include underscores at this point?

app/.env
```diff
@@ -3,6 +3,7 @@
 # ##############################################################################
 NYC_PINWHEEL_ENVIRONMENT=sandbox
 MA_PINWHEEL_ENVIRONMENT=sandbox
+AZ_DES_PINWHEEL_ENVIRONMENT=sandbox
```
jeffcatania-usds, line 6: Idea: Do we have any area of documentation about all of our environment variables? It's starting to feel like we should have a separate file in the repo or a confluence document to document all of these and what the potential values are.

jeffcatania-usds (approved):
Nothing major.  I was extra thorough on this review because this is a new part of the codebase for me and its helpful for me to see the specific configuration needed to launch a pilot.

app/spec/controllers/api/invitations_controller_spec.rb
```diff
@@ -2,15 +2,15 @@
 
 RSpec.describe Api::InvitationsController do
   describe "#create" do
-    # must be existing user
+    let(:client_agency_id) { "ma".to_sym }
```
clediggins-usds, line 5: Does this work? Looks like 'create' will always have agency id being MA, even when we are trying to do AZ tests.

app/app/controllers/cbv/submits_controller.rb
tdooner: It's just something I thought would be nice (after testing the flow locally and being surprised to see the underscore in there).

For any other agency with underscores in their id, it will remove all of those as well in the confirmation numbers.

app/spec/models/cbv_applicant/az_des_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe CbvApplicant::AzDes, type: :model do
+  let(:ma_attributes) { attributes_for(:cbv_applicant, :az_des) }
```
clediggins-usds, line 4: Why is this ma_attributes for a az_des test?

app/spec/controllers/api/invitations_controller_spec.rb
```diff
@@ -2,15 +2,15 @@
...
     let(:api_access_token_instance) do
-      user = create(:user, :with_access_token, email: "test@test.com", client_agency_id: 'ma', is_service_account: true)
+      user = create(:user, :with_access_token, email: "test@test.com", client_agency_id: client_agency_id, is_service_account: true)
```
jeffcatania-usds, line 7: It's super weird that client_agency_id gets overwritten by the AZ test, but it's not apparent with how this is written. In JS, i would make this api_access_token_instance a function with a parameter being client_agency_id that returns the result of create.

app/app/models/cbv_applicant.rb
tdooner: Yeah - `camelize` should be generic to future agencies. This change was needed here as this is the first agency with an underscore in its name. So any future agencies that have IDs consisting of `[a-z_]` should just work.

app/app/views/caseworker/cbv_flow_invitations/_az_des.html.erb
tdooner: We are moving away from the `snap_application_date` field for new partner agencies since it was primarily used as the basis of the 90-day earnings lookback, and we think we can just use the current date instead.

tdooner: And you're right, it's not ideal to do this in the view. But we wouldn't have to do it unless the view was rendering the form. So it makes a little sense either way. In general, yeah, the controller makes more sense (and would allow this to be DRY across all the different agencies' views.)

app/app/views/caseworker/cbv_flow_invitations/_az_des.html.erb
tdooner: Yeah, it's "Case Number". No idea on the formatting rules on it, though. (We still need to ask.)

app/config/client-agency-config.yml
tdooner: Sure, yeah, it's [FFS-2628](https://jiraent.cms.gov/browse/FFS-2628).

app/config/locales/en.yml
tdooner: These are only used on the caseworker invitation page, which is English-only (and won't even be used for AZ. I only added it in this PR to make it easier for us to test with.)

app/spec/factories/cbv_applicant.rb
tdooner: We don't know whether their case numbers have underscores. I've seen a couple screenshots where they are just autoincrement integers.

app/.env
tdooner: Yeah - I'm trying to make this .env file be that documentation by listing all the environment variables that can be configured along with containing their default values.

app/spec/controllers/api/invitations_controller_spec.rb
tdooner: See line 45 (where we override this value to be "az_des") - although it may look weird to hardcode "ma" on this line, it's to allow us to override its value in a different rspec context.

app/spec/controllers/api/invitations_controller_spec.rb
tdooner: Yeah, I see where you're coming from, but that's just the rspec idiom - overriding variables like that within lower contexts, and re-using `let` instance variables from above to DRY up the object creation.

app/spec/models/cbv_applicant/az_des_spec.rb
tdooner: Copypasta. Fixing...

PR #533, opened 2025-04-03
FFS-2598 explore using polling instead of action cable to pull synchroni…

iannorriscms:
Showing what it might look like to make the queries polling-focused, since we're doing that anyway via check_pinwheel_account_synchrony.

Rather than resolve the issues mentioned in the ticket by tackling the action cable streaming, it seems like we already are treating the streaming as unreliable in a situation where we just want to wait for some actions we don't control to resolve---a great case for polling. 

Also, we can get rid of the POST hack if we turbo redirect.

Turbo redirect pattern found from https://www.ducktypelabs.com/turbo-break-out-and-redirect/

Things to figure out:

* what the safe backup strategy is to avoid polling infinitely
* how to properly debind it if a turbo redirect is occurring---did a hack for now because still net cleaner, but please let me know the proper way to deal with that!
* do we still want to keep actioncable in for other use cases? only usage so far.


app/app/views/cbv/synchronizations/show.html.erb
```diff
@@ -1,20 +1,4 @@
...
-  <% end %>
-  <div
-    class="grid-container"
-  >
-    <%= turbo_stream_from(@pinwheel_account) %>
-    <% if @pinwheel_account.present? %>
-      <%= render partial: "cbv/synchronizations/indicators", locals: { pinwheel_account: @pinwheel_account } %>
-    <% end %>
-  </div>
+<div data-controller="polling" data-polling-url-value="<%= cbv_flow_synchronizations_path(params: params) %>">
```
jeffcatania-usds, line 2: this broke without the params.to_unsafe_h

PR #534, opened 2025-04-03
FFS-2552: Add applicant information to caseworker PDF

millerti:
## Changes
The data on the applicant_information page will now show in the Caseworker version of the PDF report.

image

## Context for reviewers
In the future, the PDF will show when the user has edited the applicant_information data. But we don't load any data from our data providers yet, so it is not yet called out in these places. 

Note that I am currently having a discussion with Design about whether the prompts on the PDF should read "First name" or "Client's first name". Since this is just a string token change, I think it's safe to proceed with CR before that decision is made.

## Testing instructions
0. Complete the normal Generic Link flow (starting at localhost:3000/cbv/links/sandbox )
1. Opt to download the PDF
2. Verify that you still do NOT see the Client Information section 
3. Append ?is_caseworker=1 to the URL to trigger the Caseworker PDF generation. The Client Information section should appear near the top
4. Verify that it contains all the information provided by the client, and that all pieces of info provided by the client are denoted with the relevant superscript (in the future this will need to be more dynamic, but for now, all the info is provided by the client!)

## Acceptance testing


daphnegold (approved):
Looks amazing. Maybe a test eventually to make sure that these strings don't appear for clients. :shipit:

millerti:
AC testing notes: Both Daniella and Eryn approved the screenshot.

PR #535, opened 2025-04-03
FFS-2546: Allow Argyle JS to support "production" environment

daphnegold:
## Ticket

Resolves [FFS-2545](https://jiraent.cms.gov/browse/FFS-2545).

## Changes
- Argyle controller sets the `isSandbox` flag based on the environment set in `client_agency_config.rb`
- `ArgyleModalAdapter` can receive and handle environment via token endpoint request

## Acceptance testing

### DIFF
--- a/app/app/controllers/api/argyle_controller.rb
+++ b/app/app/controllers/api/argyle_controller.rb
@@ -3,7 +3,7 @@ class Api::ArgyleController < ApplicationController
     @cbv_flow = CbvFlow.find(session[:cbv_flow_id])
     argyle = argyle_for(@cbv_flow)
     user = argyle.create_user
-    is_sandbox_environment = ENV["ARGYLE_SANDBOX"].present? && ENV["ARGYLE_SANDBOX"] == "true"
+    is_sandbox_environment = ENV["ARGYLE_SANDBOX"] == "true"
 
     render json: { status: :ok, user: user, isSandbox: is_sandbox_environment }
   end
--- a/app/app/controllers/api/argyle_controller.rb
+++ b/app/app/controllers/api/argyle_controller.rb
@@ -3,7 +3,8 @@ class Api::ArgyleController < ApplicationController
     @cbv_flow = CbvFlow.find(session[:cbv_flow_id])
     argyle = argyle_for(@cbv_flow)
     user = argyle.create_user
+    is_sandbox_environment = ENV["ARGYLE_SANDBOX"].present? && ENV["ARGYLE_SANDBOX"] == "true"
 
-    render json: { status: :ok, user: user }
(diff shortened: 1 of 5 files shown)


millerti:
Seems like there may be a few tasks left on the ticket! Passing back.

app/app/controllers/api/argyle_controller.rb
```diff
@@ -3,7 +3,8 @@ def create
     @cbv_flow = CbvFlow.find(session[:cbv_flow_id])
     argyle = argyle_for(@cbv_flow)
     user = argyle.create_user
+    is_sandbox_environment = ENV["ARGYLE_SANDBOX"] == "true"
```
tdooner, line 6: We'll want this to be configurable per agency, rather than global, so we'll want to use the value in config/client-agency-config.yml for each agency under the `argyle.environment` key.

You should be able to get that environment by something like:

```ruby
  agency_config[@cbv_flow.client_agency_id].argyle_environment
```
(agency_config is a controller helper that loads the agency config. It's in ApplicationController.)

So that is the value we should be comparing against, and it would be something like:

```ruby
  agency_config[@cbv_flow.client_agency_id].argyle_environment == "sandbox"
```

tdooner (commented):
I synced up with Daphne regarding some of the ticket's items.

We talked about a way to test if there's a discrepancy between backend/frontend Argyle environment (by hardcoding `isSandbox: "true"` while setting SANDBOX_ARGYLE_ENVIRONMENT=production), but also agreed that it can be timeboxed to an hour since it's an unlikely failure case and we will be doing a lot of E2E testing.

The only blocker here is the per-agency config (see comment). After that, this lgtm.

daphnegold:
> We talked about a way to test if there's a discrepancy between backend/frontend Argyle environment (by hardcoding `isSandbox: "true"` while setting SANDBOX_ARGYLE_ENVIRONMENT=production), but also agreed that it can be timeboxed to an hour since it's an unlikely failure case and we will be doing a lot of E2E testing.

I verified that using mismatched environments (setting 'production' in the config while only having sandbox tokens) triggers an error in ArgyleService. This confirms security is enforced by the Argyle SDK, as expected. Tokens from one environment cannot be used in another, and environments use entirely different URLs. The `isSandbox` flag is set server side based on agency specific config and isn't influenced by client side code. We can do further testing in the future once we have production access and keys set up, with the whole system working together.

**Observed behavior**
1. Argyle modal displays a "Connection Error" and closes
Screenshot 2025-04-04 at 9 26 23 AM

2. Console logs indicate "User does not have environment access" 
Screenshot 2025-04-04 at 9 31 24 AM

millerti (approved):

PR #537, opened 2025-04-03
FFS-2598 Re-architect payment details sync

clediggins-usds:
## Ticket

Resolves FFS-2598

## Changes
* updated polling for completion of pinwheel api calls and refresh of page
* removed actioncable polling
* updated to use turboframe/stimulusjs refresh
* put the status indicators in a partial so they can be partially refreshed by turboframe
* updated translations for new partial

## Context for reviewers
Previous behavior: on slow connections the aggregator calls would loop, causing payment details to never show.
New behavior: on all connection speeds the refreshing function stops refreshing once the redirect has been initiated

Updated to use turboframe/stimulusjs best practices for refresh.

## Acceptance testing
- [] Acceptance testing after merge


app/config/locales/es.yml
```diff
@@ -243,6 +243,8 @@ es:
         income: Ingresos
         paystubs: Talones de pago
       show:
+        header: Estamos recopilando los detalles de pago de su empleador.
```
jeffcatania-usds, line 246: Is this new text or grabbed from somewhere else.  @millerti - does this look right to you?

app/app/javascript/controllers/polling_controller.js
```diff
@@ -0,0 +1,25 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+    static values = { url: String }
+    headers = { 'Accept': 'text/vnd.turbo-stream.html' };
+
+    connect() {
+        this.interval = setInterval(() => {
+            const csrfToken = document.querySelector("meta[name='csrf-token']").content;
```
jeffcatania-usds, line 9: use utilities.  There is the utilities/csrf.ts and utilities/fetchInternal that could be used here to resolve CSRF and/or help with wrapping fetch with CSRF.

app/app/views/cbv/synchronizations/show.html.erb
```diff
@@ -1,20 +1,4 @@
...
-  <% end %>
-  <div
-    class="grid-container"
-  >
-    <%= turbo_stream_from(@pinwheel_account) %>
-    <% if @pinwheel_account.present? %>
-      <%= render partial: "cbv/synchronizations/indicators", locals: { pinwheel_account: @pinwheel_account } %>
-    <% end %>
-  </div>
+<div data-controller="polling" data-polling-url-value="<%= cbv_flow_synchronizations_path(params: { user: { account_id: params.to_unsafe_h[:user][:account_id] } }) %>">
```
jeffcatania-usds, line 2: This is hard to read. It could benefit from a breakpoint or two. 

Also - are we sure we want to use to_unsafe_h?  Is there a safer way to handle this?

jeffcatania-usds (changes requested):
Great job! Looks like a very straightforward approach to the synchronizations page.  A few small nits.

app/app/views/cbv/synchronizations/show.html.erb
iannorriscms: i think we can remove to unsafe_h --> that was a legacy of when we were doing the whole params instead of just grabbing the user_account_id

app/app/javascript/controllers/index.js
```diff
@@ -1,15 +1,15 @@
...
 
 import CbvEmployerSearch from "./cbv/employer_search"
-import CbvSynchronizationsController from "./cbv/synchronizations_controller"
 import CbvSessionsTimeoutController from "./cbv/sessions_controller.js"
 import HelpController from "./help"
+import PollingController from "./polling_controller.js"
 
 application.register("cbv-employer-search", CbvEmployerSearch)
-application.register("cbv-synchronizations", CbvSynchronizationsController)
+application.register("polling", PollingController)
```
tdooner, line 9: Hmm. I'm not sure the generalized name `PollingController` will end up being useful on other pages, but I suppose I'm ok with hoping it will be.

app/app/views/cbv/synchronizations/_status.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= turbo_frame_tag "synchronization" do %>
+  <h1>
+    <%= t(".header") %>
+  </h1>
+  <p class="usa-prose">
+    <%= t(".fetching_payroll_description") %>
+  </p>
+    <div class="grid-container">
+    <% if @pinwheel_account.present? %>
+      <%= render partial: "cbv/synchronizations/indicators", locals: { pinwheel_account: @pinwheel_account } %>
```
tdooner, line 10: Hmm, at some point maybe we should remove one layer of these partials. We now have "status" rendering "indicators" which renders 4 "indicator" partials. Maybe "status" could render the 4 "indicator"s directly. (Could be done as a separate cleanup.)

app/config/locales/es.yml
tdooner: Looks like it's duplicated from two lines below on account of being used both in the top-level "show" view and the "status" partial. Can we DRY it up and use the same key by either doing

```
content_for :title, t("cbv.synchronizations.status.header")
```

or moving the `<h1>` out of the turbo frame?

tdooner (approved):
Approach seems fine in general once addressing Jeff's comments. It will have a merge conflict with George's PR about the webhooks, but I can rebase that PR on this.

I am currently on a plane. I will test and let you know how it works. (Even a windows machine is faster than plane Wi-fi!)

app/app/javascript/controllers/index.js
iannorriscms: might be a different understanding of stimulus controllers--but my understanding of stimulus is that you remove as much domain from the actions as possible, and its mostly about their function--so instead of a CbvFlowController, its orienting around polling and refreshing and whatever DOM manipulation you're supposed to be doing.

app/app/views/cbv/synchronizations/_status.html.erb
iannorriscms: agreed, lets make that a seperate followup refactor to merge the indicators back into the status partial.

app/config/locales/es.yml
clediggins-usds: This is pulled from review, not a new text

app/app/views/cbv/synchronizations/_status.html.erb
clediggins-usds: As Ian said, changing the indicators should be part of a different ticket. There is follow-on work to make these loading indicators work with both pinwheel and argyle and this will change anyway. for the scope of this ticket, this partial is effective.

app/app/javascript/controllers/index.js
tdooner: Makes sense. And I do like the idea of using stimulus for *components within a page* rather than the entire page itself, just to keep the stimulus controllers somewhat tighter-scoped in responsibility. We kind of abuse it with the CbvEmployerSearch page, but I suppose we also need some way to orchestrate page-level things as well.

app/app/views/cbv/synchronizations/show.html.erb
clediggins-usds: to_unsafe_h is not necessary here. removed

app/app/javascript/controllers/index.js
clediggins-usds: controller is generic, and anywhere that we are using turboframe redirect this can be used.

app/config/locales/es.yml
millerti: Yeah I recommend Tom's solution here -- if we're using the same string, might as well save the translation money! Unless the meaning of the token is really different in these two places, but I doubt it.

app/app/javascript/controllers/polling_controller.js
clediggins-usds: updated

app/config/locales/es.yml
jeffcatania-usds: @clediggins-usds - it looks like the request is to remove these translation changes and either reference the existing variable or the <h1> out of the frame.

jeffcatania-usds (approved):

PR #538, opened 2025-04-03
FFS-??? create base feature spec for creating a new client

iannorriscms:
## Changes

Implements changes to enable running a feature test. If this looks good to folks, I will happily expand it to capture the full end to end flow.


iannorriscms:
TODO: investigate merging this into the current e2e tests, and coming back to understanding how to set up those tests in a way that is reliable.

PR #541, opened 2025-04-07
FFS-2540-4: corrected sending some attributes to the event logger

millerti:
## Changes
Corrected an issue Eryn noticed during acceptance testing: not all new events were properly getting the client_agency_id or locale attributes set. I fixed this for the new events and re-tested many of the existing ones.

There are a few events directly from pinwheel that still don't get these 100% correctly, but I think that's acceptable since so many other events do; it will be easy to determine what the values for these attributes likely were if they're missing, since they'll be on a user with many other events with these attributes set correctly.

## Acceptance testing

### DIFF
--- a/app/app/controllers/cbv/applicant_informations_controller.rb
+++ b/app/app/controllers/cbv/applicant_informations_controller.rb
@@ -64,12 +64,14 @@ class Cbv::ApplicantInformationsController < Cbv::BaseController
       event_logger.track("ApplicantClickedEditInformationLink", request, {
         timestamp: Time.now.to_i,
         cbv_applicant_id: @cbv_flow.cbv_applicant_id,
+        client_agency_id: current_agency&.id,
         cbv_flow_id: @cbv_flow.id
       })
     else
       event_logger.track("ApplicantAccessedInformationPage", request, {
         timestamp: Time.now.to_i,
         cbv_applicant_id: @cbv_flow.cbv_applicant_id,
+        client_agency_id: current_agency&.id,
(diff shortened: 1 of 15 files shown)


app/app/controllers/cbv/applicant_informations_controller.rb
```diff
@@ -64,12 +64,14 @@ def track_applicant_information_access_event
       event_logger.track("ApplicantClickedEditInformationLink", request, {
         timestamp: Time.now.to_i,
         cbv_applicant_id: @cbv_flow.cbv_applicant_id,
+        client_agency_id: current_agency&.id,
```
jeffcatania-usds, line 67: What is this syntax? I haven't seen it before.  Why can we not use "current_agency.id" directly?

jeffcatania-usds: Answer to my own question:

It is called the Safe Navigation Operator. Introduced in Ruby 2.3.0, it lets you call methods on objects without worrying that the object may be nil(Avoiding an undefined method for nil:NilClass error), similar to the [try method in Rails](http://api.rubyonrails.org/v4.2/classes/Object.html#method-i-try).

So you can write

@person&.spouse&.name
instead of

@person.spouse.name if @person && @person.spouse

jeffcatania-usds: Why would current_agency ever be null?

jeffcatania-usds (approved):
LGTM

PR #542, opened 2025-04-08
FFS-2549: Spanish translations for sprint 12.2 and earlier

millerti:
## Changes
This ticket includes all missing Spanish translations for the past few sprints.
Note that there are currently two missing strings that will need to be translated in a future ticket. These are noted in the testing instructions below.

## Testing instructions
In the Spanish locale, verify that the following sections of the site seem to be translated properly:
1. The expired invitations page (see screenshot below)
2. The help modal
3. The timeout page - these strings are translated, but there is no Español button on this page. I think we should write a new ticket to make sure users in the ES locale get an ES timeout page.
4. The 404 and 500 error pages
5. The invalid link page - generate a new invite, then change one of the characters in the generated URL to get to this page.
6. The /applicant_information page
7. The /submit page
8. The /summary page (there are two exceptions here: the description and the header, which will be translated in a later ticket)

Here's a screenshot of the expired invitation since that's a bit of a pain to get to:
image

## Acceptance testing


app/app/javascript/controllers/help.js
```diff
@@ -10,7 +10,7 @@ export default class extends Controller {
         source: event.target.dataset.source,
       })
       // reset the help modal src on mousedown to ensure the help modal src is reset to "/help"
-      document.querySelector("#help_modal_content").src = "/help"
+      document.querySelector("#help_modal_content").src = event.target.dataset.helpUrl
```
millerti, line 13: These helpUrl changes allow the help modal to respond to the locale switch. Previously, we were hardcoding this to the "/help" path, which does not include the locale piece. Since English is the default locale, it was never switching to Spanish. Now we store the URL with the locale included in _help_modal.html.erb and use it here.

daphnegold (commented):
Code is looking good, but the help modal still appears in English for me. I like the changes you're going for though. Am I testing wrong?

Screenshot 2025-04-09 at 10 45 40 AM

PR #547, opened 2025-04-08
FFS-2501: Track analytics of payroll aggregator fields

tdooner:
## Ticket

Resolves [FFS-2501](https://jiraent.cms.gov/browse/FFS-2501).

Pinwheel:
image

Argyle:
image

## Changes

- **Add fields for DOB/Email/SSN/Phone to ResponseObjects::Identity**
- **Track which fields reports contain when they finish syncing**

## Acceptance testing

### DIFF
--- a/app/app/controllers/webhooks/argyle/events_controller.rb
+++ b/app/app/controllers/webhooks/argyle/events_controller.rb
@@ -131,6 +131,12 @@ class Webhooks::Argyle::EventsController < ApplicationController
         paystubs_deductions_count: report.paystubs.sum { |p| p.deductions.length },
         paystubs_hours_by_earning_category_count: report.paystubs.sum { |p| p.hours_by_earning_category.length },
         paystubs_hours_present: report.paystubs.first&.hours.present?,
+        paystubs_earnings_count: report.paystubs.sum { |p| p.earnings.length },
+        paystubs_earnings_with_hours_count: report.paystubs.sum { |p| p.earnings.count { |e| e.hours.present? } },
+        paystubs_earnings_type_base_count: report.paystubs.sum { |p| p.earnings.count { |e| e.category == "base" } },
+        paystubs_earnings_type_bonus_count: report.paystubs.sum { |p| p.earnings.count { |e| e.category == "bonus" } },
+        paystubs_earnings_type_overtime_count: report.paystubs.sum { |p| p.earnings.count { |e| e.category == "overtime" } },
+        paystubs_earnings_type_commission_count: report.paystubs.sum { |p| p.earnings.count { |e| e.category == "commission" } },
 
         # Employment fields (originally from "identities" endpoint)
         employment_success: @payroll_account.job_succeeded?("employment"),
@@ -145,7 +151,7 @@ class Webhooks::Argyle::EventsController < ApplicationController
         # Gigs fields
         gigs_success: @payroll_account.job_succeeded?("gigs"),
         gigs_supported: @payroll_account.supported_jobs.include?("gigs")
-        # TODO: Add gig fields
+        # TODO: Add fields from /gigs after FFS-2575.
       })
     end
   rescue => ex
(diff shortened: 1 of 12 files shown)


app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -92,17 +92,60 @@ def track_events(webhook_event)
...
+        employment_employer_phone_number_present: report.employments.first&.employer_name&.present?,
+        employment_start_date: report.employments.first&.start_date,
+        employment_termination_date: report.employments.first&.termination_date,
+
+        # Gigs fields
         gigs_success: @payroll_account.job_succeeded?("gigs"),
-        gigs_supported: @payroll_account.supported_jobs.include?("gigs"),
-        sync_duration_seconds: Time.now - @payroll_account.created_at
+        gigs_supported: @payroll_account.supported_jobs.include?("gigs")
+        # TODO: Add gig fields
```
tdooner, line 148: This TODO is because Jeff and Daphne are currently working on implementing the gig/shifts endpoints, so if I end up rebasing on their work, it's possible to resolve the TODO. If I merge first, then I'll have to come back to this.

app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -92,17 +92,66 @@ def track_events(webhook_event)
...
+      )
+      report.fetch
+
       event_logger.track("ApplicantFinishedArgyleSync", request, {
         cbv_applicant_id: @cbv_flow.cbv_applicant_id,
         cbv_flow_id: @cbv_flow.id,
         invitation_id: @cbv_flow.cbv_flow_invitation_id,
-        identities_success: @payroll_account.job_succeeded?("identities"),
-        identities_supported: @payroll_account.supported_jobs.include?("identities"),
+        sync_duration_seconds: Time.now - @payroll_account.created_at,
```
jeffcatania-usds: I am hesitant to keep this entirely within the provider-specific events_controller.  It seems like 80% of this is consistent between providers, with the exception of @payroll_account.job_succeeded? calls and specific areas for paystub earning categories.

My other concern, especially without DRY code, is how easy this would be to in the future introduce PII data to the logs by dropping the present? tags.  At minimum, we should protect this with a comment.

app/spec/rails_helper.rb
```diff
@@ -41,7 +41,7 @@
 end
 RSpec.configure do |config|
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
-  config.fixture_path = "#{::Rails.root}/spec/fixtures"
+  # config.fixture_path = "#{::Rails.root}/spec/fixtures"
```
jeffcatania-usds, line 44: Did you mean to comment this out?

jeffcatania-usds (changes requested):

app/spec/rails_helper.rb
tdooner: Yeah - I should have broken this into the other "random stuff" PR. It eliminates a deprecation warning we get every time we run the test suite:
image

Since we don't use Rails fixtures (instead relying on FactoryBot factories), we don't need to even configure this at all.

app/app/controllers/webhooks/argyle/events_controller.rb
tdooner: I'll add a scary warning comment.

We can touch base to figure out what to do about DRYing it up. I'm open to it if you can think of a place that makes sense, but I did start down the path of trying to DRY this up into some place that inferred these fields dynamically. It got really complicated, though, and then we would still probably still want the 20% customization on a per-field-level anyway, so I thought doing it this way was not the most terrible thing because even though it's kind of massive, at least it's straightforward to understand.

jeffcatania-usds (approved):

tdooner:
Going to merge this and then seek acceptance testing from the demo environment.

PR #549, opened 2025-04-09
Jeffcatania usds/e2e rspec tests

jeffcatania-usds:
## Ticket

## Acceptance testing


iannorriscms:
Outstanding work left:

* switching the driver to selenium_chrome_headless changed the nature of the puffing billy mocks required to simulate moving through the flow. That will require tweaking so it passes locally and in rspec CI.
* the help specs need to be rewritten in the VCR/puffing billy format so they can pass as well.
* the remainder of the e2e flow for pinwheel needs to be fleshed out (generating a pdf, simulating webhooks for all events)
* for the simulation of webhooks, there might be a way to fake encrypt the message so it can be decrypted by the webhooks controller. at worst case, refactoring so the simulation doesn't drift would be desirable. 
* once the flow is fleshed out for pinwheel, a followup story to do the same for argyle also should happen.

tdooner:
I've done a spike to address all the outstanding TODO items on this PR in https://github.com/DSACMS/iv-cbv-payroll/pull/786, so I'm going to close this out.

PR #550, opened 2025-04-09
FFS-2607: Add `employment_type` field to ResponseObjects::Employment

tdooner:
## Ticket

Resolves [FFS-2607](https://jiraent.cms.gov/browse/FFS-2607).

## Changes

* **Add `employment_type` field to ResponseObjects::Employment**

## Context for reviewers

This field, whose value is currently only `:w2` or `:gig`, is determined
by our own logic so we can use it in future stories (FFS-2542):
* For Pinwheel, we use a list of employer names from Pinwheel's support
  team. We have to match the name because the /employments endpoint
  doesn't have any ID to associate that employment with the employer
  that was linked.
* For Argyle, we can just look at whether they classify the user as a
  "contractor".

## Acceptance testing

### DIFF
--- a/app/app/services/aggregators/sdk/argyle_service.rb
+++ b/app/app/services/aggregators/sdk/argyle_service.rb
@@ -168,7 +168,7 @@ module Aggregators::Sdk
       @http.get(GIGS_ENDPOINT, params).body
     end
 
-    # https://docs.argyle.com/api-reference/gigs#list
+    # https://docs.argyle.com/api-reference/shifts#list
     def fetch_shifts_api(**params)
       @http.get(SHIFTS_ENDPOINT, params).body
     end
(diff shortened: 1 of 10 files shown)


app/spec/controllers/webhooks/argyle/events_controller_spec.rb
```diff
@@ -111,6 +111,9 @@ def process_webhook(event_type)
         allow_any_instance_of(Aggregators::Sdk::ArgyleService)
           .to receive(:fetch_paystubs_api)
           .and_return(argyle_load_relative_json_file("sarah", "request_paystubs.json"))
+        allow_any_instance_of(Aggregators::Sdk::ArgyleService)
+          .to receive(:fetch_gigs_api)
+          .and_return(argyle_load_relative_json_file("bob", "request_gigs.json"))
```
tdooner, line 116: I broke this out into https://github.com/DSACMS/iv-cbv-payroll/pull/552 as well, in case we need to merge it independently (since it fixes a test failure on `main`).

app/app/services/aggregators/sdk/argyle_service.rb
```diff
@@ -168,7 +168,7 @@ def fetch_gigs_api(account: nil, user: nil,
       @http.get(GIGS_ENDPOINT, params).body
     end
 
-    # https://docs.argyle.com/api-reference/gigs#list
+    # https://docs.argyle.com/api-reference/shifts#list
```
jeffcatania-usds, line 171: good catch

jeffcatania-usds (approved):

PR #557, opened 2025-04-11
[FFS-2739] when the poller is firing to collect information, dont fire a pageview event

iannorriscms:
we only want to track the initial hitting of the page (which is done via the show action) when we're waiting for the webhooks to sync, rather than collect the poller as well.

## Ticket

Resolves [FFS-2739](https://jiraent.cms.gov/browse/FFS-2739).

## Changes

Ensures tracking not applied when firing the update event, because that is polled for and therefore adds noise.

## Acceptance testing

  * You can test this by confirming that the PageView events only fire once when you hit the synchronizations page.


jeffcatania-usds (approved):
LGTM

PR #560, opened 2025-04-14
FFS-2542: Validate that reports are "useful"

jeffcatania-usds:
## Ticket
Resolves [FFS-2542](https://jiraent.cms.gov/browse/FFS-2542).

## TODO

- [x] Add is_w2 calculation into EventsController's validate_minimum_reporting_requirements method.
- [x] Test conditions for is_w2.  "bob"'s payroll argyle fixture should fail if is_w2 is set to true. It should succeed if set to false. 
- [x] Update Webhooks::Pinwheel::EventsController
- [x] Update rspec tests
- [x] Smoke test

Test cases to verify:
- [x] Pinwheel successful sync (meets useful requirements)
- [x] Pinwheel with `user_data_not_available` (fails)
- [x] Pinwheel with `voe_and_voie_unavailable` (fails)
- [x] Argyle successful sync
- [ ] Argyle gig user

## Changes

Introduced Minimum Reporting Requirements check
- Redirects users to the Synchronizations Failures page when minimum reporting requirements fail
- Added the dry-validation gem to the Gemfile, enabling validation of business logic requirements.
- Introduced a new MinimumReportingRequirementsSchema class in app/services/aggregators/contracts/ to validate minimum reporting requirements.

Updated Webhooks::Argyle::EventsController:
- Renamed the track_events method to process_webhook_event.
- Added logic to log synchronization finish and validate minimum reporting requirements.
- Introduced error handling and logging for tracking and validation failures.
- Added a redirect to synchronization failure page when validation requirements are not met.

## Context for reviewers
- This PR enhances the application's ability to validate and enforce minimum reporting requirements for payroll account synchronization.
- The dry-validation gem was chosen for its flexibility and readability in defining schema-based rules.
Reviewers should pay attention to:
- The correctness of the MinimumReportingRequirementsSchema validations.
- The error handling and logging mechanisms added to the event processing workflow.

## Acceptance testing

- [ ] No acceptance testing needed
  * This change will not affect the user experience (bugfix, dependency updates, etc.)
- [x] Acceptance testing prior to merge
  * This change can be verified visually via screenshots attached below or by sending a link to a local development environment to the acceptance tester
  * Acceptance testing should be done by **design** for visual changes, **product** for behavior/logic changes, **or both** for changes that impact both.
- [ ] Acceptance testing after merge
  * This change is hard to test locally, so we'll test it in the demo environment (deployed automatically after merge.)
[...]

PR #561, opened 2025-04-15
Fix translate_aggregator_value to handle `nil` value

tdooner:
## Ticket

N/A - Tiny bugfix needed for #560.

## Context for reviewers

This was breaking by returning a hash when value was nil, because it
would look for a key like "aggregator_reports.employment_statuses.",
which I18n would happily resolve to:

```
{:employed=>"employed", :inactive=>"inactive", :terminated=>"terminated"}
```

This behavior seems incorrect, so we should ensure that a nil value in
results in a nil value out.

## Acceptance testing

### DIFF
--- a/app/app/helpers/view_helper.rb
+++ b/app/app/helpers/view_helper.rb
@@ -37,6 +37,8 @@ module ViewHelper
   end
 
   def translate_aggregator_value(namespace, value)
+    return unless value.present?
+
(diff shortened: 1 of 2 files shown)


millerti (approved):
Ah ok, so when value was nil, it was just returning the hash of aggregator_strings.#{namespace}? I think I agree with your solution, and I also think I agree that we shouldn't log the Unknown Aggregator Value warning in this case, so I'll approve.

PR #564, opened 2025-04-15
FFS-2694: Added the LA LDH client agency configuration

GeorgeCodes19:
## Ticket

Resolves [FFS-2694](https://jiraent.cms.gov/browse/FFS-2694).

- [x] Add LA LDH CbvApplicant subclass and model
- [x] Write tests that ensures that a `CbvApplicant::LaLdh` instance can be properly created
- [x] Update any applicable terraform/infrastructure with `LA_PINWHEEL_ENVIRONMENT`

## Changes

- Added LA LDH (Louisiana Department of Health) configuration the client agency configuration
- Added tests and partial for collecting/correcting indexing data

## Acceptance testing

1. Run the app locally and visit [http://localhost:3000/cbv/links/la_ldh](http://localhost:3000/cbv/links/la_ldh) to initiate a generic link flow


app/app/models/cbv_applicant/la_ldh.rb
```diff
@@ -0,0 +1,13 @@
...
+  VALID_ATTRIBUTES = %i[
+    first_name
+    middle_name
+    last_name
+    case_number
+  ]
+
+  # Making case_number optional.
+  # TODO: Confirm the case_number format.
+  validates :case_number, presence: false
```
GeorgeCodes19, line 12: Technically we can simply omit this field, but I wanted to be a bit explicit here. This line also acts as a placeholder for validating the optional `case_number` format. What is the `case_number` format?

app/app/views/cbv/applicant_informations/_la_ldh.html.erb
```diff
@@ -0,0 +1,7 @@
+<% cbv_applicant = @cbv_flow.cbv_applicant %>
+<%= f.fields_for :cbv_applicant, cbv_applicant do |f2| %>
+  <%= f2.text_field :first_name, label: t(".fields.first_name.prompt") %>
```
GeorgeCodes19, line 3: Note, there are translations registered for these indexing fields.

Screenshot 2025-04-15 at 2 45 22 PM

tdooner: Let's fix this before merging.

app/app/views/cbv/applicant_informations/_la_ldh.html.erb
```diff
@@ -0,0 +1,7 @@
...
+  <%= f2.text_field :middle_name, label: t(".fields.middle_name.prompt") %>
+  <%= f2.text_field :last_name, label: t(".fields.last_name.prompt") %>
```
tdooner, line 5: The name fields are actually not called-for in the story. We only need the case number.

app/config/client-agency-config.yml
```diff
@@ -83,6 +83,34 @@
...
+  #   client_id:     <%= ENV["AZURE_LA_LDH_CLIENT_ID"] %>
+  #   client_secret: <%= ENV["AZURE_LA_LDH_CLIENT_SECRET"] %>
+  #   tenant_id:     <%= ENV["AZURE_LA_LDH_TENANT_ID"] %>
+  #   scope: "openid"
+  #   name: "la_ldh"
+  pay_income_days: 90
+  invitation_valid_days: 14
+  # TODO[FFS-2690]: Enable the weekly report
+  # weekly_report:
+  #   recipient: <%= ENV['LA_LDH_WEEKLY_REPORT_RECIPIENTS'] %>
```
tdooner, line 112: We're going to need an `applicant_attributes` here, in order for it to work with the generic link, I expect?

app/spec/factories/cbv_applicant.rb
```diff
@@ -75,5 +75,14 @@
         ]
       end
     end
+
+    trait :la_ldh do
+      client_agency_id { "la_ldh" }
+
+      case_number do
+        # TODO: Determine actual LA LDH case number format.
```
tdooner, line 83: Can you ask in the LA pilot slack channel? I don't remember whether we know this.

app/app/models/cbv_applicant/la_ldh.rb
tdooner: I'd lean against including it. We can always add it back in when we want to do a validation. That, in my opinion, is more maintainable than keeping disabled code checked in.

app/app/views/cbv/applicant_informations/_la_ldh.html.erb
tdooner: Also, what happened to the agency name in the header? You may need some I18n strings in en.yml in order for this to work. (Check out the other PR linked in the ticket for inspiration)

app/config/client-agency-config.yml
GeorgeCodes19: Whoops, yeah I missed this. It makes sense since these fields are validated when supplying applicant info.

app/spec/factories/cbv_applicant.rb
GeorgeCodes19: This will have to remain a to do until Mak gets this from LA, but it's on the radar

infra/app/app-config/env-config/environment-variables.tf
```diff
@@ -114,6 +114,10 @@ locals {
       manage_method     = "manual"
       secret_store_name = "/service/${var.app_name}-${var.environment}/ma-pinwheel-environment"
     },
+    LA_LDH_PINWHEEL_ENVIRONMENT = {
+      manage_method     = "manual"
+      secret_store_name = "/service/${var.app_name}-${var.environment}/la-ldh-pinwheel-environment"
+    },
```
tdooner, line 120: Reminder to create this parameter in AWS before merging so that the deploy will succeed.

app/config/client-agency-config.yml
```diff
@@ -83,6 +83,37 @@
...
+  agency_short_name: LDH
+  agency_contact_website: https://ldh.la.gov/contact-us
+  # caseworker_feedback_form: n/a
+  # TODO[FFS-2689]: Add Louisiana DPH logo, these assets are not available yet.
+  # logo_path: ldh_logo.png
+  # logo_square_path: ldh_logo_square.png
+  pinwheel:
+    environment: <%= ENV["LA_LDH_PINWHEEL_ENVIRONMENT"] %>
+  argyle:
+    environment: <%= ENV["SANDBOX_ARGYLE_ENVIRONMENT"] %>
```
tdooner, line 97: This needs an environment variable too.

app/app/models/cbv_applicant/la_ldh.rb
tdooner: Did you plan to address my comment? If you disagree, that's fine, I don't think this is a blocker, but it would be nice to have a reply with your thought process.

tdooner (approved):
Looks good once you make a LA_LDH_ARGYLE_ENVIRONMENT variable.

infra/app/app-config/env-config/environment-variables.tf
GeorgeCodes19: Thanks for the reminder! I took care of it earlier this morning 👍🏾

app/app/models/cbv_applicant/la_ldh.rb
GeorgeCodes19: I think we were referring to two separate things which lead to some confusion. When I mentioned

> Technically we can simply omit this field

I meant that we don't need to include this line of code in the file at all and that it will still be treated as optional without an explicit `validates` clause.

Are you referring to making it a required field until we know the validation format?

PR #566, opened 2025-04-15
FFS-2753: Update /employer_search precedence logic

tdooner:
## Ticket

Resolves [FFS-2753](https://jiraent.cms.gov/browse/FFS-2753).

## Changes

**Update /employer_search precedence logic**
> Upon further discussion of requirements with Eryn, we want to update the
> precedence logic to be:

> 1. Search Argyle. If any result is an exact match (after basic string
>    normalization), then stop and show all Argyle results. If no exact
>    match, then
> 2. Search Pinwheel. If any result is an exact match (after basic string
>    normalization), then stop and show all Pinwheel results. If no exact
>    match, then
> 3. Show Argyle results, if any. If none, then
> 4. Show Pinwheel results.

> This commit implements that and refactors the tests a bit to be able to
> operate without stubbing the constant.

**Update Argyle search params to exclude unavailable results**
> [Argyle's API docs][1] say that some search results may be returned that
> have status = "unavailable". These cannot be linked, therefore let's
> hide them from search results and hope our fallback-to-pinwheel logic
> will fulfill the user need in linking these employers.

> [1]: https://docs.argyle.com/api-reference/items#object-status

## Acceptance testing

### DIFF
--- a/app/app/services/provider_search_service.rb
+++ b/app/app/services/provider_search_service.rb
@@ -20,7 +20,7 @@ class ProviderSearchService
     if @providers.include?(:pinwheel) && (results.length == 0 || !any_exact_matches?(results, query))
       pinwheel_service = Aggregators::Sdk::PinwheelService.new(@client_agency_config.pinwheel_environment)
 
-      pinwheel_results = pinwheel_service.fetch_items(q: query)["data"].map do |result|
+      pinwheel_results = pinwheel_service.fetch_items(q: query, supported_jobs: %w[paystubs])["data"].map do |result|
         Aggregators::ResponseObjects::SearchResult.from_pinwheel(result)
       end
 
--- a/app/app/services/provider_search_service.rb
+++ b/app/app/services/provider_search_service.rb
@@ -1,22 +1,32 @@
 class ProviderSearchService
   SUPPORTED_PROVIDERS = (ENV["SUPPORTED_PROVIDERS"] || "pinwheel")&.split(",")&.map(&:to_sym)
 
-  def initialize(client_agency_id)
+  def initialize(client_agency_id, providers: SUPPORTED_PROVIDERS)
     @client_agency_config = site_config[client_agency_id]
+    @providers = providers
   end
 
   def search(query = "")
     results = []
-    if SUPPORTED_PROVIDERS.include?(:argyle)
-      results = Aggregators::Sdk::ArgyleService.new(@client_agency_config.argyle_environment).items(query)["results"].map do |result|
+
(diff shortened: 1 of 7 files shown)


tdooner:
Eryn and I huddled yesterday to acceptance test this, and she gave it the thumbs-up after using the following test cases:

* Exact match in both Argyle and Pinwheel
  * "American Tire Distributors" - Argyle is preferred
  * "Allegion" - There is an exact match in both Argyle and Pinwheel, but Argyle's result is status="unavailable" so we fall back to Pinwheel's direct match
* Exact match in Argyle only
  * "Nava PBC" - Argyle is preferred
* Exact match in Pinwheel only
  * "Barings LLC" - Pinwheel is preferred
* No exact match in either
  * "Barings" - Argyle
  * "Nava" - Argyle

app/spec/support/pinwheel_api_helper.rb
```diff
@@ -10,16 +10,7 @@ def pinwheel_stub_request_items_response
...
-          data: [
-            {
-              id: "12345",
-              name: "Some Employer Name",
-              logo_url: "https://example.com/logo.jpg",
-              response_type: "employer"
-            }
-          ]
-        }.to_json,
+        body: pinwheel_load_relative_json_file('request_items_response.json').to_json,
```
GeorgeCodes19, line 13: Good catch, we should be sticking with fixtures when possible.

GeorgeCodes19 (approved):
LGTM!

app/app/services/provider_search_service.rb
```diff
@@ -1,22 +1,32 @@
...
 
   def search(query = "")
     results = []
-    if SUPPORTED_PROVIDERS.include?(:argyle)
-      results = Aggregators::Sdk::ArgyleService.new(@client_agency_config.argyle_environment).items(query)["results"].map do |result|
+
+    if @providers.include?(:argyle)
+      argyle_service = Aggregators::Sdk::ArgyleService.new(@client_agency_config.argyle_environment)
+
+      results = argyle_service.items(query)["results"].map do |result|
```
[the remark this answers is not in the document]
tdooner: I see what you're saying, that the logic of reassigning `results` makes this method somewhat hard to read. I think I'll defer that for now though. The test cases will make refactoring that later easy.

app/spec/support/pinwheel_api_helper.rb
tdooner: Yeah, I was surprised by the discrepancies between this file and what was being returned by the stub method!

PR #567, opened 2025-04-16
[FFS-2750] submit to client asynchronously via solid queue

iannorriswork:
Resolves [FFS-2750](https://jiraent.cms.gov/browse/FFS-2750).

## Changes

 This removes sidekiq, which was kinda the active job setup kinda not, in favor of keeping everything in database via solid queue.
 Extracts out most of the work done by the submits controller into a contained class. Did some refactoring of the tests while I was at it,
 but there are some extra test cases to add in a follow up PR.

 There are two choices that I think need to be made as a team, with some tradeoffs in what that entails.

 Tradeoff 1: do we like the Procfile suggests run the puma server seperately or jointly with the job manager?

 In the default gem configuration, it suggests we can add a plugin to Puma that would allow idle servers to pick off jobs off the queue. Personally not a fan of that which is why I went for seperate servers in dev, but worth a discussion. (not a fan due to the 12 factor willies that gives to have the background processes hog front end user servers potentially.)

Conclusion:
After discussion with Tom, we think, especially due to the PDF generation jobs being memory intensive, it's safer if we have separate Rails containers for the job servers. Will be doing a followup PR to enable this for demo and production.

 Tradeoff 2: do we do as they suggest and put in a seperate server configuration for production, or have them combined?
 Interestingly, they do suggest that the database be a seperate process by default-- 'Running Solid Queue in a separate database is recommended'. Given the autoscaling nature of our DB and small size I could see an argument for and against choosing to do so--> wanting to leave that up to the team before making a call.

Conclusion; 
After discussion with Tom, we decided that given our job size spawning will be pretty deliberate and small, and the nature of the Postgres autoscaling, that we're safe with keeping them in the same database for now. We can revisit this later if the scale and nature of the jobs we enqueue changes.

## Acceptance testing

  * Turn ACTIVEJOB_ENABLED to true
  * Go through the flow and generate a pdf
  * See the pdf is sent to the client

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -26,9 +26,6 @@ brew "graphviz"
 # used in many scripts in "infra"
 brew "jq"
 
-# queue for sidekiq jobs
-brew "redis"
-
 # ngrok local tunnel to receive webhooks
 cask "ngrok"
 
--- a/README.md
+++ b/README.md
@@ -48,9 +48,8 @@ Most developers on the team code using macOS, so we recommend that platform if p
 1. Install JS dependencies
    * `nodenv rehash`
    * `npm install`
-1. Start postgres & redis:
+1. Start postgres:
    * `brew services start postgresql@12`
-   * `brew services start redis`
 1. Get development credentials from 1Password: search for "CBV .env.local secrets" and copy its ".env.local" section into a file called that in the app directory.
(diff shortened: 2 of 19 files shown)


iannorriswork:
closing in favor of a branch on the DSACMS doing this

PR #568, opened 2025-04-16
[FFS-2662] fix sandbox via introducing something for it to validate

iannorriswork:
i also like the spirit of the shared example test, but think a more direct call of the configuration drives which tests are required

## Ticket

Resolves ability to do sandbox invitations.

## Changes
Fix sandbox invitation flow to be runnable again.

Restructures the tests so this will be less likely to happen in the future.

## Acceptance testing

  * Can attempt a sandbox invite.

### DIFF
--- a/app/spec/models/cbv_applicant_spec.rb
+++ b/app/spec/models/cbv_applicant_spec.rb
@@ -1,12 +1,14 @@
 require 'rails_helper'
 
-RSpec.shared_examples "a CbvApplicant subclass" do |agency_id|
-  it "has a list of VALID_ATTRIBUTES" do
-    expect(CbvApplicant.valid_attributes_for_agency(agency_id)).to be_present
-  end
-end
-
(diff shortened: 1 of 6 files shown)


app/spec/models/cbv_applicant_spec.rb
```diff
@@ -1,12 +1,14 @@
 require 'rails_helper'
 
-RSpec.shared_examples "a CbvApplicant subclass" do |agency_id|
-  it "has a list of VALID_ATTRIBUTES" do
-    expect(CbvApplicant.valid_attributes_for_agency(agency_id)).to be_present
+RSpec.describe CbvApplicant, type: :model do
```
GeorgeCodes19, line 3: Thanks for introducing this. It does make sense vs having to include the shared example in all of these tests.

GeorgeCodes19 (approved):
Thanks for introducing this. It does make sense vs having to include the shared example in all of these tests.

PR #570, opened 2025-04-16
Remove `gigs_synced_at` column

tdooner:
## Ticket

N/A - Addresses dev UX  

## Context for reviewers

This column was mistakenly added in 7dfb18a5, and never populated. Since
the `db/schema.rb` file was not updated to include the column in the
original commit, we've all been getting diffs in our `db/schema.rb` ever
since.

A column created via migrations, but not existing in the schema.rb, will
result in there being a different set of columns running in places where
the schema is never recreated (demo, production) versus places that
reinitialize the schema from the schema.rb file (tests, local dev).

Let's just remove this unnecessary column so our schemas don't differ.

## Acceptance testing


iannorriswork:
LGTM

millerti (commented):
Lookin good. I know this is merged already but figure I should cross my ts.

PR #572, opened 2025-04-17
FFS-2724: Use custom argyle flowId to allow multiple employers

tdooner:
## Ticket

Resolves [FFS-2724](https://jiraent.cms.gov/browse/FFS-2724).

## Changes

**FFS-2724: Use custom argyle flowId to allow multiple employers**
> Argyle was restricting users to only linking one account because any
> subsequent accounts would open an uncloseable "Your Connections" screen
> before allowing future subsequent account linking.

> This page can be disabled in Argyle settings by defining a custom
> "flow". We just have to pass this ID into the modal initialization.

> Tested locally and it worked like a charm.

## Acceptance testing

### DIFF
--- a/app/app/services/aggregators/sdk/argyle_service.rb
+++ b/app/app/services/aggregators/sdk/argyle_service.rb
@@ -21,6 +21,9 @@ module Aggregators::Sdk
       }
     }
 
+    # See: https://console.argyle.com/flows
+    FLOW_ID = "BXSHLUUJ"
+
     ITEMS_ENDPOINT = "items"
     PAYSTUBS_ENDPOINT = "paystubs"
     IDENTITIES_ENDPOINT = "identities"
--- a/app/spec/controllers/cbv/payment_details_controller_spec.rb
+++ b/app/spec/controllers/cbv/payment_details_controller_spec.rb
@@ -40,6 +40,7 @@ RSpec.describe Cbv::PaymentDetailsController do
       pinwheel_stub_request_end_user_paystubs_response
       pinwheel_stub_request_income_metadata_response if supported_jobs.include?("income")
       pinwheel_stub_request_employment_info_response
+      pinwheel_stub_request_shifts_response
     end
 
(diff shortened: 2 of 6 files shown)


app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -40,6 +40,7 @@
       pinwheel_stub_request_end_user_paystubs_response
       pinwheel_stub_request_income_metadata_response if supported_jobs.include?("income")
       pinwheel_stub_request_employment_info_response
+      pinwheel_stub_request_shifts_response
```
tdooner, line 43: This is unrelated but needed to address a test failure on `main`.

app/app/services/aggregators/sdk/argyle_service.rb
```diff
@@ -21,6 +21,9 @@ class ArgyleService
       }
     }
 
+    # See: https://console.argyle.com/flows
+    FLOW_ID = "BXSHLUUJ"
```
iannorriswork, line 25: so is that a custom flow specific to something we've made?

wonder how we can document anything we've set up on the argyle side.

app/spec/controllers/cbv/payment_details_controller_spec.rb
millerti: I've also added this on 2652 so fyi whoever's first will have a conflict

millerti (approved):
Works for me too! Nice.

app/app/services/aggregators/sdk/argyle_service.rb
tdooner: Yeah, I made it in the Argyle console. Surprisingly, their flows seem to not be scoped by environment, so it's currently going to be the same in both environments. I wish I could "infrastructure as code" it or document it in some sane fashion, but I think I'd need a suggestion as to how to do so usefully, since I can't think of a great way myself.

PR #573, opened 2025-04-17
FFS-2777: Fix CompositeReport on /summary and PDF rendering

tdooner:
## Ticket

Resolves [FFS-2777](https://jiraent.cms.gov/browse/FFS-2777).

## Changes

Fixes a couple bugs in the CompositeReport:
* Forgetting a `.new`
* Rewriting `earlier_date`/`later_date` so they don't `Date.parse` an existing date
* Writing specs, including a test factory for Argyle bob.

## Acceptance testing

### DIFF
--- a/app/app/services/aggregators/aggregator_reports/composite_report.rb
+++ b/app/app/services/aggregators/aggregator_reports/composite_report.rb
@@ -10,6 +10,7 @@ module Aggregators::AggregatorReports
     end
 
     private
+
     def combine_reports(aggregator_reports)
       aggregator_reports.each do |aggregator_report|
         @payroll_accounts += aggregator_report.payroll_accounts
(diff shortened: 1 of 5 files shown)


app/app/services/aggregators/aggregator_reports/composite_report.rb
```diff
@@ -25,12 +25,8 @@ def combine_reports(aggregator_reports)
     end
 
     def earlier_date(a, b)
-      if a.present? && b.present?
-        if Date.parse(a) < Date.parse(b)
-          a
-        else
-          b
-        end
+      if a.is_a?(Date) && b.is_a?(Date)
```
tdooner, line 28: Not sure this is correct, but I do think the dates are always passed in as Date objects and never strings.

iannorriswork:
Looks good, though I am concerned about how we missed this the first time. 🤦

iannorriswork (approved):

tdooner:
@iannorriswork Agreed.

PR #576, opened 2025-04-18
[FFS-2750] add terraform to create solidqueue app workers

iannorriswork:
given we dont want them in a load balancer, it seemed copying the task definition was least worst for now.

Might be worth thinking about how to modularize the task if it needs to be modified frequently.

## Ticket

Resolves [FFS-2750](https://jiraent.cms.gov/browse/FFS-2750).

## Acceptance testing

- [X ] Acceptance testing prior to merge
  * Confirm the terraform applied is spawning up the correct workers and that setting the environment variable leads to the CBV flows still performing correctly.

### DIFF
--- a/infra/modules/service/main.tf
+++ b/infra/modules/service/main.tf
@@ -38,6 +38,104 @@ locals {
 # Service Execution
 #-------------------
 
+resource "aws_ecs_service" "solid_queue" {
+  name                   = "${var.service_name}-solid_queue"
+  cluster                = aws_ecs_cluster.cluster.arn
+  launch_type            = "FARGATE"
+  task_definition        = aws_ecs_task_definition.app.arn
+  desired_count          = var.desired_solidqueue_instance_count
+  enable_execute_command = var.enable_command_execution ? true : null
+
+  lifecycle {
+    ignore_changes = [desired_count]
+  }
+
+  network_configuration {
+    assign_public_ip = false
+    subnets          = var.private_subnet_ids
+    security_groups  = [aws_security_group.app.id] # Or another SG if needed
+  }
+
+}
+
+
+
+resource "aws_ecs_task_definition" "solid_queue" {
+  family             = var.service_name
+  execution_role_arn = aws_iam_role.task_executor.arn
+  task_role_arn      = aws_iam_role.app_service.arn
+
+  container_definitions = jsonencode([
+    {
+      name        = local.container_name,
+      image       = local.image_url,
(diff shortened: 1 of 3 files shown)


daphnegold (commented):
Looks like infra checks are failing. Might be worth it to poke and see if we have a reusable solution from platform, and I agree on modularized.

infra/modules/service/main.tf
```diff
@@ -38,6 +38,104 @@ locals {
...
+      # to use AWS's defaults, otherwise the terraform plan will show a diff
+      # that will force a replacement of the task definition
+      healthCheck = {
+        interval = 30,
+        retries  = 3,
+        timeout  = 5,
+        command = ["CMD-SHELL",
+          "curl --fail http://localhost:${var.container_port}/health"
+        ]
+      },
```
daphnegold, line 92: I think it might be prudent for the worker to have its own health check

infra/modules/service/main.tf
```diff
@@ -38,6 +38,104 @@ locals {
...
+      command     = ["bin/rails", "solid_queue:start"],
+      # TODO: Reenable readonlyRootFilesystem when we can have it behave
+      # consistently in dev (demo) and production.
+      # readonlyRootFilesystem = !var.enable_command_execution,
+      readonlyRootFilesystem = false,
+
+      # Need to define all parameters in the healthCheck block even if we want
+      # to use AWS's defaults, otherwise the terraform plan will show a diff
+      # that will force a replacement of the task definition
+      healthCheck = {
```
daphnegold, line 85: Will this health check effectively for the worker since it's hitting the main app endpoint? Should we check something solid queue specific?

iannorriswork: good idea. let me think about what we should hit instead!

infra/modules/service/main.tf
```diff
@@ -38,6 +38,104 @@ locals {
...
+
+  volume {
+    name = "${var.service_name}-tmp"
+  }
+
+  requires_compatibilities = ["FARGATE"]
+
+  # Reference https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-networking.html
+  network_mode = "awsvpc"
+}
```
tdooner, line 137: If we're messing with the `infra/modules` directory, we run a high risk of merge conflicts with future infrastructure upgrades. Perhaps breaking this into its own file, e.g. `infra/modules/service/async_workers.tf` would be a way to avoid that.

PR #579, opened 2025-04-18
FFS-2746: tweak pinwheel account syncher to be more resilient

iannorriswork:
## Ticket

Resolves [FFS-2746](https://jiraent.cms.gov/browse/FFS-XXXX).

 * Fail to sync an account in a cbv flow then sync an account. Now data will show instead of error page.

### DIFF
--- a/app/app/services/aggregators/aggregator_reports/aggregator_report.rb
+++ b/app/app/services/aggregators/aggregator_reports/aggregator_report.rb
@@ -24,20 +24,20 @@ module Aggregators::AggregatorReports
       @has_fetched
     end
 
-    def payroll_fetchable_accounts
+    def fetchable_payroll_accounts
       @payroll_accounts.select do |payroll_account|
-        payroll_account.has_fully_synced?
+        payroll_account.has_fully_synced? && payroll_account.suppor
       end
     end
 
     def is_ready_to_fetch?
-      payroll_fetchable_accounts.present?
+      fetchable_payroll_accounts.present?
     end
 
     def fetch_report_data
       begin
         all_successful = true
-        payroll_fetchable_accounts.each do |payroll_account|
+        fetchable_payroll_accounts.each do |payroll_account|
           fetch_report_data_for_account(payroll_account)
         end
       rescue StandardError => e
--- a/app/app/services/aggregators/aggregator_reports/aggregator_report.rb
+++ b/app/app/services/aggregators/aggregator_reports/aggregator_report.rb
@@ -25,7 +25,9 @@ module Aggregators::AggregatorReports
     end
 
     def is_ready_to_fetch?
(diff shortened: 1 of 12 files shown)


app/app/services/aggregators/aggregator_reports/pinwheel_report.rb
```diff
@@ -3,6 +3,7 @@ class PinwheelReport < AggregatorReport
     include ActiveModel::Validations
 
     attr_accessor :pinwheel_service
+    attr_reader :employments
```
tdooner, line 6: This shouldn't be necessary as the base class exposes `employments` via attr_accessor already.

app/spec/factories/payroll_account.rb
```diff
@@ -15,6 +15,19 @@
...
+      after(:build) do |payroll_account, evaluator|
+        event_name = PayrollAccount::Pinwheel::JOBS_TO_WEBHOOK_EVENTS["identity"]
+
+        payroll_account.webhook_events << build(
+          :webhook_event,
+          event_name: event_name,
+          event_outcome: "success"
+        )
+      end
+    end
```
tdooner, line 29: Is this representative of the actual data condition that was causing this error to occur - i.e. that we only got back a single webhook from pinwheel? I would expect that we actually did get back the other webhooks, but they were errors. At least that was the case for the variant of this bug that I saw most recently - the income webhook had failed to fetch with Pinwheel.

If that's the case, a more representative test setup would be:

```ruby
create(:payroll_account, :pinwheel_fully_synced, errored_jobs: %i[income paystubs])
```

or something similar.

app/app/services/aggregators/aggregator_reports/aggregator_report.rb
```diff
@@ -24,16 +24,20 @@ def has_fetched?
       @has_fetched
     end
 
-    def is_ready_to_fetch?
-      @payroll_accounts.all? do |payroll_account|
+    def payroll_fetchable_accounts
+      @payroll_accounts.select do |payroll_account|
```
tdooner, line 28: I would prefer `fetchable_payroll_accounts` here to keep the model name together.

tdooner (commented):
What is the user-facing behavior when a report is not fetchable? Does it still show on the /summary page? I think we will want those accounts to be hidden (see [this slack convo](https://cmsgov.slack.com/archives/C08BBUALWQN/p1744678138062599)).

app/app/services/aggregators/aggregator_reports/pinwheel_report.rb
iannorriswork: ah yup, i made too many assumptions from neighbor specs!

tdooner (approved):

PR #581, opened 2025-04-18
FFS-2779: Event isn't logged when switching languages

daphnegold:
## Ticket

Resolves [FFS-2779](https://jiraent.cms.gov/browse/FFS-2779).

## Changes
**Adds:**
- Stimulus controller for language
- Mixpanel tracking when users switch languages in the header 
- Locale in all Mixpanel events, even without locale in url
- Events testing

**Testing:**
- Visit CBV entry
- Swap between Spanish and English
- Verify "UserManuallySwitchedLanguage" appears in log stream 
- Check all Mixpanel events have proper locale

## Acceptance testing

**Before**
  Screenshot 2025-04-18 at 2 07 25 PM
**After**
Screenshot 2025-04-18 at 2 08 30 PM

Screenshot 2025-04-18 at 2 04 32 PM

### DIFF
--- a/app/app/controllers/api/user_events_controller.rb
+++ b/app/app/controllers/api/user_events_controller.rb
@@ -30,6 +30,7 @@ class Api::UserEventsController < ApplicationController
     ApplicantEncounteredArgyleInvalidAuthLoginError
     ApplicantEncounteredArgyleMfaCanceledLoginError
     ApplicantUpdatedArgyleSearchTerm
+    UserManuallySwitchedLanguage
   ]
 
   # Maps aggregator event names (keys) to new Mixpanel event names (values) we're using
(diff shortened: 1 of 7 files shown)


app/app/controllers/api/user_events_controller.rb
```diff
@@ -30,6 +30,7 @@ class Api::UserEventsController < ApplicationController
     ApplicantEncounteredArgyleInvalidAuthLoginError
     ApplicantEncounteredArgyleMfaCanceledLoginError
     ApplicantUpdatedArgyleSearchTerm
+    UserManuallySwitchedLanguage
```
millerti, line 33: Applicant instead of User, I think, unless we're worried a caseworker would trigger this. But I don't think we translate cw workflows.

GeorgeCodes19 (changes requested):
Looks good with the exception of the naming convention that Tim pointed out. I'll approve after that gets changed.

app/app/controllers/api/user_events_controller.rb
daphnegold: 'User' is specified by product as the expected event in the dashboard. I thought about adding it to the naming map below or should we just outright change it?

GeorgeCodes19 (approved):

PR #582, opened 2025-04-18
FFS-2784: Update /synchronizations to only wait for a partial sync of 90 days

tdooner:
## Ticket

Resolves [FFS-2784](https://jiraent.cms.gov/browse/FFS-2784).

## Changes

**Allow passing in `config` when creating Argyle webhook subscriber**
> The config hash is useful for providing the `days_synced` argument to be
> able to register partially-synced events. This will be useful later for
> us for real. For now, this was just a testing thing that helped me out.

**Wait for Argyle `partially_synced` events (rather than `fully_synced`)**
> The Argyle sync anecdotally takes a very long time, and especially for
> gig accounts that have been open for a while (even if there isn't
> activity going back very far). We want to take advantage of Argyle's
> `paystubs.partially_synced` and `gigs.partially_synced` to allow the
> user to proceed to the /payment-details page even if the account may
> still technically be syncing in the background.

> This raises some complexity: the `partially_synced` webhooks must be
> registered with a "config" parameter specifying the definition of
> "partial" - i.e. `{ "days_synced": 90 }` for the partially_synced
> webhook to trigger after >= 90 days have synced.

> So, we need to update the ArgyleWebhooksManager to manage *two* webhook
> subscriptions now - a "partial" webhook for these new events, and a
> "non-partial" webhook for the others.

## Acceptance testing

### DIFF
--- a/app/app/services/aggregators/sdk/argyle_service.rb
+++ b/app/app/services/aggregators/sdk/argyle_service.rb
@@ -71,11 +71,12 @@ module Aggregators::Sdk
       @http.get(build_url(WEBHOOKS_ENDPOINT)).body
     end
 
-    def create_webhook_subscription(events, url, name)
+    def create_webhook_subscription(events, url, name, config = {})
       payload = {
         events: events,
         name: name,
         url: url,
+        config: config.presence,
         secret: @webhook_secret
       }
 
--- a/app/app/controllers/webhooks/argyle/events_controller.rb
+++ b/app/app/controllers/webhooks/argyle/events_controller.rb
@@ -74,11 +74,13 @@ class Webhooks::Argyle::EventsController < ApplicationController
   def authorize_webhook
     argyle_service = @cbv_flow.present? ? argyle_for(@cbv_flow) : ArgyleService.new("sandbox")
 
-    unless Aggregators::Webhooks::Argyle.get_webhook_events.include?(params["event"])
+    unless Aggregators::Webhooks::Argyle.get_webhook_events(type: :all).include?(params["event"])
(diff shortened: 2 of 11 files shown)


app/app/services/aggregators/sdk/argyle_service.rb
```diff
@@ -71,11 +71,12 @@ def get_webhook_subscriptions
       @http.get(build_url(WEBHOOKS_ENDPOINT)).body
     end
 
-    def create_webhook_subscription(events, url, name)
+    def create_webhook_subscription(events, url, name, config = {})
       payload = {
         events: events,
         name: name,
         url: url,
+        config: config.presence,
```
jeffcatania-usds, line 79: this is strange to me.  We have a config hash, but it only has one potential key. That key is then provided as a payload under "config".  Presence should either be a named parameter or the whole object should be passed in.

jeffcatania-usds (approved):
LGTM

app/app/services/aggregators/sdk/argyle_service.rb
tdooner: Well the whole config object _is_ passed in, it would be passed in as `{ "days_synced": 90 }` or similar.

The `presence` method is to turn an empty `config` hash (i.e. `{}`) into `nil`. This is an odd quirk of the Argyle API: that an empty hash is not valid but `nil` is. So if we're not going to be passing the `days_synced` argument, we need to be passing `config: nil`.

tdooner:
Acceptance tested with Eryn. ✅

PR #583, opened 2025-04-21
FFS-2782: round hours to the nearest tenth on the screen in cases where argyle might report as many decimals as it likes

iannorriswork:
## Acceptance testing

  * Generate report with argyle test1 user, see no hours with more than one decimal point.


app/app/helpers/money_helper.rb
```diff
@@ -0,0 +1,6 @@
+module MoneyHelper
+  def format_hours(hour)
+    return hour unless Float(hour, exception: false).present?
+    hour.to_f.round(1)
+  end
+end
```
tdooner, line 6: Would prefer a different option rather than making a new helper module... two ideas come to mind:
1. Put this in the existing (poorly factored, poorly named) `ViewHelper` module, which already has other formatting methods used during report rendering.
2. Refactor `ViewHelper` into a new helper, including this method and any others necessary for presenting the aggregator data. We [used to have a ReportsHelper module](https://github.com/DSACMS/iv-cbv-payroll/commit/effd2c6a70d97b68ee5db5d9fcd34b299ee36c7a#diff-b260f938d1ccf37fc6df34570922f7ebe484ac9388602313f3655e2ededd1f59) that was removed for being poorly organized - as it included both formatting logic and also the report fetching logic. It would make sene to me to bring it back as a purely formatting module with methods like this, and some of the other `format_*` methods in ViewHelper.

app/app/mailers/application_mailer.rb
```diff
@@ -2,6 +2,7 @@ class ApplicationMailer < ActionMailer::Base
   default from: "noreply@mail.#{ENV["DOMAIN_NAME"]}"
   layout "mailer"
   after_deliver :track_delivery
+  helper MoneyHelper
```
tdooner, line 5: Is this necessary? I would have expected this in ApplicationController, not ApplicationMailer...?

app/app/helpers/money_helper.rb
```diff
@@ -0,0 +1,6 @@
+module MoneyHelper
```
tdooner, line 1: `MoneyHelper` is not a great name since this method has nothing to do with money.

tdooner (changes requested):
In theory the code looks good, but I don't love the location of the code -- I'd like to minimize the number of helpers we have that aren't related directly to a controller. Let's just put this in `ViewHelper`?

app/app/helpers/money_helper.rb
iannorriswork: interesting, in my head, the rounding associated with this is very related to the rounding we do with money--it's hours worked to dollars. Compromise in ReportViewHelper?

app/app/mailers/application_mailer.rb
iannorriswork: https://stackoverflow.com/questions/3681607/where-do-i-put-helper-methods-for-actionmailer-views

app/app/helpers/money_helper.rb
tdooner: Cool, thanks, ReportViewHelper is what I was trying to suggest with my second idea.

app/app/views/cbv/submits/show.pdf.erb
```diff
@@ -125,9 +125,9 @@
         <%= table.with_data_point(:pay_period_with_frequency, paystub.start, paystub.end, t("cbv.payment_details.show.frequency_unknown"), highlight: is_caseworker) %>
       <% end %>
       <%= table.with_data_point(:pay_gross, paystub.gross_pay_amount, highlight: is_caseworker) %>
-      <%= table.with_data_point(:number_of_hours_worked, paystub.hours, highlight: is_caseworker) %>
+      <%= table.with_data_point(:number_of_hours_worked, format_hours(paystub.hours), highlight: is_caseworker) %>
       <% paystub.hours_by_earning_category.each do |category, total_hours| %>
-        <%= table.with_data_point(:earnings_entry, category, total_hours) %>
+        <%= table.with_data_point(:earnings_entry, category, format_hours(total_hours)) %>
```
tdooner, line 130: One last nit. Can we put the calls to `format_hours` into the `PinwheelDataPointComponent` methods for `earnings_entry`, `number_of_hours_worked`, etc., instead of being here? That is where all the other similar methods are called. After that, Lgtm.

tdooner (approved):

PR #584, opened 2025-04-21
 FFS-2750: add ability to see the dashboard across environments

iannorriswork:
by default, the jobs dashboard engine we use leverages rails credentials. This trick lets us hook into the configuration directly instead so we can inject environment variables, since I don't think we have time right now to refactor into rails creds.

## Acceptance testing

set MISSION_CONTROL_USER and MISSION_CONTROL_PASSWORD environment variables
run locally (bin/start_dev)
visit localhost:3000/jobs, putting in user/password you defined earlier.
if the page loads with a job dashboard, success!

### DIFF
--- a/app/config/environment.rb
+++ b/app/config/environment.rb
@@ -1,5 +1,15 @@
 # Load the Rails application.
 require_relative "application"
 
+
 # Initialize the Rails application.
 Rails.application.initialize!
+
+Rails.application.config.after_initialize do
+  # Your code here
+  # You can call services, start background tasks, preload data, etc.
+  # if the password is not defined, Mission Control will 401
+  # TODO once we're on rails credentials, remove this and replace with rails creds
+  MissionControl::Jobs.http_basic_auth_user = ENV["MISSION_CONTROL_USER"]
+  MissionControl::Jobs.http_basic_auth_password = ENV["MISSION_CONTROL_PASSWORD"]
+end
--- a/app/config/environment.rb
+++ b/app/config/environment.rb
@@ -4,12 +4,3 @@ require_relative "application"
 
 # Initialize the Rails application.
 Rails.application.initialize!
-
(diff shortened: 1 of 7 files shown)


app/config/environment.rb
```diff
@@ -1,5 +1,15 @@
...
 Rails.application.initialize!
+
+Rails.application.config.after_initialize do
+  # Your code here
+  # You can call services, start background tasks, preload data, etc.
+  # if the password is not defined, Mission Control will 401
+  # TODO once we're on rails credentials, remove this and replace with rails creds
+  MissionControl::Jobs.http_basic_auth_user = ENV["MISSION_CONTROL_USER"]
+  MissionControl::Jobs.http_basic_auth_password = ENV["MISSION_CONTROL_PASSWORD"]
+end
```
tdooner, line 15: The comment includes extraneous text. Also, let's put this in its own initializer file.

As for basic auth, in general I would prefer we use Devise so that we only have a single authentication regime. MissionControl seems designed to handle that by providing a [`base_controller_class`](https://github.com/rails/mission_control-jobs#custom-authentication) config... but, that said, I'm ok with it for now, since are likely to remove Devise in the future and we don't have any concept of an "admin" user or anything like that, so I suppose Basic Auth is fine for now.

app/config/environments/development.rb
```diff
@@ -11,8 +11,6 @@
   # it changes. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
-  config.mission_control.jobs.http_basic_auth_enabled = false
```
tdooner, line 14: Does removing this mean we will need default values for the basic auth env variables in dev, in order for the dashboard to load? If so, please add them to the checked-in .env file.

tdooner (approved):

PR #585, opened 2025-04-21
FFS-2693 - Enable client agency resolution via request Host

GeorgeCodes19:
## Ticket

Resolves [FFS-2693](https://jiraent.cms.gov/browse/FFS-2693).

## Acceptance testing

### DIFF
--- a/app/app/controllers/application_controller.rb
+++ b/app/app/controllers/application_controller.rb
@@ -82,6 +82,13 @@ class ApplicationController < ActionController::Base
         return agency_id
       end
     end
+
+    raise "Unknown domain #{request.host}"
+  rescue => ex
+    raise ex unless Rails.env.production?
+
+    Rails.logger.error ex.to_s
+    nil
   end
 
   protected
--- a/app/app/controllers/application_controller.rb
+++ b/app/app/controllers/application_controller.rb
@@ -76,19 +76,10 @@ class ApplicationController < ActionController::Base
   def detect_client_agency_from_domain
     return nil unless request.host.present?
 
-    agency_config.client_agency_ids.each do |agency_id|
+    agency_config.client_agency_ids.find do |agency_id|
       agency = agency_config[agency_id]
-      if [ agency.agency_demo_domain, agency.agency_production_domain ].compact.include?(request.host)
-        return agency_id
-      end
+      [ agency.agency_demo_domain, agency.agency_production_domain ].compact.include?(request.host)
     end
-
-    raise "Unknown domain #{request.host}"
-  rescue => ex
-    raise ex unless Rails.env.production?
-
-    Rails.logger.error ex.to_s
-    nil
   end
 
   protected
--- a/app/app/controllers/application_controller.rb
+++ b/app/app/controllers/application_controller.rb
@@ -46,7 +46,21 @@ class ApplicationController < ActionController::Base
   end
 
   def current_agency
-    @current_agency ||= agency_config[params[:client_agency_id]]
+    # First try to get the agency from the client_agency_id parameter
+    return @current_agency if @current_agency.present?
+
+    if params[:client_agency_id].present?
+      @current_agency = agency_config[params[:client_agency_id]]
+      return @current_agency if @current_agency.present?
+    end
+
+    # If not found from params, try to detect from domain
+    client_agency_id_from_domain = detect_client_agency_from_domain
+    if client_agency_id_from_domain.present?
+      @current_agency = agency_config[client_agency_id_from_domain]
+    end
+
+    @current_agency
   end
 
   def enable_mini_profiler_in_demo
@@ -57,6 +71,17 @@ class ApplicationController < ActionController::Base
 
   def demo_mode?
     ENV["DOMAIN_NAME"] == "verify-demo.navapbc.cloud"
+  end
+
+  def detect_client_agency_from_domain
+    return nil unless request.host.present?
+
+    agency_config.client_agency_ids.each do |agency_id|
+      agency = agency_config[agency_id]
+      if [ agency.agency_demo_domain, agency.agency_production_domain ].compact.include?(request.host)
+        return agency_id
+      end
+    end
   end
 
   protected
--- a/app/config/application.rb
+++ b/app/config/application.rb
@@ -46,15 +46,13 @@ module IvCbvPayroll
     config.client_agencies = ClientAgencyConfig.new(Rails.root.join("config", "client-agency-config.yml"))
 
     # Configure allowed hosts
-    if ENV["DOMAIN_NAME"].present?
-      config.hosts << ENV["DOMAIN_NAME"]
-    end
+    config.hosts << ENV["DOMAIN_NAME"]
 
-    # Add alternate domain names if specified
-    if ENV["ALTERNATE_DOMAIN_NAMES"].present?
-      ENV["ALTERNATE_DOMAIN_NAMES"].split(",").map(&:strip).each do |domain|
-        config.hosts << domain if domain.present?
-      end
+    # Configure allowed hosts inferred from the client-agency-config.yml file
+    config.client_agencies.client_agency_ids.each do |agency_id|
+      agency = config.client_agencies[agency_id]
+      config.hosts << agency.agency_demo_domain
+      config.hosts << agency.agency_production_domain
     end
 
     # Health check endpoints should be accessible from any host
--- a/app/config/application.rb
+++ b/app/config/application.rb
@@ -45,6 +45,23 @@ module IvCbvPayroll
     config.autoload_paths += %W[#{config.root}/app/controllers/concerns]
     config.client_agencies = ClientAgencyConfig.new(Rails.root.join("config", "client-agency-config.yml"))
 
+    # Configure allowed hosts
+    if ENV["DOMAIN_NAME"].present?
+      config.hosts << ENV["DOMAIN_NAME"]
+    end
+
+    # Add alternate domain names if specified
+    if ENV["ALTERNATE_DOMAIN_NAMES"].present?
+      ENV["ALTERNATE_DOMAIN_NAMES"].split(",").map(&:strip).each do |domain|
+        config.hosts << domain if domain.present?
+      end
+    end
+
+    # Health check endpoints should be accessible from any host
+    config.host_authorization = {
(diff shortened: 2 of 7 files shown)


app/app/controllers/application_controller.rb
```diff
@@ -59,6 +73,24 @@ def demo_mode?
     ENV["DOMAIN_NAME"] == "verify-demo.navapbc.cloud"
   end
 
+  def detect_client_agency_from_domain
+    return nil unless request.host.present?
```
iannorriswork, line 77: could we extract this code into a service object to test, or otherwise not rely on sending messages to a private method?

or alternatively, test it through the intended behavior via a generic controller endpoint?

app/config/client-agency-config.yml
```diff
@@ -5,6 +5,8 @@
   caseworker_feedback_form: https://forms.office.com/r/B66WeqsBFE
   logo_path: hra_logo.png
   logo_square_path: hra_logo_square.png
+  agency_demo_domain: nyc.verify-demo.navapbc.cloud
```
GeorgeCodes19, line 8: I actually didn't think that this was initially required. It's a bit ancillary to the requested feature. I suppose we can also configure things so that the subdomain maps to the client agency id. i.e. `az_des.reportmyincome.org`,  but that seems like a lot going on and isn't fun to type.

app/config/environments/production.rb
```diff
@@ -38,10 +38,6 @@
   config.action_mailer.asset_host = "https://#{ENV["DOMAIN_NAME"]}"
 
   routes.default_url_options[:host] = ENV["DOMAIN_NAME"]
-  config.hosts << ENV["DOMAIN_NAME"]
```
GeorgeCodes19, line 41: This logic has been moved to [app/config/application.rb](https://github.com/DSACMS/iv-cbv-payroll/pull/585/files#diff-2de1bd08ded7571eb74064b1e9d582115a0d5c9693ef2bf943976ce833f43bbfR49) since it's applicable to multiple environments.

app/app/controllers/application_controller.rb
```diff
@@ -59,6 +73,24 @@ def demo_mode?
...
+
+    agency_config.client_agency_ids.each do |agency_id|
+      agency = agency_config[agency_id]
+      if [ agency.agency_demo_domain, agency.agency_production_domain ].compact.include?(request.host)
+        return agency_id
+      end
+    end
+
+    raise "Unknown domain #{request.host}"
```
GeorgeCodes19, line 86: I figured we should raise an error here since it's configuration related

app/config/application.rb
```diff
@@ -45,6 +45,23 @@ class Application < Rails::Application
...
     config.client_agencies = ClientAgencyConfig.new(Rails.root.join("config", "client-agency-config.yml"))
 
+    # Configure allowed hosts
+    if ENV["DOMAIN_NAME"].present?
+      config.hosts << ENV["DOMAIN_NAME"]
+    end
+
+    # Add alternate domain names if specified
+    if ENV["ALTERNATE_DOMAIN_NAMES"].present?
+      ENV["ALTERNATE_DOMAIN_NAMES"].split(",").map(&:strip).each do |domain|
```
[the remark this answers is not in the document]
GeorgeCodes19: This is a good point. There some seem to be some duplicative declarations going on here. I'm in favor of inferring the `config.hosts` from the agency configuration rather than an env var.

app/config/client-agency-config.yml
```diff
@@ -91,6 +97,8 @@
   # TODO[FFS-2689]: Add Louisiana DPH logo, these assets are not available yet.
   # logo_path: ldh_logo.png
   # logo_square_path: ldh_logo_square.png
+  agency_demo_domain: la.verify-demo.navapbc.cloud
```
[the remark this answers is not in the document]
GeorgeCodes19: You're right. The. proper subdomain is `la-verify-demo.navapbc.cloud`. Since it's already setup as a valid CNAME I'll stick with the fully hyphenated variant.

app/config/environments/test.rb
```diff
@@ -29,6 +29,11 @@
   config.cache_store = :null_store
   routes.default_url_options[:host] = ENV.fetch("DOMAIN_NAME", "localhost")
 
+  # Allow localhost for testing
+  config.hosts << "localhost"
+  config.hosts << "127.0.0.1"
+  config.hosts << "example.org"
```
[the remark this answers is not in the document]
GeorgeCodes19: This is no longer needed. I had run into some issues earlier on that I resolved and this is residual.

app/app/controllers/application_controller.rb
GeorgeCodes19: Do these tests align with what you're thinking? [https://github.com/DSACMS/iv-cbv-payroll/pull/585/files#diff-76a2f7bd084e2eeb53e1d06ea8579b95594724cebe8bb668f4fcf0c503b22210](https://github.com/DSACMS/iv-cbv-payroll/pull/585/files#diff-76a2f7bd084e2eeb53e1d06ea8579b95594724cebe8bb668f4fcf0c503b22210)

GeorgeCodes19: Yeah, good call. I'll create another test that invokes a controller rather than directly invoking the method.

app/config/environments/test.rb
```diff
@@ -29,6 +29,9 @@
   config.cache_store = :null_store
   routes.default_url_options[:host] = ENV.fetch("DOMAIN_NAME", "localhost")
 
+  # Allow the various hosts that tests are using
+  config.hosts << "www.example.com"
```
GeorgeCodes19, line 33: The `Pages` tests fail without this line... for some reason.

`
expected the response to have a success status code (2xx) but it was 403

  0) Pages GET /home returns http success
     Failure/Error: expect(response).to have_http_status(:success)
       expected the response to have a success status code (2xx) but it was 403
     # ./spec/requests/pages_spec.rb:7:in `block (3 levels) in <main>'

expecting <"home"> but rendering with <["rescues/blocked_host", "rescues/layout"]>

  0) Pages GET /home when the maintenance mode is disabled renders the correct path
     Failure/Error: expect(response).to render_template(:home)
       expecting <"home"> but rendering with <["rescues/blocked_host", "rescues/layout"]>
     # ./spec/requests/pages_spec.rb:30:in `block (4 levels) in <main>'
     # ./spec/support/test_helpers.rb:5:in `stub_environment_variable'
     # ./spec/requests/pages_spec.rb:24:in `block (4 levels) in <main>'

Expected response to be a <3XX: redirect>, but was a <403: Forbidden>

  0) Pages GET /home when the maintenance mode is enabled redirects to maintenance path
     Failure/Error: expect(response).to redirect_to(maintenance_path)
       Expected response to be a <3XX: redirect>, but was a <403: Forbidden>
     # ./spec/requests/pages_spec.rb:18:in `block (4 levels) in <main>'
     # ./spec/support/test_helpers.rb:5:in `stub_environment_variable'
     # ./spec/requests/pages_spec.rb:12:in `block (4 levels) in <main>'`

tdooner: Weird. I would guess there's some logic that's like "if anything is added to config.hosts, then enable the host header validation", and since we're always adding ENV['DOMAIN_NAME'] to config.hosts now, the validation is being triggered now when it wasn't before.

Given all that, this seems fine. There is probably a more optimal way to set this up in test, but I'm not sure it's worth the time figuring that out.

app/spec/controllers/application_controller_spec.rb
```diff
@@ -9,6 +9,11 @@ def test_action
         render plain: I18n.locale.to_s
       end
     end
+
+    def show
+      @agency = current_agency
+      render plain: @agency.id
+    end
```
tdooner, line 16: This should be defined as an anonymous controller in the spec, not in ApplicationController itself. Refer to: https://rspec.info/features/6-0/rspec-rails/controller-specs/anonymous-controller/

The `controller do` block actually goes in the spec (similar to a `before do` block) and defines a new controller that is a subclass of the controller being tested.

iannorriswork (commented):
Fix the tests and I think we'll be pretty good.

One route for those tests might be something like this. This depends on whether you want to mock the client or lean on the existing clients for test data---I could go half and half on that.

 before do
      agency_config_mock = instance_double(ClientAgencyConfig)
      allow(controller).to receive(:agency_config).and_return(agency_config_mock)
      allow(agency_config_mock).to receive(:[]).with("la_ldh").and_return(OpenStruct.new(id: "la_ldh", agency_demo_domain: "la.reportmyincome.org", agency_production_domain: "la_prod"))
      allow(agency_config_mock).to receive(:client_agency_ids).and_return(["la_ldh"])
    end

    controller do
      def index
        render plain: current_agency.id
      end
    end

    it "identifies the correct agency config based on the domain name" do
      request.host = "la.reportmyincome.org"
      get :index
      expect(response.body).to eq("la_ldh")
    end

GeorgeCodes19:
> Fix the tests and I think we'll be pretty good.
> 
> One route for those tests might be something like this. This depends on whether you want to mock the client or lean on the existing clients for test data---I could go half and half on that.
> 
> ```
>  before do
>       agency_config_mock = instance_double(ClientAgencyConfig)
>       allow(controller).to receive(:agency_config).and_return(agency_config_mock)
>       allow(agency_config_mock).to receive(:[]).with("la_ldh").and_return(OpenStruct.new(id: "la_ldh", agency_demo_domain: "la.reportmyincome.org", agency_production_domain: "la_prod"))
>       allow(agency_config_mock).to receive(:client_agency_ids).and_return(["la_ldh"])
>     end
> 
>     controller do
>       def index
>         render plain: current_agency.id
>       end
>     end
> 
>     it "identifies the correct agency config based on the domain name" do
>       request.host = "la.reportmyincome.org"
>       get :index
>       expect(response.body).to eq("la_ldh")
>     end
> ```

I actually think that we had this pattern at one point. We do have a this nifty utility [here](https://github.com/DSACMS/iv-cbv-payroll/blob/cc519f478025296078c748345380aeaf0640a559/app/spec/support/test_helpers.rb#L101)

app/spec/controllers/application_controller_spec.rb
tdooner: My bad - I misread the filename of this file - we are in the spec all along.

tdooner (approved):

PR #587, opened 2025-04-22
FFS-2789: Fix duplicate Argyle search results

tdooner:
## Ticket

Resolves [FFS-2789](https://jiraent.cms.gov/browse/FFS-2789).

## Changes

**FFS-2789: Use Argyle /employer-search endpoint for search**
> We previously used Argyle's /items endpoint for employer search, but
> this returns duplicate results (e.g. for "Allstate").

> Argyle customer support suggests using /employer-search to match the
> search logic on their end. This is a drop-in replacement because the
> endpoints take the same arguments and return the data in the same
> format.

## Context for reviewers

This fixes duplicate search results because the /employer-search endpoint is smarter and doesn't have duplicates.

## Acceptance testing


tdooner:
Just reviewed with Eryn - she acceptance tested it in my dev environment.

jeffcatania-usds (approved):
LGTM

PR #589, opened 2025-04-22
Fix reviewbot by removing conditional logic

tdooner:
## Ticket

N/A - Fixes regression.

## Changes

Remove the conditional logic. It's not working for some reason (it results in
the job always being skipped). I tried debugging it for 15 minutes, to no
avail, so let's just remove it.

This reverts commit f7cbef7c44e2b965b949d5a384bcacd108fb813d.

## Acceptance testing


tdooner:
Going to self-merge since it's just reverting my own thing.

PR #591, opened 2025-04-22
FFS-2729: re-link continue button to applicant info page in case they want to continue forward

iannorriswork:
Changes
Links the next link on the missing results page such that it would proceed to the application information page, which they will need to fill out if in the LA flow. The application controller already checks if it is unnecessary to be there, so won't hurt non-LA flows while avoiding too much spreading logic.

Context for reviewers
In the future, there might be need for different action text. For now, a talk with design says they're ok with this language for now.

Testing instructions
  * With an LA applicant, get one payroll account added, then search and fail to find a second payroll account. Click the button--> direct to the applicant information page.

## Acceptance testing

  * With an LA applicant, get one payroll account added, then search and fail to find a second payroll account. Click the button--> direct to the applicant information page.

### DIFF
--- a/app/app/controllers/cbv/applicant_informations_controller.rb
+++ b/app/app/controllers/cbv/applicant_informations_controller.rb
@@ -38,11 +38,7 @@ class Cbv::ApplicantInformationsController < Cbv::BaseController
   def redirect_when_info_present
     return if params[:force_show] == "true"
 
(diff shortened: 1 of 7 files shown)


jeffcatania-usds (commented):
Code looks good.  Please update description of PR to match our template.

PR #594, opened 2025-04-22
FFS-2754 Refactor to begin ability to remove maybe_later

iannorriswork:
## Changes

Refactored GenericEventTracker to be executeable within the context of an ActiveJob.

## Context for reviewers
 refactor to begin ability to remove maybe_later
    
    i suspect this might also grant some performance boosts by removing thread burden on our application servers.
    
    this is a refactor with the goal of first behaving exactly as it currently does--putting it into MaybeLater still, but in a way that ActiveJob understands it.
    
    Then if this refactor still behaves as expected, we can move to the next phase --> putting it into asynchronous jobs.
    
    GenericEventTracker itself wasnt very tested, but the controller specs poked at various aspects of it--but can add more unit test coverage of the job/tracker itself now that its responsibilities from the request cycle are way more obvious.

## Acceptance testing

  * See that mixpanel and newrelic events are still firing as expected.


app/lib/generic_event_tracker.rb
```diff
@@ -1,7 +1,26 @@
...
+  def track(event_type, request, attributes = {})
+    merged_attributes = attributes.with_defaults(prep_request_attributes(request))
+    if request.present?
+      request_data = { headers: { "User-Agent": request.headers["User-Agent"] }, remote_ip: request.remote_ip  }
+    else
+      request_data = nil
+    end
+    if ENV["ACTIVEJOB_ENABLED"] == "true"
+      EventTrackingJob.perform_later(event_type, request_data, merged_attributes)
+    elsif Rails.env.test?
```
iannorriswork, line 13: i know this is cursed, but following the instructions in MaybeLater were NOT successful in running tests.

app/app/jobs/event_tracking_job.rb
```diff
@@ -0,0 +1,30 @@
...
+      attributes[:device_name] = device_detector.device_name
+      attributes[:device_type] = device_detector.device_type
+      attributes[:browser] = device_detector.name
+    end
+
+    [ MixpanelEventTracker.new, NewRelicEventTracker.new ].each do |service|
+      begin
+        service.track(event_type, request, attributes)
+      rescue StandardError => e
+        Rails.logger.error "  Failed to track #{event_type} in #{service}: #{e.message}"
```
tdooner, line 26: Hmm, I'm wondering if we should re-raise these in production so we can have the events retry later via ActiveJob?

app/lib/generic_event_tracker.rb
tdooner: Not sure I understand what problem you're solving with the cursed code? We stub out MaybeLater in a couple places in the test suite, with:

```ruby
allow(MaybeLater).to receive(:run) do |&block|
  block.call
end
```

Is it just because there are so many places we'd have to do it as to make it impossible? If so, stubbing MaybeLater globally seems fine with me given that we're about to remove it.

app/lib/generic_event_tracker.rb
```diff
@@ -15,41 +34,6 @@ def self.for_request(request)
         user_agent: request.headers["User-Agent"]
```
tdooner, line 34: Hmm, this is probably out-of-scope for your commit, but, we track a `timestamp` on a number of our events (but probably not all of them). It may be nice for us to standardize the `timestamp` field into the GenericEventTracker and then make sure Mixpanel gets the timestamp in whatever field allows it to be used for backdating the event (in case the event was delayed in the queue). I'll make a ticket for it.

tdooner (approved):
Seems like a good start.

app/lib/generic_event_tracker.rb
iannorriswork: ah, i see. that's a fun way to stub it--> if that's the case, i'll just hoist that into a global and call it a day to make the application code at least uniform.

app/app/jobs/event_tracking_job.rb
iannorriswork: agreed, added

app/lib/generic_event_tracker.rb
iannorriswork: wasn't that hard to add, so i just did

PR #595, opened 2025-04-23
FFS-2780: Redirect to Synchronizations page when argyle encounters "account system_error" 

jeffcatania-usds:
## Ticket

Resolves [FFS-2780](https://jiraent.cms.gov/browse/FFS-2780).

## Changes

* When a user encounters the "accounts.updated" event with a payload that states "system error", the user will be redirected to the synchronizations failures page and fires "ApplicantEncounteredArgyleAccountSystemError" mixpanel event 
* Introduced new "-include-resource" Webhook event subscription to subscribe to accounts.updated event with resource payload so we can handle everything within the webhook response. 

## Notes to operations
We need to add the new "-include-resource" webhook subscription in demo and production. 
image

Or via code:
```
{
    "next": null,
    "previous": null,
    "results": [
        {
            "id": "205c027e-8a4d-4248-b395-d4f345dfd49f",
            "name": "demo-include-resource",
            "events": [
                "accounts.updated"
            ],
            "config": {
                "include_resource": true
            },
            "secret": "optional-secret",
            "url": "https://demo-url/webhook",
            "created_at": "2024-08-30T13:58:53.683338Z",
            "updated_at": "2024-10-07T15:20:49.926525Z",
            "last_sent_at": "2024-10-07T15:19:49.412642Z"
        }
    ]
}
```

## Bug report
To simulate error condition:

1. Simulate a sync failure in the Argyle flow using Ann Loanseeker test credentials.
2. User never passes the /synchronizations page to the /synchronizations_failure. Income, Employment, Paystubs & Personal Details spin indefinitely.

Sync Fails - Anne Loanseeker

Ann Loanseeker
username: ann_loanseeker
password: passgood
email: [ann_loanseeker@argyle.com](mailto:ann_loanseeker@argyle.com)
phone: (800)-900-0121
 Drivers License: D1230121
MFA Code: 9071

## Acceptance testing


app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -75,16 +75,34 @@ def authorize_webhook
...
+      cbv_applicant_id: @cbv_flow.cbv_applicant_id,
+      cbv_flow_id: @cbv_flow.id,
+      invitation_id: @cbv_flow.cbv_flow_invitation_id,
+      connection_status: connection_status,
+      argyle_error_code: error_code,
+      argyle_error_message: params.dig("data", "resource", "connection", "error_message"),
+      argyle_error_updated_at: params.dig("data", "resource", "connection", "updated_at")
+    })
+
+    payroll_account.update(identity_errored_at: Time.now)
```
tdooner, line 102: We should be trying to get away from these columns, instead relying on the `event_outcome` column of `WebhookEvent` for this. Can we update

```
Aggregators::Webhooks::Argyle.get_webhook_event_outcome(event_name)
```

to conditionally return `:error` for the failure cases?

app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -75,16 +75,34 @@ def authorize_webhook
...
     end
   end
 
+  def process_accounts_updated_event(webhook_event)
+    payroll_account = webhook_event.payroll_account
+    connection_status = params.dig("data", "resource", "connection", "status")
+    error_code = params.dig("data", "resource", "connection", "error_code")
+    return unless connection_status == "error" && error_code == "system_error"
+
+    event_logger.track("ApplicantEncounteredArgyleSyncError", request, {
```
jeffcatania-usds, line 92:
```suggestion
    event_logger.track("ApplicantEncounteredArgyleAccountSyncSystemError", request, {
```

jeffcatania-usds: Could use some suggestions on this. I lean toward pedantic, but it's quite a mouthful. Almost looks like a German word.

app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -75,16 +75,34 @@ def authorize_webhook
     argyle_service = @cbv_flow.present? ? argyle_for(@cbv_flow) : ArgyleService.new("sandbox")
 
     unless Aggregators::Webhooks::Argyle.get_webhook_events(type: :all).include?(params["event"])
-      Rails.logger.info "Ignoring unhandled webhook: #{params["event"]}"
       render json: { info: "Unhandled webhook" }, status: :ok
     end
 
     unless Aggregators::Webhooks::Argyle.verify_signature(request.headers["X-Argyle-Signature"], request.raw_post, argyle_service.webhook_secret)
-      Rails.logger.info "Ignoring webhook with invalid signature: #{params["event"]}"
```
tdooner, line 83: Not a fan of these? I added them because I found the status code to be a bit hard to understand (i.e. we return a 200 OK if the webhook is unknown, but a 401 Unauthorized if the signature is wrong)

app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -75,16 +75,34 @@ def authorize_webhook
...
+      cbv_flow_id: @cbv_flow.id,
+      invitation_id: @cbv_flow.cbv_flow_invitation_id,
+      connection_status: connection_status,
+      argyle_error_code: error_code,
+      argyle_error_message: params.dig("data", "resource", "connection", "error_message"),
+      argyle_error_updated_at: params.dig("data", "resource", "connection", "updated_at")
+    })
+
+    webhook_event.update(event_outcome: "error")
+    update_synchronization_page(payroll_account)
```
tdooner, line 103: Oh, hm,`update_synchronization_page` probably does nothing now that we're polling on /synchronizations instead of using action cable. I'll make a TODO on my list to clean this up.

app/app/models/payroll_account/argyle.rb
```diff
@@ -25,9 +25,20 @@ def job_succeeded?(job)
...
+    else
+      :succeeded
+    end
+  end
+
   def synchronization_status(job)
     if supported_jobs.exclude?(job)
       :unsupported
+    elsif job == "accounts"
+      synchronization_status_for_accounts_job
```
tdooner, line 41: I don't think you need this special handling? Since you added `jobs: %w[accounts]` I would expect that you could call `synchronization_status("accounts")` and it would be using the check below to determine failure? (this one:

```
    elsif find_webhook_event(self.class.event_for_job(job), "error").present?
      :failed
```
)

app/spec/factories/payroll_account.rb
```diff
@@ -56,5 +56,17 @@
...
+      argyle
+
+      after(:build) do |payroll_account, evaluator|
+        payroll_account.webhook_events << build(
+          :webhook_event,
+          event_name: "accounts.updated",
+          event_outcome:  "error"
+        )
+      end
+    end
```
tdooner, line 70: I believe this is duplicative of `argyle_fully_synced` with `errored_jobs: ["accounts"]`?

app/app/models/payroll_account/argyle.rb
jeffcatania-usds: the problem is that the status is set as :in_progress while it awaits for this event to resolve. So i had to short circuit.

app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -75,16 +75,34 @@ def authorize_webhook
     argyle_service = @cbv_flow.present? ? argyle_for(@cbv_flow) : ArgyleService.new("sandbox")
 
     unless Aggregators::Webhooks::Argyle.get_webhook_events(type: :all).include?(params["event"])
```
jeffcatania-usds, line 77:
```suggestion
    unless Aggregators::Webhooks::Argyle.get_webhook_events(type: :all).include?(params["event"])
          Rails.logger.info "Ignoring unhandled webhook: #{params["event"]}"
```

app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -75,16 +75,34 @@ def authorize_webhook
...
-      Rails.logger.info "Ignoring unhandled webhook: #{params["event"]}"
       render json: { info: "Unhandled webhook" }, status: :ok
     end
 
     unless Aggregators::Webhooks::Argyle.verify_signature(request.headers["X-Argyle-Signature"], request.raw_post, argyle_service.webhook_secret)
```
jeffcatania-usds, line 81:
```suggestion
    unless Aggregators::Webhooks::Argyle.verify_signature(request.headers["X-Argyle-Signature"], request.raw_post, argyle_service.webhook_secret)
          Rails.logger.info "Ignoring webhook with invalid signature: #{params["event"]}"
```

app/app/models/payroll_account/argyle.rb
tdooner: > the status is set as :in_progress while it awaits for this event to resolve

I'm still not sure I understand, since I would expect that to be the correct behavior (it should be `:in_progress` while waiting for the event to resolve), and for it to work fine with the logic you added in Cbv::SynchronizationsController#update.

app/spec/controllers/webhooks/argyle/events_controller_spec.rb
```diff
@@ -141,6 +142,7 @@ def process_webhook(event_type)
 
         # expect the PayrollAccount to be fully synced
         expect(payroll_account.has_fully_synced?).to be_truthy
+        expect(payroll_account.identity_errored_at).to be_nil
```
tdooner, line 145: You could remove this assertion since we're no longer using that column.

app/spec/factories/payroll_account.rb
jeffcatania-usds: If i test like that, i would need to remove the "accounts" job from "accounts.connected" event.  If I do this, it breaks tests in Argyle::EventsController happy path testing because the assumption for find_webhook_event(self.class.event_for_job(job)) expects a 1:1 relationship between jobs and events.

tdooner: Oh, I see, ok, I was getting `accounts.connected` confused with `accounts.updated`. Makes sense.

app/app/models/payroll_account/argyle.rb
tdooner: I suppose I am also getting `accounts.connected` and `accounts.updated` confused here, so I suppose I can see why we might need special logic here since the `event_for_job("accounts")` will be `accounts.connected`.

jeffcatania-usds: It's definitely a headache.  As per slack convo, i'd love to have this move back to a state on the payroll_account that could be set on fully_sync status.

PR #596, opened 2025-04-23
clean up some code that was unrelated to current work but seemed nice to clean up

iannorriswork:
## Changes
rename something to avoid thinking about it in terms of pinwheel for a non pinwheel specific component controller spec missing the controller name was making rubymine navigation sad

## Context for reviewers
was some small refactors i had the time to get around to

## Acceptance testing

- [ X] No acceptance testing needed


daphnegold (approved):
🚢

PR #597, opened 2025-04-23
 FFS-2754: finish extracting out maybe later

iannorriswork:
## Ticket

Resolves [FFS-2754](https://jiraent.cms.gov/browse/FFS-2754).

## Changes
this removes the usage of the environment variable for job enqueueing, which enables us to remove maybe later.

## Context for reviewers
this favors unit testing that event job tracking fires successfully as one set of tests, and the integration tests now don't check newrelic/mixpanel in favor of checking that the async job was fired.

To fix in a followup PR maybe: right now, if the job fails because it needed to connect to mixpanel or newrelic, the job will fail and retry the whole re-sync. at some point might be worth having a job fire off jobs? (mixpanel and newrelic seperately) [or rip out the newrelic events]

## Acceptance testing
 * go through the flow, make sure mixpanel events are still firing

### DIFF
--- a/app/spec/controllers/help_controller_spec.rb
+++ b/app/spec/controllers/help_controller_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe HelpController, type: :controller do
         valid_params[:client_agency_id] = cbv_flow.client_agency_id
       end
 
-      it "tracks events with both trackers" do
+      it "tracks events for help topic" do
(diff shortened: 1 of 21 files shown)


app/spec/controllers/help_controller_spec.rb
```diff
@@ -28,29 +18,9 @@
       end
 
       it "tracks events with both trackers" do
```
daphnegold, line 20: Nit: this one needs a name update maybe

daphnegold (approved):

PR #599, opened 2025-04-23
FFS-2798: Add terraform to support ReportMyIncome.org

tdooner:
## Ticket

Resolves [FFS-2798](https://jiraent.cms.gov/browse/FFS-2798).

## Changes

To support this new domain, this commit adds terraform to:
1. Replace the old domain (snap-income-pilot.com) with the new domain.
2. We intend to use the new domain name with subdomains for
   white-labeling. This commit adds an `additional_domains`
   configuration variable that gets passed into the service module in
   order to create the DNS records for the wildcard domain.
3. The wildcard domain will also need to be added to the TLS
   certificate. This commit adds a new `subject_alternative_names`
   configuration item to the `certificate_configs` variable that allows
   for specifying additional hosts that should also be included in the
   TLS certificate.

Finally, to remain backwards-compatible (at least temporarily) with
snap-income-pilot.com, this commit:

1. Adds a "deprecated_domains.tf" file in the services and network
   modules which are containers to hold the old resources for
   snap-income-pilot.com, to still be used by the ALB managed by the
   service module. We will have to perform some terraform state surgery
   to move the old TLS certificate (for snap-income-pilot.com) from
   terraform, so terraform doesn't clean it up while provisioning the
   ReportMyIncome.org resources.

## Context for reviewers

This is probably intertwined with deploying Multi-Aggregator to production, as
there are shared terraform issues that we need to resolve before the terraform
will be able to create the resources in this PR.

## Acceptance testing

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -42,3 +42,4 @@ brew "gpg"
 # linters / workflow tools
 brew "pre-commit"
 brew "shellcheck"
+brew "checkov"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,3 +47,15 @@ repos:
     entry: bin/run-linter --cd app -- npm run format:precommit
     language: node
     files: '^app/.*\.(js|ts)$'
+
+  - id: terraform_fmt
+    name: Terraform Fmt
+    entry: terraform fmt
+    language: system
+    files: 'infra/.*\.tf$'
+
+  - id: checkov
+    name: Checkov
+    entry: checkov --quiet --output sarif --framework terraform -f
+    language: system
+    files: 'infra/.*\.tf$'
--- a/infra/app/app-config/dev.tf
+++ b/infra/app/app-config/dev.tf
@@ -17,6 +17,10 @@ module "dev_config" {
   service_memory                 = 4096
   service_desired_instance_count = 3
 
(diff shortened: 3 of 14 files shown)


Brewfile
```diff
@@ -42,3 +42,4 @@ brew "gpg"
 # linters / workflow tools
 brew "pre-commit"
 brew "shellcheck"
+brew "checkov"
```
GeorgeCodes19, line 45: This is pretty cool. I had no idea it existed. Was this a suggestion from the platforms team or something that you've previously known about and decided to add?

tdooner: It's [already in use by CI](https://github.com/DSACMS/iv-cbv-payroll/blob/6731e1dc652f268b820f59929c58f4f55ef78518/.github/workflows/ci-infra.yml#L63-L81), I'm just adding it to pre-commit here. (Homebrew seems to be installing a newer version of Checkov than is used in CI, which will get interesting, but hopefully it works decently at spotting errors early.)

daphnegold (approved):
As we discussed offline, I think minimizing the need for state surgery as we have an allowance for downtime, and questioning whether it is necessary to maintain backwards compatibility with the domains (if they will ever get cleaned up), is necessary. Otherwise, looked it over and great work!

PR #602, opened 2025-04-24
FFS-2795: hours fallback logic

millerti:
## Changes
The calculation for hours now falls back to other computations if hours is null or 0.

## Context for reviewers
Jeff realized that his real production data was failing to validate. Check out the ticket for the culprit: he had a top-level hours value of 0 and some nuanced gross_pay_list data that did include a nonzero hours value. This ticket allows us to gracefully accept this kind of real production data.

Currently if hours is 0, we will pick the largest hours_by_earning_category value we find and use that instead. This may not be right long-term: we may want to sum those categories, for example. That is out of the scope of this ticket.

## Acceptance testing


app/app/services/aggregators/format_methods/argyle.rb
```diff
@@ -21,12 +21,21 @@ def self.format_date(date)
   end
 
   def self.format_currency(amount)
-    return unless amount
-    dollars, cents = amount.split(".").map(&:to_i)
+    return if amount.blank?
+    return unless amount.include?(".")
```
tdooner, line 25: Let's just put in a realistic-looking `AMOUNT` in Jeff's request spec so we don't have to actually make this method more permissive than necessary, unless you think it's good to do anyway. (I kind of err on the side of being conservative with not making these methods accept too much before we know we'll need to, but I suppose this is just a preference.)

app/spec/services/aggregators/response_objects/paystub_spec.rb
```diff
@@ -93,4 +93,53 @@
...
+          { "name" => "tax", "amount" => "600.90" },
+          { "name" => "insurance", "amount" => "120.34" }
+        ]
+      }
+    end
+
+    it 'creates a Paystub object from argyle response' do
+      paystub = described_class.from_argyle(argyle_response)
+
+      expect(paystub.hours).to eq(80)
```
millerti, line 142: I am worried that I should make sure to preserve some kind of formatting here. Should I massage this into "80.0000" and check for that instead?

app/app/services/aggregators/format_methods/argyle.rb
```diff
@@ -21,12 +21,21 @@ def self.format_date(date)
...
 
+    dollars, cents = amount.split(".").map(&:to_i)
     (dollars * 100) + cents
   end
 
+  def self.hours_computed(response_hours, response_gross_pay_list)
+    if response_hours.present? && response_hours.to_f > 0
+      response_hours
+    else
+      hours_by_earning_category(response_gross_pay_list)&.max_by { |_, value| value }&.[](1)
```
tdooner, line 35: `&.[](1)` could be more readable... Also, I think the Safe Navigation Operators should be unnecessary as the `hours_by_earning_category` should always return a hash. How about instead:

```suggestion
      hours_by_earning_category(response_gross_pay_list).map { |_category, hours| hours }.max
```

app/spec/services/aggregators/response_objects/paystub_spec.rb
```diff
@@ -93,4 +93,53 @@
       expect(paystub.earnings.first.name).to eq("Regular")
     end
   end
+
+
+  describe '.from_argyle with realistic USDS employee data structure' do
```
tdooner, line 98: nit: Idiomatic rspec test structure has a single top-level `describe` for every method being tested, and then `context`s within it for the different test cases. It would be a bit nicer to make this a `context` so it's clear it's also testing the same method.

tdooner (approved):

app/spec/services/aggregators/response_objects/paystub_spec.rb
tdooner: This seems fine to me. You could add an assertion that it `is_a?(Float)` I suppose if you really wanted to check the type. But, uh, I don't think we'll get much from that.

app/app/services/aggregators/format_methods/argyle.rb
millerti: You mean replace the amount: "AMOUNT" in the spec with something like amount: "0.0000"? I'm worried that "AMOUNT" is a real value that we can get back because I'm pretty sure this is what Jeff did get in production. And it made this validation crash / be invalid.

tdooner: I am almost positive that `AMOUNT` is Jeff censoring of the actual decimal.

millerti:
The calculation for hours now falls back to other computations if hours is null or 0.

PR #603, opened 2025-04-25
FFS-2796-1: Rename sync tracking methods in PayrollAccount

tdooner:
## Ticket

First half of [FFS-2796](https://jiraent.cms.gov/browse/FFS-2796). This is the refactoring part of the branch, upon which the actual fix is implemented.

## Changes

**Rename existing synchronization_status -> job_status**
> This is in preparation for adding a `synchronization_status` field
> to the PayrollAccount itself, which tracks the overall synchronization
> status (not just for a single job).

**Rename `#has_required_data?` -> `#necessary_jobs_succeeded?`**
> This name better reflects that the logic does not actually depend on the
> data that's present.

**Improve documentation of PayrollAccount base class**

**Rewrite #job_succeeded? to defer to #job_status(job)**
> Just some logic simplifications in these methods.

## Acceptance testing

### DIFF
--- a/app/app/controllers/cbv/synchronizations_controller.rb
+++ b/app/app/controllers/cbv/synchronizations_controller.rb
@@ -7,15 +7,15 @@ class Cbv::SynchronizationsController < Cbv::BaseController
   end
 
   def update
-    if @payroll_account&.synchronization_status("accounts") == :failed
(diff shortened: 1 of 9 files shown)


jeffcatania-usds (commented):
Approved. This PR is mostly name changes and the new method names make sense.

PR #604, opened 2025-04-25
FFS-2796-2: Store sync status as an enum on PayrollAccount

tdooner:
## Ticket

Resolves [FFS-2796](https://jiraent.cms.gov/browse/FFS-2796). Built atop #603.

## Changes

**Store `synchronization_status` as an enum on PayrollAccount**
> This tracks the overall result of the synchronization process (including
> validation of the data contained within the report.) This can be used to
> filter the list of accounts that are shown to the user and rendered in the
> report.

**Use enum prefix to dynamically define `sync_successful?` method**
> This allows us to replace `successfully_synced?` with
> `sync_successful?`, which is based on the enum's value.

## Acceptance testing


app/spec/controllers/cbv/submits_controller_spec.rb
```diff
@@ -84,9 +84,10 @@
         let(:supported_jobs) { %w[income paystubs employment] }
         let(:errored_jobs) { [ "employment" ] }
 
-        it "bounces to synchronization failures" do
+        it "renders pdf properly" do
           get :show, format: :pdf
-          expect(response).to redirect_to(cbv_flow_synchronization_failures_path)
+          expect(response).to be_successful
+          expect(response.header['Content-Type']).to include 'pdf'
```
tdooner, line 90: This is a little hard to explain. "employment" is not a "necessary job" for pinwheel (although its data is required to pass the useful report validations). So, this test, which relies only on the webhook success, will consider the report successful, but in reality, the `synchronization_status` would have been set to `failed` by the useful report validator.

Not sure what to do with this awkward disagreement of logic here, but I decided not to try and untangle it too much because this series of PRs is already a bit complex.

app/app/models/payroll_account.rb
```diff
@@ -12,6 +12,13 @@ def self.sti_class_for(type_name)
   belongs_to :cbv_flow
   has_many :webhook_events
 
+  enum :synchronization_status, {
+    unknown: "unknown",              # defines the method: sync_unknown?
```
jeffcatania-usds, line 16: what other conditions would cause an unknown sync status other than initializing without a state?  Should this be renamed "not_started" or something similar?

jeffcatania-usds (commented):
Looks good. Would like more definition around the "unknown" state.

app/app/models/payroll_account.rb
tdooner: Good question, thanks for drawing my attention to it. My intent was to make it the default so that the synchronization_status would be "unknown" unless otherwise updated by the webhooks/sync process.

However, as I was testing this now I realized that the `default` configuration needs to be on the database column, not here, so I'll push a quick fix for that onto this PR.

tdooner: So to directly answer your question, `unknown` should never actually be a synchronization_status, otherwise something has gone wrong.

PR #605, opened 2025-04-25
FFS-2688-add_jobs: implemented add_jobs copy changes

millerti:
## Changes
Implemented LA copy for /add_jobs
Removed ma/nyc copy for /add_jobs

## Testing instructions
2. Go through the workflow to /add_jobs
3. Verify the copy matches [the design](https://confluenceent.cms.gov/pages/viewpage.action?pageId=748887627)

## Acceptance testing

## Screenshot
image

### DIFF
--- a/app/config/locales/en.yml
+++ b/app/config/locales/en.yml
@@ -231,21 +231,20 @@ en:
         notice_no_answer: You must select an answer to continue.
       show:
         answer_no_bullet_1: You don't have another job (current or recent)
-        answer_no_bullet_2: You have a job that pays you via PayPal, Venmo, Cash App, etc
+        answer_no_bullet_2: You have a job that pays you via PayPal, Venmo, Cash App, etc.
         answer_no_bullet_3: You have a job that pays you in cash
-        answer_no_bullet_4: You have a job that pays you with a paper check and you cannot view your paycheck online
+        answer_no_bullet_4: You have a job that pays you with a paper check and you can't view your paycheck online
         answer_no_bullet_5: You have a job that issues a 1099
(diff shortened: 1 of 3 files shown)


app/config/locales/en.yml
```diff
@@ -231,21 +231,20 @@ en:
         notice_no_answer: You must select an answer to continue.
       show:
         answer_no_bullet_1: You don't have another job (current or recent)
-        answer_no_bullet_2: You have a job that pays you via PayPal, Venmo, Cash App, etc
+        answer_no_bullet_2: You have a job that pays you via PayPal, Venmo, Cash App, etc.
         answer_no_bullet_3: You have a job that pays you in cash
-        answer_no_bullet_4: You have a job that pays you with a paper check and you cannot view your paycheck online
+        answer_no_bullet_4: You have a job that pays you with a paper check and you can't view your paycheck online
```
jeffcatania-usds, line 236: Isn't cannot more formal than can't?

jeffcatania-usds (changes requested):
Can't vs cannot.  The rest is good.

app/config/locales/en.yml
millerti: This is the copy design sent us! I'll defer to them.

millerti: I checked with design and actually it was never supposed to read cannot -- it was always supposed to be can't. We aren't taking a very formal tone. So I think we should keep it this way.

jeffcatania-usds: OK! Approved.

jeffcatania-usds (approved):

PR #606, opened 2025-04-25
FFS-2695: Set report delivery email subject line to "CBV Report"

daphnegold:
## Ticket

Resolves [FFS-2695](https://jiraent.cms.gov/browse/FFS-2695).

## Changes
- Agency config for caseworker mailer email subject line
- Custom subject for LA
- Testing

## Acceptance testing

### DIFF
--- a/app/app/mailers/caseworker_mailer.rb
+++ b/app/app/mailers/caseworker_mailer.rb
@@ -11,7 +11,9 @@ class CaseworkerMailer < ApplicationMailer
     attachments[filename] = generate_pdf
     mail(
       to: @email_address,
-      subject: I18n.t("caseworker_mailer.summary_email.subject", case_number: case_number),
+      subject: view_context.agency_translation("caseworker_mailer.summary_email.subject",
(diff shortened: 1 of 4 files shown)


app/app/mailers/caseworker_mailer.rb
```diff
@@ -11,7 +11,9 @@ def summary_email
     attachments[filename] = generate_pdf
     mail(
       to: @email_address,
-      subject: I18n.t("caseworker_mailer.summary_email.subject", case_number: case_number),
+      subject: view_context.agency_translation("caseworker_mailer.summary_email.subject",
+                                case_number: case_number,
```
jeffcatania-usds, line 15: what would happen if an agency does not have a "case_number"?

tdooner (approved):

app/app/mailers/caseworker_mailer.rb
daphnegold: I'm not sure, is that possible for LA, @tdooner?

PR #607, opened 2025-04-25
FFS-2778 (Bugfix) Redirect when linking a previously linked account 

GeorgeCodes19:
## Ticket

Resolves [FFS-2778](https://jiraent.cms.gov/browse/FFS-2778).

## Changes

- Updated the `ArgyleModalAdapter` to send the payroll provider's identifier e.g. `item_id` in the request to create an Argyle token for displaying Argyle's Link Modal.
- Updated the `fetchArgyleToken` method to handle redirects
- Updated `Api::ArgyleController::create` to fetch the current user and check if the `item_id` (payroll provider/gig) is present in the user's list of connected accounts. If so then we'll send a redirect response to the client.

## Context for reviewers
This PR introduces functionality that prevents an applicant from entering Argyle's Link flow when attempting to link a previously linked account. 

## Acceptance testing

1. Start a new CBV session and login using ADP
2. After successfully authenticating you'll be redirected to the synchronization page. Press the back arrow to navigate back to the employer search page and attempt to connect ADP once more. If the `/synchronization` process has not be completed then you'll be redirected back to the `synchronizations` page. If the payroll account has fully synchronized then you'll be redirected to the `/payment_details` page.


app/spec/controllers/api/argyle_controller_spec.rb
```diff
@@ -3,16 +3,22 @@
 RSpec.describe Api::ArgyleController do
   include ArgyleApiHelper
 
-  context "#create" do
+  describe "#create" do
     let(:cbv_flow) { create(:cbv_flow) }
+    # get the argyle user id and item e.g. payroll/provider to include in the fetch_user_api_response stub
+    let(:argyle_user_id) { argyle_load_relative_json_file('', 'response_create_user.json')["id"] }
```
GeorgeCodes19, line 9: Plucking the value from the `response_create_user.json` file means that the Argyle supplied User `id` will match that of other stubbed requests i.e. `stub_create_user_response`

app/spec/support/argyle_api_helper.rb
```diff
@@ -71,6 +62,18 @@ def stub_get_webhook_subscriptions_response
       )
   end
 
+  def argyle_stub_fetch_user_api_response(user_folder = 'bob', override_properties = {})
```
GeorgeCodes19, line 65: I wanted to recycle the logic and the actual stubbed data, but be able to override what web mock will actually return. This is useful for chaining requests together to create a user flow.

app/app/controllers/api/argyle_controller.rb
```diff
@@ -25,6 +26,19 @@ def create
                    response["user_token"]
                  end
 
+    # Redirect if the user is attempting connect a previously connected account
+    argyle_user = argyle.fetch_user_api(user: @cbv_flow.argyle_user_id)
+    payroll_account = PayrollAccount::Argyle.find_by(cbv_flow: @cbv_flow)
```
tdooner, line 31: Ah, but, I think this might need to be more complex because there could be multiple Argyle PayrollAccounts, so we don't know which one is related to this Item ID. We would need to get all of the user's connected accounts, and see if any of the account IDs are in a payroll account. I think it would be something like:

```ruby
  connected_argyle_accounts = argyle.fetch_accounts_api(user: @cbv_flow.argyle_user_id, item: item_id)["results"]
  
  if connected_argyle_accounts.any?
    argyle_account = connected_argyle_accounts.first
    payroll_account = @cbv_flow.payroll_accounts.find_by(pinwheel_account_id: argyle_account["id"])

    # ... now use payroll_account to determine whether it's fully synced ...
  end
```

app/app/javascript/utilities/api.js
```diff
@@ -19,8 +19,20 @@ export const fetchPinwheelToken = (response_type, id, locale) => {
...
+    headers: {
+      "X-CSRF-Token": CSRF.token,
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({item_id: itemId})
   })
+
+  if (response.redirected) {
+    window.location.href = response.url
+    return
```
tdooner, line 34: Will returning nothing from this function cause any javascript errors in the console for the brief moment before the redirect occurs? Might be nice to do some null checking in the previous method or something? Not sure how JS works with stuff like that.

app/spec/controllers/api/argyle_controller_spec.rb
tdooner: Hmm, maybe we should just assign these as constants in ArgyleApiHelper, e.g.

```ruby
module ArgyleApiHelper
  BOB_USER_ID = "019571bb-e574-a88e-7d7b-15c886499696"
end
```
so we could refer to them in tests without having to do stuff like this. (I think we hardcode other IDs like this in our tests in other places.

app/spec/factories/payroll_account.rb
```diff
@@ -43,6 +43,21 @@
...
+        # Only sync the first 2 jobs
+        payroll_account.supported_jobs.first(2).each do |job|
+          payroll_account.webhook_events << build(
+            :webhook_event,
+            event_name: PayrollAccount::Argyle.event_for_job(job),
+            event_outcome: evaluator.with_errored_jobs.include?(job) ? "error" : "success"
+          )
+        end
+      end
+    end
```
tdooner, line 59: Nice! After my trio of PRs lands, this could probably be simplified since the `synchronization_status` will be stored in the database, but, I think this makes sense until then.

app/spec/javascript/controllers/__snapshots__/employer_search.spec.js.snap
```diff
@@ -50,7 +50,11 @@ exports[`EmployerSearchController with argyle > calls trackUserAction with data
 ]
 `;
 
-exports[`EmployerSearchController with argyle > fetches argyle token 1`] = `[]`;
+exports[`EmployerSearchController with argyle > fetches argyle token 1`] = `
+[
+  "uuid",
+]
+`;
```
tdooner, line 57: Huh. Did you have to do anything special to get vite to save these, or does it save them automatically as it runs?

app/spec/support/argyle_api_helper.rb
tdooner: Hmm. I see what you're trying to do, and it seems ok to do for this test case, but in general I think we need a better way to utilize the IDs of the objects within these request stubs. Maybe worth a team discussion about how to tame these.

tdooner (approved):
Looks very good to me - great work. In particular, thank you for cleaning up the tests as you went!

One logic/argyle implementation error I'd like to see you address before merging in the controller. We'll have to make sure to test this with multiple Argyle accounts to ensure it works for the subsequent accounts, too.

app/spec/javascript/controllers/__snapshots__/employer_search.spec.js.snap
GeorgeCodes19: [vitest automatically updates snapshots](https://vitest.dev/guide/snapshot.html#updating-snapshots) if you using the `u` arg.

app/spec/controllers/api/argyle_controller_spec.rb
GeorgeCodes19: I like the idea of the IDs being inferred from somewhere. The ArgyleApiHelper makes sense, but hardcoding them essentially duplicates the code as opposed to referencing the static JSON fixture.

app/spec/support/argyle_api_helper.rb
GeorgeCodes19: Fortunately we don't have to worry about it this time around since we don't need this method atm. The switch to use `@cbv_flow.argyle_user_id` and `fetch_accounts_api` pretty much nullifies this.

app/app/javascript/utilities/api.js
GeorgeCodes19: Hmm... I'm not sure. I'll do some follow up after merging this to unblock UAT. I do see where the value is destructured and I assume that would cause some type of runtime error.

app/spec/controllers/api/argyle_controller_spec.rb
GeorgeCodes19: I added a [test helper](https://github.com/DSACMS/iv-cbv-payroll/blob/e5e4e2b76ccc2afffb6122f33b9e69004c2f8553/app/spec/support/argyle_api_helper.rb#L134) that makes this more manageable.

PR #620, opened 2025-04-29
FFS-2688-paystub_failed: implemented LA copy

millerti:
## Changes
Implemented LA copy for /paystub_failed

## Testing instructions
2. Running locally, edit synchrnonizations_controller.rb:10 to be "if true" to force a sync failure
3. Verify the copy matches [the design](https://confluenceent.cms.gov/pages/viewpage.action?pageId=815436921)
4. Redo the workflow but with one employer already linked and review the copy again. (You may also just be able to reload the page, since it seems like the sync "completes" in the background somehow.)

## Acceptance testing

## Screenshot
image

### DIFF
--- a/app/config/locales/es.yml
+++ b/app/config/locales/es.yml
@@ -180,8 +180,7 @@ es:
         body_1: Usted ya ha completado con éxito su proceso de verificación, o está intentando acceder a esta invitación después de que ha expirado.
         body_2: Si todavía no termina de verificar sus ingresos, envíe su información de ingresos desde el mismo sitio web de %{agency_short_name}.
         cta_button_html:
-          ma: <a href="https://www.mass.gov/guides/how-to-contact-dta" class="usa-button">Visite el sitio web del DTA</a>
-          nyc: <a href="https://www.nyc.gov/site/hra/help/snap-application-frequently-asked-questions.page" class="usa-button">Visite el sitio web de la HRA</a>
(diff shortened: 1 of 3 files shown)


app/config/locales/es.yml
```diff
@@ -180,8 +180,7 @@ es:
         body_1: Usted ya ha completado con éxito su proceso de verificación, o está intentando acceder a esta invitación después de que ha expirado.
         body_2: Si todavía no termina de verificar sus ingresos, envíe su información de ingresos desde el mismo sitio web de %{agency_short_name}.
         cta_button_html:
-          ma: <a href="https://www.mass.gov/guides/how-to-contact-dta" class="usa-button">Visite el sitio web del DTA</a>
-          nyc: <a href="https://www.nyc.gov/site/hra/help/snap-application-frequently-asked-questions.page" class="usa-button">Visite el sitio web de la HRA</a>
+          la_ldh: <a href="http://mymedicaid.la.gov/" class="usa-button">Visite el sitio web del LDH</a>
```
GeorgeCodes19, line 183: Oh interesting. I wasn't aware that we have full HTML components inside our localization files.

GeorgeCodes19 (approved):

app/config/locales/es.yml
millerti: Yep, as long as you put _html at the end of the key, it works fine!

PR #621, opened 2025-04-29
FFS-2688-submit: implemented LA copy changes

millerti:
## Changes
Implemented LA copy for /submit
Removed ma/nyc strings

## Testing instructions
2. Proceed through the flow to the /submit page
3. Verify the copy matches [the design](https://confluenceent.cms.gov/pages/viewpage.action?pageId=932554140)

## Acceptance testing

## Screenshot
image

### DIFF
--- a/app/config/locales/en.yml
+++ b/app/config/locales/en.yml
@@ -398,8 +398,6 @@ en:
       show:
         additional_information_title: Is there anything else you'd like your caseworker to know about your income?
         application_or_recertification_date:
-          ma: Date client was contacted
-          nyc: SNAP application or recertification interview date
           sandbox: SNAP application or recertification interview date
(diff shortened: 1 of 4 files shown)


app/config/locales/en.yml
```diff
@@ -443,7 +443,8 @@ en:
             report_details: Report Details
         share_report_button: Share my report with %{agency_acronym}
         table_caption_no_name: Employer %{number}
-        warning_html: Your income report will be attached to your SNAP application.<br>You must check the legal agreement box to share your report.
+        warning_1: Your income report will be attached to your Medicaid application.
```
millerti, line 446: There is some question here about whether the <benefit> and the <application or renewal> pieces should be settings at the client_config level or if we can get away with just creating state-specific strings like I've done it here. Since these will both be static per each site, I have elected to do things this way. If that may change, then we will want to switch to the more complicated configuration solution.

tdooner: Not sure I understand exactly what the other option is, but I think I probably like this approach better than putting any more copy-related config into the client agency config.

tdooner (approved):

danielladevera-nava:
Note from acceptance testing, this should say (first line) "Your income report will be attached to _your Medicaid benefits_" (not super graceful, but for LA technically this is not an application nor a renewal. This is the copy we got from Mak/LA)

PR #627, opened 2025-04-29
FFS-2716 Generic link URL is the home page for Louisana

GeorgeCodes19:
Resolves [FFS-2716](https://jiraent.cms.gov/browse/FFS-2716).

## Changes

- Added a new method `redirect_to_client_agency_entries` which resolves the `client_agency_id` from the `Host` header in the request and redirects to the generic link entry page for the respective agency.

## Acceptance testing

To test this feature locally you have a couple options:
1. Modify the sandbox `agency_demo_domain` to be `sandbox.lvh.me`
2. Visit `sandbox.lvh.me:3000` which now resolves to the generic link page (`cbv/links/:client_agency_id`) for each respective client agency.
3. Users are redirected to the CBV entry page `cbv/entry` with a new session with proper `client_agency_id`

- [X] Acceptance testing after merge


app/spec/controllers/application_controller_spec.rb
```diff
@@ -78,7 +81,8 @@ def show
     it "identifies the correct agency config based on the domain name" do
```
GeorgeCodes19, line 81: This test may ultimately bring little value since we're already doing something similar [here](https://github.com/DSACMS/iv-cbv-payroll/pull/627/files#diff-76a2f7bd084e2eeb53e1d06ea8579b95594724cebe8bb668f4fcf0c503b22210R103).

app/spec/controllers/api/argyle_controller_spec.rb
```diff
@@ -55,9 +55,7 @@
 
         allow(CbvFlow).to receive(:find).and_return(cbv_flow)
         allow(controller).to receive(:argyle_for).and_return(argyle)
-        allow(controller).to receive(:agency_config).and_return({
-          cbv_flow.client_agency_id => double(argyle_environment: 'sandbox')
-        })
```
tdooner, line 60: This change appears residual from your other branch. The new value for this stub doesn't really make sense because you're stubbing a method with the same implementation it already has.
image

app/spec/controllers/application_controller_spec.rb
```diff
@@ -68,7 +71,7 @@ def show
     end
   end
 
-  describe 'client_agency_config can be resolved by domain name' do
+  describe '#detect_client_agency_from_domain' do
```
tdooner, line 74: Thanks. Looks like you've been reading that "better specs" page I sent 😄

app/spec/controllers/application_controller_spec.rb
```diff
@@ -87,4 +91,26 @@ def show
...
+      get :index
+      expect(response).to redirect_to(cbv_flow_new_path(client_agency_id: 'la_ldh'))
+    end
+
+    it 'does not redirect when hostname does not match any client agency domain' do
+      request.host = 'unknown.example.org'
+      get :index
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to eq('root')
+    end
```
tdooner, line 114: Let's add a test to the Cbv::GenericLinksController to verify that it works when accessed via a client agency domain.

app/spec/controllers/application_controller_spec.rb
GeorgeCodes19: Yup, no more guess work.

app/app/controllers/pages_controller.rb
```diff
@@ -1,7 +1,16 @@
...
+
   def home
   end
 
+  def redirect_to_client_agency_entries
+    client_agency_id = detect_client_agency_from_domain
+    if client_agency_id.present?
+      redirect_to cbv_flow_new_path(client_agency_id: client_agency_id)
+    end
+  end
```
tdooner, line 12: nit: In controllers, methods that aren't actually actions (i.e. that aren't renderable/don't have views) are idiomatically put at the bottom of the file under a `private` access modifier.

app/spec/controllers/cbv/generic_links_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "rails_helper"
+
+RSpec.describe Cbv::GenericLinksController do
+  describe '#show' do
+    context 'when the hostname matches a client agency domain' do
+      it do
```
tdooner, line 6: Let's name our specs as long as they are multi-line.

```suggestion
      it "starts a CBV flow" do
```

app/spec/controllers/cbv/generic_links_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "rails_helper"
+
+RSpec.describe Cbv::GenericLinksController do
+  describe '#show' do
+    context 'when the hostname matches a client agency domain' do
```
tdooner, line 5: The behavior being tested doesn't actually need the hostname to match, right? It's all based on `params[:client_agency_id]` in the GenericLinksController.

tdooner (approved):

PR #635, opened 2025-05-01
enforce last 4 of ssn for argyle

jeffcatania-usds:
## Changes
* Argyle sends back the full SSN on the identities endpoint.  This ensures that we only capture the last 4 characters max.

## Acceptance testing

- [x ] No acceptance testing needed

### DIFF
--- a/app/app/services/aggregators/response_objects/identity.rb
+++ b/app/app/services/aggregators/response_objects/identity.rb
@@ -26,7 +26,7 @@ module Aggregators::ResponseObjects
         full_name: identity_response_body["full_name"],
         date_of_birth: identity_response_body["birth_date"],
         emails: [ identity_response_body["email"] ],
-        ssn: identity_response_body["ssn"],
+        ssn: (identity_response_body["ssn"]&.last(4) if identity_response_body["ssn"].present?),
         phone_numbers: [ identity_response_body["phone_number"] ]
       )
     end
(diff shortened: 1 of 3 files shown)


app/app/services/aggregators/response_objects/identity.rb
```diff
@@ -26,7 +26,7 @@ def self.from_argyle(identity_response_body)
         full_name: identity_response_body["full_name"],
         date_of_birth: identity_response_body["birth_date"],
         emails: [ identity_response_body["email"] ],
-        ssn: identity_response_body["ssn"],
+        ssn: (identity_response_body["ssn"]&.last(4) if identity_response_body["ssn"].present?),
```
tdooner, line 29: Thoughts on obfuscating it similar to how we do for pinwheel?

```suggestion
        ssn: ("XXX-XX-#{identity_response_body["ssn"]&.last(4)}" if identity_response_body["ssn"].present?),
```

I guess this doesn't handle the case where SSN is only two digits long. Not sure if we're likely to hit that or if Argyle will give us back a proper length SSN all the time.

app/spec/services/aggregators/response_objects/identity_spec.rb
```diff
@@ -31,8 +31,34 @@
...
+        expect(identity.ssn).to eq("2222")
+      end
+
+      it "shortens 2-digit ssn to 4 digits" do
+        argyle_response["ssn"] = "22"
+        identity = described_class.from_argyle(argyle_response)
+        expect(identity.ssn).to eq("22")
+      end
+
+      it "nil" do
```
tdooner, line 57:
```suggestion
      it "handles a nil ssn" do
```

tdooner (approved):

PR #640, opened 2025-05-01
FFS-2821: /success does not show confirmation number

daphnegold:
## Ticket

Resolves [FFS-2821](https://jiraent.cms.gov/browse/FFS-2821).

## Changes
Fix race condition with job-generated confirmation code not appearing on the `/success` page by moving the `confirmation_code` generation to the `submits_controller`.

## Acceptance testing
- Proceed through the flow until the `/success` page
- Observe that confirmation code is populated

Screenshot 2025-05-01 at 4 33 20 PM
Screenshot 2025-05-01 at 4 32 25 PM

### DIFF
--- a/app/app/jobs/case_worker_transmitter_job.rb
+++ b/app/app/jobs/case_worker_transmitter_job.rb
@@ -13,11 +13,6 @@ class CaseWorkerTransmitterJob < ApplicationJob
     @cbv_flow = cbv_flow
     current_agency = current_agency(@cbv_flow)
 
-    # Confirmation code is now generated in the SubmitsController
-    if cbv_flow.confirmation_code.blank?
-      Rails.logger.warn("CbvFlow #{cbv_flow_id} does not have a confirmation code. This should have been generated in the SubmitsController.")
-    end
-
     if current_agency.transmission_method.empty?
       Rails.logger.info("No transmission method found for client agency #{current_agency.id}")
       return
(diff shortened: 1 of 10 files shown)


app/spec/support/title_checker_helper.rb
```diff
@@ -47,6 +47,7 @@ def should_check_title?(request, response)
       !%w[new create].include?(request.params[:action]) &&
       response.body.present? &&
       response.status == 200 &&
-      response.content_type =~ /html/
+      response.content_type =~ /html/ &&
+      request.format != :turbo_stream
```
daphnegold, line 51: skip title checking for turbo stream responses since they're partial updates without html tags

app/app/jobs/case_worker_transmitter_job.rb
```diff
@@ -13,18 +13,16 @@ def perform(cbv_flow_id)
     @cbv_flow = cbv_flow
     current_agency = current_agency(@cbv_flow)
 
+    # Confirmation code is now generated in the SubmitsController
     if cbv_flow.confirmation_code.blank?
-      confirmation_code = generate_confirmation_code(cbv_flow)
-      cbv_flow.update!(confirmation_code: confirmation_code)
+      Rails.logger.warn("CbvFlow #{cbv_flow_id} does not have a confirmation code. This should have been generated in the SubmitsController.")
```
tdooner, line 18: This strikes me as a bit paranoid, but, I suppose it won't hurt anything here to have this check.

tdooner (approved):

app/app/jobs/case_worker_transmitter_job.rb
daphnegold: I added it because we have lots of rails loggers in the job, to follow convention. But I see your point, happy to remove either way.

tdooner (approved):

PR #641, opened 2025-05-01
FFS-2749: Display LA Indexing Fields

jeffcatania-usds:
## Ticket

Resolves [FFS-2749](https://jiraent.cms.gov/browse/FFS-2749).

## Changes

Updates to PDF Summary:
* (Caseworker only) Include obfuscated SSN under Payment details section.  The format will be XXX-XX-####.
* Include date of birth in Client and Report information section based on the user's manual indexing fields.  Date of birth is rendered as a human readable date format (e.g. "May 10, 2025")
* Include "Client-provided information" superscript disclosure on both caseworker and applicant views if there are any applicant_attributes detected.

### Applicant View
image

### Caseworker View
image

## Context for reviewers

* Recommend refactoring the `applicant_attributes` object away from a simple list of symbols to having the full configuration.  This would allow required field attributes and types within the field itself so that it doesn't have to be hardcoded the way it is in this PR.

## Acceptance testing


app/app/views/cbv/submits/show.pdf.erb
```diff
@@ -93,8 +93,9 @@
         <%= t(".pdf.shared.employment_information") %>
       </h4>
     <% end %>
-    <% if is_caseworker && summary[:has_identity_data] %>
+    <% if summary[:has_identity_data] && is_caseworker %>
```
jeffcatania-usds, line 96:
```suggestion
    <% if is_caseworker && summary[:has_identity_data] %>
```

app/app/views/cbv/submits/show.pdf.erb
```diff
@@ -35,7 +35,7 @@
 <% end %>
 
 <h2><%= t(".pdf.client.client_report_information") %></h2>
-<% if is_caseworker %>
+<% if @cbv_flow.cbv_applicant.applicant_attributes&.filter {|item| item.present?}.length > 0 %>
```
iannorriswork, line 38: shouldn't need the &.filter here --> set_applicant_attributes establishes applicant_attributes as an empty array by default.

iannorriswork: actually, looking at what you actually want here, i'd suggest @cbv_flow.cbv_applicant.applicant_attributes.present?

iannorriswork:
one minor comment on the PR, otherwise LGTM

we should watch out for if we do more customization on this page/branching display logic--> it's getting pretty close to time to refactor it into a component.

iannorriswork (changes requested):
change the applicant_information.find check then we'll be good to go

app/app/views/cbv/submits/show.pdf.erb
jeffcatania-usds: Thanks! I tried to make it the same logic as what actually renders the table. But I agree that we just need to see if there is anything in this array.

PR #644, opened 2025-05-02
FFS-2761: Add DOB field to /applicant_info page LA_LDH & Sandbox

GeorgeCodes19:
## Ticket

Resolves [FFS-2761](https://jiraent.cms.gov/browse/FFS-2761).

## Changes
- [x] Add date_of_birth field to the CbvApplicant model
- [x] Set the field as required for sandbox and la_ldh agencies
- [x] Display the field in the applicant information forms
- [x] Add appropriate error messages in English and Spanish
- [x] Add field validations for `date_of_birth`

## Context for reviewers
- We've opted to use the date_picker component from the [uswds_form_builder](https://github.com/DSACMS/iv-cbv-payroll/blob/158d2a59096077a5141056e39d96826d80aa3cbb/app/app/helpers/uswds_form_builder.rb)
- No DB-level constraints were added since requirements are agency-specific

## Acceptance testing

Screenshot 2025-05-02 at 5 10 29 PM

Screenshot 2025-05-02 at 3 05 53 PM
Screenshot 2025-05-02 at 3 05 17 PM


app/app/controllers/cbv/applicant_informations_controller.rb
```diff
@@ -7,32 +7,34 @@ def show
...
-    rescue => e
-      Rails.logger.error("Error updating applicant: #{e.message}")
-      flash[:alert] = t(".error_updating_applicant")
-      return redirect_to cbv_flow_applicant_information_path
+    @cbv_applicant.assign_attributes(applicant_params[:cbv_applicant])
+    @cbv_applicant.valid? # run the model validations
+    @cbv_applicant.validate_required_applicant_attributes # run the attributes validations
+
+    if @cbv_applicant.errors.empty?
+      @cbv_applicant.save!
```
GeorgeCodes19, line 15: I plan to catch this exception if it's raised and log it.

app/app/models/cbv_applicant.rb
```diff
@@ -96,4 +115,9 @@ def set_applicant_attributes
     @applicant_attributes = Rails.application.config.client_agencies[client_agency_id]&.applicant_attributes&.compact&.keys&.map(&:to_sym) || []
     @required_applicant_attributes = Rails.application.config.client_agencies[client_agency_id]&.applicant_attributes&.select { |key, attributes| attributes["required"] }&.keys&.map(&:to_sym) || []
   end
+
+  def date_of_birth_required?
```
GeorgeCodes19, line 119: We could consider making this more generic. Something like `client_attribute_required(attribute)?`

app/app/models/cbv_applicant.rb
```diff
@@ -53,6 +68,10 @@ def has_applicant_attribute_missing?
     end
   end
 
+  def is_valid?
```
GeorgeCodes19, line 71: @jeffcatania-usds suggested using a more descriptive name here due to the `validate_required_applicant_attributes` happening in the background.

app/app/models/cbv_applicant.rb
```diff
@@ -73,27 +94,42 @@ def set_snap_application_date
...
+    .include?(attribute)
+  end
+
+  def get_applicant_attribute_type(attribute)
+    Rails.application.config.client_agencies[client_agency_id]&.applicant_attributes&.dig(attribute, "type")
+  end
+
+  private
+
+  def get_required_applicant_attributes
```
GeorgeCodes19, line 132: This method was broken out so that we can reuse it a couple times since `@required_applicant_attributes` wasn't available for use in the `before_validation` - which we rely on to parse date inputs.

app/app/models/cbv_applicant.rb
```diff
@@ -32,9 +32,23 @@ def self.valid_attributes_for_agency(client_agency_id)
   has_many :cbv_flows
   has_many :cbv_flow_invitations
 
-  before_validation :parse_snap_application_date
+  before_validation { parse_date("snap_application_date") }
+  before_validation { parse_date("date_of_birth") if self["date_of_birth"].present? }
```
GeorgeCodes19, line 36: Not all clients have `date_of_birth` set in their `applicant_attributes` so it's not guaranteed that `date_of_birth` will be present on the model instance.

app/config/client-agency-config.yml
```diff
@@ -154,3 +157,6 @@
       required: true
     case_number:
       required: true
+    date_of_birth:
+      required: true
+      type: date
```
GeorgeCodes19, line 162: I don't think we need this. We could have a `before_validation` method on the `CbvClient` that parses all date fields, but I'm explicitly parsing this field instead. I think that might be easier to reason with as opposed to an iterator that extracts and parses specific fields.

app/app/models/cbv_applicant.rb
```diff
@@ -73,27 +94,42 @@ def set_snap_application_date
...
-    @required_applicant_attributes = Rails.application.config.client_agencies[client_agency_id]&.applicant_attributes&.select { |key, attributes| attributes["required"] }&.keys&.map(&:to_sym) || []
+    @required_applicant_attributes = get_required_applicant_attributes
+  end
+
+  def is_applicant_attribute_required?(attribute)
+    get_required_applicant_attributes
+    .include?(attribute)
+  end
+
+  def get_applicant_attribute_type(attribute)
```
GeorgeCodes19, line 126: This is dead code is we're not going to dynamically parse all `applicant_attributes` with `type = date`

app/app/models/cbv_applicant.rb
```diff
@@ -73,27 +94,42 @@ def set_snap_application_date
     self.snap_application_date ||= Date.current
   end
 
-  def parse_snap_application_date
-    raw_snap_application_date = @attributes["snap_application_date"]&.value_before_type_cast
-    return if raw_snap_application_date.is_a?(Date)
+  def parse_date(attribute)
```
GeorgeCodes19, line 97: This method was made more generic to support parsing various `date_picker` inputs.

app/app/controllers/cbv/applicant_informations_controller.rb
```diff
@@ -7,32 +7,37 @@ def show
...
   def update
-    begin
-      @cbv_applicant.update!(applicant_params[:cbv_applicant])
-    rescue => e
-      Rails.logger.error("Error updating applicant: #{e.message}")
-      flash[:alert] = t(".error_updating_applicant")
-      return redirect_to cbv_flow_applicant_information_path
+    @cbv_applicant.assign_attributes(applicant_params[:cbv_applicant])
+
+    if @cbv_applicant.is_valid?
```
GeorgeCodes19, line 12: `.is_valid` wraps all validation operations (ActiveRecord and `validate_required_applicant_attributes`) into a single operation that reliably appends errors to model which are then used via `@cbv_applicant.errors`

app/app/controllers/cbv/applicant_informations_controller.rb
```diff
@@ -7,32 +7,37 @@ def show
...
-      @cbv_applicant.update!(applicant_params[:cbv_applicant])
-    rescue => e
-      Rails.logger.error("Error updating applicant: #{e.message}")
-      flash[:alert] = t(".error_updating_applicant")
-      return redirect_to cbv_flow_applicant_information_path
+    @cbv_applicant.assign_attributes(applicant_params[:cbv_applicant])
+
+    if @cbv_applicant.validate_base_and_applicant_attributes?
+      begin
+        @cbv_applicant.save
```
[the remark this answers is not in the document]
jeffcatania-usds: Note: this is a one off comment based. Not a full review.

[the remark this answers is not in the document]
tdooner: fyi - `model.save` is the method that returns a boolean (there is no such thing as `model.save?`)

`model.save` will only raise if there is some kind of database connection issue, so rescuing here is probably a bit of overkill, but it at least seems functional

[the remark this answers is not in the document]
iannorriswork: i think it's a question of why the begin/rescue existed in the first place--i'm pretty sure that's because they save!'d then checked for errors. that's not standard rails practice---> in fact, we could be simplifying our code a lot 

```
if  @cbv_applicant.validate_base_and_applicant_attributes? && @cbv_applicant.save
      return redirect_to next_path
    end
    
    error_header = t(".error_header", count: error_count)

    # Use @cbv_applicant.errors directly since we added all errors back to it
    error_messages = @cbv_applicant.errors.map { |error| "<li>#{error.message}</li>" }.join
    error_messages = "<ul>#{error_messages}</ul>"

    flash.now[:alert_heading] = error_header
    flash.now[:alert] = error_messages.html_safe

    track_applicant_information_error_event(error_messages.html_safe)

    render :show, status: :unprocessable_entity
    
# note: if we TRULY want that rescue behavior, keep it here. i still think this was a legacy of the save! hack, but it can go here
    rescue Exception => e
      Rails.logger.error("Error updating applicant: #{e.message}")
      flash[:alert] = t(".error_updating_applicant")
      redirect_to cbv_flow_applicant_information_path
    end
```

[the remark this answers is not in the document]
GeorgeCodes19: All good points and thanks for the more simple approach @iannorriswork. So if I'm understanding things correctly- Rails will only raise an exception if we explicitly add a `!` to the method. Because we're validating prior to saving the likelihood of this query failing is negligible. This is a lot easier to reason with.

app/config/client-agency-config.yml
```diff
@@ -154,3 +157,5 @@
       required: true
     case_number:
       required: true
+    date_of_birth:
+      required: true
```
millerti, line 161: Why is type: date needed above but not here?

app/app/models/cbv_applicant.rb
```diff
@@ -44,15 +62,28 @@ def self.valid_attributes_for_agency(client_agency_id)
     case_number: :string,
     agency_id_number: :string,
     beacon_id: :string,
-    snap_application_date: :date
+    snap_application_date: :date,
+    date_of_birth: :date
   )
 
+  def date_of_birth=(value)
```
GeorgeCodes19, line 69: These methods were introduced because `ActiveRecord::assign_attributes` will call these setter methods by default rather than relying on Rails' default date parsing. 

Rails' default date parsing is not configured to parse dates in the `mm/dd/YYYY` string format. By default Rails handles an `iso8601`, `dd/mm/YYYY` (day/month/year) formatted date strings  when parameters in the request are identified to be a `Date` type (as defined by the db schema). Since `mm` and `dd` are swapped this means that Rails is not expecting a value higher than 12 for the month.

The result would be an invalid date when assigning parameters to a model  e.g. `@cbv_applicant.assign_attributes(applicant_params[:cbv_applicant])`- thus resulting in Date fields not being persisted to the model. This is exactly why we witnessed providing a date such as `03/19/1992` not being persisted on the model, but `03/03/1992` being persisted.

app/config/client-agency-config.yml
GeorgeCodes19: Good catch! The one above isn't needed anymore due to how dates are handled in the model.

PR #646, opened 2025-05-02
FFS-2827: Log /gigs and /shifts analytics for Pinwheel and Argyle

tdooner:
## Ticket

Resolves [FFS-2827](https://jiraent.cms.gov/browse/FFS-2827).

## Changes

**FFS-2827: Log /gigs and /shifts analytics for Pinwheel and Argyle**

## Acceptance testing

### DIFF
--- a/app/app/models/payroll_account/pinwheel.rb
+++ b/app/app/models/payroll_account/pinwheel.rb
@@ -5,11 +5,13 @@ class PayrollAccount::Pinwheel < PayrollAccount
     "paystubs" => "paystubs.fully_synced",
     "employment" => "employment.added",
     "income" => "income.added",
-    "identity" => "identity.added"
+    "identity" => "identity.added",
+    "gigs" => "shifts.added"
   }
 
   def has_fully_synced?
     (supported_jobs.exclude?("paystubs") || find_webhook_event(JOBS_TO_WEBHOOK_EVENTS["paystubs"]).present?) &&
+      (supported_jobs.exclude?("shifts") || find_webhook_event(JOBS_TO_WEBHOOK_EVENTS["shifts"]).present?) &&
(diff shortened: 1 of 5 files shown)


app/app/models/payroll_account/pinwheel.rb
```diff
@@ -5,11 +5,13 @@ class PayrollAccount::Pinwheel < PayrollAccount
...
     "employment" => "employment.added",
     "income" => "income.added",
-    "identity" => "identity.added"
+    "identity" => "identity.added",
+    "gigs" => "shifts.added"
   }
 
   def has_fully_synced?
     (supported_jobs.exclude?("paystubs") || find_webhook_event(JOBS_TO_WEBHOOK_EVENTS["paystubs"]).present?) &&
+      (supported_jobs.exclude?("shifts") || find_webhook_event(JOBS_TO_WEBHOOK_EVENTS["shifts"]).present?) &&
```
tdooner, line 14: These changes were necessary to support `job_succeeded?("shifts")`. We'll have to re-register the webhooks for Pinwheel for this.

app/spec/controllers/webhooks/pinwheel/events_controller_spec.rb
```diff
@@ -125,10 +125,10 @@
           pinwheel_stub_request_shifts_response
 
           allow(controller).to receive(:event_logger).and_return(event_logger)
-          allow(event_logger).to receive(:track)
```
tdooner, line 128: This is what was preventing the big assertion from actually causing the test to fail. I think the issue is that no matter what happened within that other assertion, this assertion would take precedence and the test would consider the test passed since `event_logger` did receive `track`, even if it wasn't with the arguments called for in the block.

jeffcatania-usds (approved):
LGTM

PR #653, opened 2025-05-06
FFS-2688: Various fixes

millerti:
## Changes
Various fixes found during AC testing of other copy update PRs. These include:
1. In LA, we should not say "Medicaid application" but instead "Medicaid benefits". This has been changed on /applicant_information, /payment_details, the PDF, and /success.
2. In the configs, I had left an old URL for Louisiana's website. But in addition to correcting that, I've updated all the string tokens to load the URL from settings instead of hardcoding it. There is now one exception: the URL on the /entry page should still go to mymedicaid.la.gov. 
3. On /entry, you should now be able to open both of the accordions at the bottom of the screen at the same time. Before, when you opened one, the other would close.
4. On the help modal under the company ID option, we had abbreviated "info" but it is now the whole word "information".
5. Because of outdated updates to the /employer_search page, I've re-implemented the changes. Verify that the copy matches [the design](https://confluenceent.cms.gov/pages/viewpage.action?pageId=748887611).
6. All links should now open in a new tab.

## Testing instructions
2. Proceed through the flow to the pages mentioned above and make sure the copy you see reflects the fix described.

## Acceptance testing

### DIFF
--- a/app/app/views/cbv/entries/show.html.erb
+++ b/app/app/views/cbv/entries/show.html.erb
@@ -28,7 +28,7 @@
 
 <div class="divider margin-top-5"></div>
 
-<div class="usa-accordion usa-accordion--bordered margin-top-5">
+<div class="usa-accordion usa-accordion--bordered usa-accordion--multiselectable margin-top-5" data-allow-multiple>
   <h4 class="usa-accordion__heading usa-prose">
     <button
       type="button"
--- a/app/app/views/cbv/entries/show.html.erb
+++ b/app/app/views/cbv/entries/show.html.erb
@@ -55,7 +55,7 @@
   <div id="what_if_i_cant_use_this_accordion" class="usa-accordion__content usa-prose">
     <p>
       <%= t(".what_if_i_cant_use_this_body_1", agency_short_name: current_agency.agency_short_name) %>
-      <%= agency_translation(".what_if_i_cant_use_this_body_2_html", agency_url: agency_url) %>
+      <%= agency_translation(".what_if_i_cant_use_this_body_2_html", agency_url: "https://mymedicaid.la.gov/") %>
     </p>
   </div>
 </div>
--- a/app/app/views/cbv/entries/show.html.erb
+++ b/app/app/views/cbv/entries/show.html.erb
@@ -55,7 +55,7 @@
   <div id="what_if_i_cant_use_this_accordion" class="usa-accordion__content usa-prose">
     <p>
       <%= t(".what_if_i_cant_use_this_body_1", agency_short_name: current_agency.agency_short_name) %>
-      <%= agency_translation(".what_if_i_cant_use_this_body_2_html") %>
+      <%= agency_translation(".what_if_i_cant_use_this_body_2_html", agency_url: agency_url) %>
     </p>
   </div>
 </div>
--- a/app/config/locales/en.yml
+++ b/app/config/locales/en.yml
@@ -246,7 +246,7 @@ en:
         criteria_disclaimer: 'Note: If you''ve had other jobs in the past %{pay_income_days} days that don''t meet these criteria, you may need to submit that income information separately.'
         header: Do you have another job to report?
         learn_more_link_html:
-          la_ldh: Learn more on <a href="%{agency_url}" target="_blank" rel="noopener noreferrer">LDH's website</a>.
+          la_ldh: Learn more on <a href="%{agency_url} target="_blank"" target="_blank" rel="noopener noreferrer">LDH's website</a>.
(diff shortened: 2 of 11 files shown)


app/app/views/cbv/entries/show.html.erb
```diff
@@ -55,7 +55,7 @@
   <div id="what_if_i_cant_use_this_accordion" class="usa-accordion__content usa-prose">
     <p>
       <%= t(".what_if_i_cant_use_this_body_1", agency_short_name: current_agency.agency_short_name) %>
-      <%= agency_translation(".what_if_i_cant_use_this_body_2_html") %>
+      <%= agency_translation(".what_if_i_cant_use_this_body_2_html", agency_url: "https://mymedicaid.la.gov/") %>
```
millerti, line 58: Confirmed that this link **should** be different than the others (with Makaela).

app/config/locales/en.yml
```diff
@@ -333,7 +332,7 @@ en:
         what_if_i_cant_use_this_body_1: If you need to report your income and can’t use this tool, then you’ll need to share your income information for this job directly with %{agency_short_name}.
         what_if_i_cant_use_this_body_2_html:
           default: Visit <a href="https://example.com/contact">CBV's website</a> for more information on how to submit documents.
-          la_ldh: 'You can: <ul><li>Upload documents through the Self Service Portal online at <a href="https://MyMedicaid.LA.GOV">MyMedicaid.LA.GOV</a></li><li>Send documents through US Mail: Louisiana Medicaid/LaCHIP, P.O. Box 91283, Baton Rouge, LA 70821-9278</li><li>Email documents: <a href="mailto:MyMedicaid@la.gov">MyMedicaid@la.gov</a></li><li>Bring your documents to the Medicaid office of choice.</li></ul>'
+          la_ldh: 'You can: <ul><li>Upload documents through the Self Service Portal online at <a href="%{agency_url}">%{agency_url}</a></li><li>Send documents through US Mail: Louisiana Medicaid/LaCHIP, P.O. Box 91283, Baton Rouge, LA 70821-9278</li><li>Email documents: <a href="mailto:MyMedicaid@la.gov">MyMedicaid@la.gov</a></li><li>Bring your documents to the Medicaid office of choice.</li></ul>'
```
GeorgeCodes19, line 335: Not a blocker, but there's a lot of HTML here. I could see this being a view since it seems dynamic and dependent on the client agency ID. I suppose that would really depend on if we expect the same options for other agencies. If we expect a templated response here a view could look something like this:

```html
<ul>
<li>
Upload documents through the Self Service Portal online at %{agency_url}
</li>
<li>Send documents through US Mail %{addresss}</li>
<li>Email documents: %{client_agency_email}</li>
<li>Bring your documents to the Medicaid office of choice</li>
```

GeorgeCodes19 (approved):
LGTM. I provided just a couple feedback points

- Adding a `target="_blank"` attribute to the link located in the "What if I can't use this tool" accordion item
- Perhaps we should consider using a view partial if we anticipated templates answers to the questions on the entry page.

app/config/locales/en.yml
millerti: I'm going to skip this but I agree that having so much HTML in a string token isn't ideal. Our translators are already instructed to work around it so that's why I'm ok with leaving it this time.

app/app/views/cbv/entries/show.html.erb
millerti: Good idea. I've made all the links open in new tabs.

app/config/locales/en.yml
```diff
@@ -246,7 +246,7 @@ en:
         criteria_disclaimer: 'Note: If you''ve had other jobs in the past %{pay_income_days} days that don''t meet these criteria, you may need to submit that income information separately.'
         header: Do you have another job to report?
         learn_more_link_html:
-          la_ldh: Learn more on <a href="%{agency_url}" target="_blank" rel="noopener noreferrer">LDH's website</a>.
+          la_ldh: Learn more on <a href="%{agency_url} target="_blank"" target="_blank" rel="noopener noreferrer">LDH's website</a>.
```
millerti, line 258: un-screw-up this

GeorgeCodes19 (approved):

danielladevera-nava:
seeing this on /payment_details, but I believe this bug is already identified in [this ticket](https://jiraent.cms.gov/browse/FFS-2842)
image

danielladevera-nava:
/add_job: the link in the body text above the CTA (https://ldh.la.gov/renew-medicaid%20target=) opens a 404 page. need to remove the extra characters at the end (should be https://ldh.la.gov/renew-medicaid)

danielladevera-nava:
/summary: this should read "your Medicaid benefits" 
image

PR #658, opened 2025-05-08
FFS-2852: Fix date format issues

jeffcatania-usds:
## Ticket

Resolves [FFS-2852](https://jiraent.cms.gov/browse/FFS-2852).

## Changes

* Fixes issue where all Spanish dates were off by a month (e.g. January 1 became 1 Febrero). 
    * This is caused by Ruby's default Date string mappings set 0'th index to nil so that January has index position 1. 
* Copied all spanish date translation configuration from [i18n gem template](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/es.yml). 
 
image
image

## Acceptance testing

### DIFF
--- a/app/spec/helpers/report_view_helper_spec.rb
+++ b/app/spec/helpers/report_view_helper_spec.rb
@@ -176,7 +176,7 @@ RSpec.describe ReportViewHelper, type: :helper do
         date_string = Date.new(2023, 8, 7) # A Tuesday
         format = "%b"
 
-        expect(helper.format_date(date_string, format: format)).to match(/ago\./)
+        expect(helper.format_date(date_string, format: format)).to match(/ago/)
       end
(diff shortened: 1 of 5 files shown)


app/spec/helpers/report_view_helper_spec.rb
```diff
@@ -87,4 +87,105 @@
       end
     end
   end
+
+  describe '#format_date and #format_parsed_date' do
```
iannorriswork, line 91: i know they're definitely adjacent functions, but rails convention really prefers you test one method at a time for the most part--> would prefer a describe "#format_date" and a describe "format_parsed_date".

iannorriswork (approved):
one nitpicky-ish comment, but otherwise awesome tests!

PR #659, opened 2025-05-08
FFS-2842: Update source of pay_frequency for argyle

millerti:
## Changes
In Argyle, pay_frequency should come from Identities.pay_cycle, not Identities.base_pay.period.

## Testing Instructions
0. Set up your local dev instance to use Argyle (set SUPPORTED_PROVIDERS).
1. Navigate through the CBV workflow until you get to /payment details and observe the pay frequency. It should match what was returned in the Identities.pay_cycle value.
2. Also verify this value on on the PDF
3. Repeat the test using Pinwheel and verify that the pay frequency reporting is unchanged in these places

## Acceptance testing

## Screenshot
image

### DIFF
--- a/app/app/services/aggregators/response_objects/income.rb
+++ b/app/app/services/aggregators/response_objects/income.rb
@@ -1,11 +1,11 @@
-INCOME_FIELDS = %i[
-  account_id
-  pay_frequency
-  compensation_amount
-  compensation_unit
-]
-
 module Aggregators::ResponseObjects
+  INCOME_FIELDS = %i[
+    account_id
+    pay_frequency
+    compensation_amount
+    compensation_unit
+  ]
+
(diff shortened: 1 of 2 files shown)


app/app/services/aggregators/response_objects/income.rb
```diff
@@ -1,11 +1,11 @@
...
-  compensation_unit
-]
-
 module Aggregators::ResponseObjects
+  INCOME_FIELDS = %i[
+    account_id
+    pay_frequency
+    compensation_amount
+    compensation_unit
+  ]
```
millerti, line 7: Moved at Tom's request.

GeorgeCodes19: Cool, this makes sense to me.

GeorgeCodes19 (approved):
LGTM!

PR #664, opened 2025-05-09
FFS-2867: Fix incorrect credentials Arygle modal not reopening

tdooner:
## Ticket

N/A

## Context for reviewers

I am performing these tests to verify correctness:
- [x] Argyle Bob / standard W2 account  - correct password
- [x] Argyle Bob / W2 - incorrect password (closing and reopening modal)
- [x] Argyle Bob / W2 - incorrect MFA (closing and reopening modal)
- [x] Argyle Bob / gig account
- [x] Argyle Bob / clicking back from /synchronizations and reopening Argyle
- [x] Argyle Bob / clicking back from /payment_details and reopening Argyle
- [x] Ann Loanseeker (has failed account sync)

## Acceptance testing

### DIFF
--- a/app/app/controllers/api/argyle_controller.rb
+++ b/app/app/controllers/api/argyle_controller.rb
@@ -9,6 +9,9 @@ class Api::ArgyleController < ApplicationController
     @cbv_flow = CbvFlow.find(session[:cbv_flow_id])
     argyle = argyle_for(@cbv_flow)
     item_id = params[:item_id]
+    unless item_id.present?
+      return render json: { status: :error, message: "Invalid item_id" }, status: :unprocessable_entity
+    end
 
     is_sandbox_environment = agency_config[@cbv_flow.client_agency_id].argyle_environment == "sandbox"
     user_token = if @cbv_flow.argyle_user_id.blank?
@@ -29,7 +32,7 @@ class Api::ArgyleController < ApplicationController
     # Redirect if the user is attempting connect a previously connected account
     connected_argyle_accounts = argyle.fetch_accounts_api(user: @cbv_flow.argyle_user_id, item: item_id)["results"]
 
-    if item_id.present? && connected_argyle_accounts.any?
+    if connected_argyle_accounts.any?
       argyle_account = connected_argyle_accounts.first
       payroll_account = @cbv_flow.payroll_accounts.find_by(pinwheel_account_id: argyle_account["id"])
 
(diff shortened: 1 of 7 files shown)


app/app/controllers/api/argyle_controller.rb
```diff
@@ -50,6 +33,42 @@ def create
...
+    return unless payroll_account.present?
+
+    # If we've made it here, there is a previous connection to that item.
+    #
+    # Redirect to the proper place based on the status of the connection.
+    if payroll_account.sync_succeeded? || payroll_account.sync_failed?
+      redirect_to cbv_flow_payment_details_path(user: { account_id: payroll_account.pinwheel_account_id })
+    elsif payroll_account.job_succeeded?("accounts")
+      # Sync is in progress, and the user has successfully connected their account.
+      redirect_to cbv_flow_synchronizations_path(user: { account_id: payroll_account.pinwheel_account_id })
```
millerti, line 68: Should there be a redirect for the else case here?

app/spec/controllers/api/argyle_controller_spec.rb
```diff
@@ -17,11 +18,11 @@
     context "when the CbvFlow does not have an argyle_user_id" do
       before do
         stub_create_user_response
-        argyle_stub_request_accounts_response('bob')
+        argyle_stub_request_empty_accounts_response
```
millerti, line 21: Why this change? Surely the tests this runs before expect bob's response?

millerti:
Can you explain a little bit more what the workflow looked like when this failed? That will help me test locally and be sure I'm hitting the case you expect to fix. I think the checked-off test cases you list are close, but maybe something like:

1. Try to log into the Argyle modal with bad credentials
2. When the modal says they're invalid, close it
3. Attempt to reopen the modal
4. At this point, before my changes, sometimes the modal just wouldn't reopen. Now, it should open as you'd expect. Test this multiple times in a row with different ways of invalidating your credentials, like getting MFA wrong, etc.

tdooner:
@millerti Your guess about the workflow is correct - it was closing and reopening the modal after receiving an error for incorrect password or MFA code.

app/app/controllers/api/argyle_controller.rb
tdooner: Nope, no redirect because the user hasn't yet connected their account successfully (with a correct password/MFA).

app/spec/controllers/api/argyle_controller_spec.rb
tdooner: This test case is in a context where the CbvFlow has no argyle_user_id, which means the user hasn't yet linked an account, so I had a problem in the actual method where it was mistakenly finding an account. So I made this change to be more representative of a real situation - where there is no account that matches the newly-created user.

tdooner:
I synced up with Eryn - she is going to acceptance test in demo. And I also synced up with Jeff about the general approach of the "unknown" webhook status. So I am going to merge this!

PR #665, opened 2025-05-09
Fix SolidQueue database schema and config

tdooner:
## Ticket

Follow-up from 9f2853834be.

## Changes
- **Add migrations from solid_queue 0.6.0**
- **Update solid_queue config for 1.x**

## Acceptance testing


app/bin/jobs
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env ruby
+
+require_relative "../config/environment"
+require "solid_queue/cli"
+
+SolidQueue::Cli.start(ARGV)
```
iannorriswork, line 6: interesting! but does that change the solid_queue start command we should be using? is bin/rails solid_queue:start no longer appropriate?

iannorriswork (approved):
looks pretty good, just one curiosity question.

app/bin/jobs
tdooner: Either work, the binstub is a new synonym!

PR #670, opened 2025-05-12
add testing routing to see if we can isolate who is responsible for the issue

iannorriswork:
## Changes

The performance under certain load is highly questionable--and we're having a hard time isolating where the issue is. This exposes a new endpoint that should assist in determining this--this will give us a newrelic query with A) exactly two queries, one to solid queue and one without and B) using a much simpler solidqueue query insert than what EventTrackingJob currently does. 

If this performs better than our current setup, that means we need to rethink how we queue more complex jobs into solid queue.
If this performs equally to our current setup OR the performance is better but the throughput does not increase, I think this increases my suspicion that something's messed up with the load balancer. 

- [ X] Acceptance testing after merge
 * we will remove this route after a round of testing in demo.

### DIFF
--- a/app/app/controllers/health_check_controller.rb
+++ b/app/app/controllers/health_check_controller.rb
@@ -5,6 +5,12 @@ class HealthCheckController < ActionController::Base
     render json: { status: "ok", version: ENV["IMAGE_TAG"] }
   end
 
+  def test_ok
+    # adding a second database query to distinguish healthiness of using one query vs another
+    PerformanceTestingJob.perform_later(Random.new.rand(1..CbvFlow.last.id))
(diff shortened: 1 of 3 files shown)


app/app/controllers/health_check_controller.rb
```diff
@@ -5,6 +5,13 @@ def ok
     render json: { status: "ok", version: ENV["IMAGE_TAG"] }
   end
 
+  def test_ok
```
tdooner, line 8: better name, like `def test_queueing` ?

tdooner (approved):
Just chatted. Looks good as a method of testing.

PR #674, opened 2025-05-13
FFS-2828: Added paystubs_days_since_last_pay_date metric

GeorgeCodes19:
Resolves [FFS-2828](https://jiraent.cms.gov/browse/FFS-2828).

## Changes
- Added a new metric `paystubs_days_since_last_pay_date` which is the difference between the latest pay date of all of the applicant's paystubs and the date the report was generated.

## Acceptance testing

### DIFF
--- a/app/app/controllers/webhooks/argyle/events_controller.rb
+++ b/app/app/controllers/webhooks/argyle/events_controller.rb
@@ -222,7 +222,7 @@ class Webhooks::Argyle::EventsController < ApplicationController
         paystubs_earnings_type_bonus_count: report.paystubs.sum { |p| p.earnings.count { |e| e.category == "bonus" } },
         paystubs_earnings_type_overtime_count: report.paystubs.sum { |p| p.earnings.count { |e| e.category == "overtime" } },
         paystubs_earnings_type_commission_count: report.paystubs.sum { |p| p.earnings.count { |e| e.category == "commission" } },
-        paystubs_days_since_last_paid: report.paystubs.map { |p| Date.parse(p.pay_date) }.compact.max&.then { |last_pay_date| (Date.current - last_pay_date).to_i },
+        paystubs_days_since_last_pay_date: report.paystubs.map { |p| Date.parse(p.pay_date) }.compact.max&.then { |last_pay_date| (Date.current - last_pay_date).to_i },
 
         # Employment fields (originally from "identities" endpoint)
         employment_success: payroll_account.job_succeeded?("employment"),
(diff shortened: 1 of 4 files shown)


app/app/controllers/webhooks/argyle/events_controller.rb
```diff
@@ -234,6 +234,7 @@ def log_sync_finish(payroll_account, report)
         paystubs_gross_pay_amounts_median: paystub_gross_pay_amounts[paystub_gross_pay_amounts.length / 2],
         paystubs_gross_pay_amounts_average: paystub_gross_pay_amounts.sum.to_f / paystub_gross_pay_amounts.length,
         paystubs_gross_pay_amounts_min: paystub_gross_pay_amounts.min,
+        paystubs_days_since_last_pay_date: report.paystubs.map { |p| Date.parse(p.pay_date) }.compact.max&.then { |last_pay_date| (Date.current - last_pay_date).to_i },
```
tdooner, line 237: Maybe the paystubs are already sorted by pay date? Might be worth verifying. Then maybe this could be simplified like:

```suggestion
        paystubs_days_since_last_pay_date: report.paystubs.last.then { |last_paystub| (Date.current - Date.parse(last_paystub.pay_date)).to_i if last_paystub },
```

app/spec/controllers/webhooks/argyle/events_controller_spec.rb
```diff
@@ -104,6 +104,10 @@ def process_webhook(event_type, variant: :connecting)
     end
 
     before do
+      # The report metric, "paystubs_days_since_last_paid" will use Timecop
+      # for a static date to reference as "now". This allows for consistent test results
+      Timecop.freeze(Time.local(2025, 5, 15))
```
tdooner, line 109: This works. I slightly prefer the `around` syntax just to keep all the time mocking in one place, but it's a stylistic change:

```ruby
around do |ex|
  Timecop.freeze(Time.local(2025, 5, 15), &ex)
end
```

(this replaces the need to have `Timecop.return` in the `after` block)

app/spec/controllers/webhooks/pinwheel/events_controller_spec.rb
```diff
@@ -130,7 +138,7 @@
               invitation_id: cbv_flow.cbv_flow_invitation_id,
               client_agency_id: "sandbox",
               pinwheel_environment: "sandbox",
-              sync_duration_seconds: within(1.second).of(5.minutes),
+              sync_duration_seconds: be_a(Numeric),
```
tdooner, line 141: Ah I see what you were saying with the `5.minutes.ago` taking affect before the Timecop is stubbed. I think this is fine.

tdooner (approved):

PR #675, opened 2025-05-13
FFS-2870: Terraform configuration to support LA launch

daphnegold:
## Ticket

Resolves [FFS-2870](https://jiraent.cms.gov/browse/FFS-2870).

## Changes
- Add `results.sarif` from Checkov to our `.gitignore`
- Environment specific Aurora RDS min and max scaling to be set by Terraform config
- Performance insights to stop Checkov from complaining
- Preserve `enable_http_endpoint` functionality that was set via AWS console
- Environment specific SolidQueue worker num can be set via Terraform config

## Context
This hasn't been applied to dev or prod, but I did do a `terraform plan` to confirm the changes are what we expect.

## Acceptance testing

### DIFF
--- a/infra/modules/database/main.tf
+++ b/infra/modules/database/main.tf
@@ -34,6 +34,7 @@ resource "aws_rds_cluster" "db" {
   storage_encrypted           = true
   kms_key_id                  = aws_kms_key.db.arn
   allow_major_version_upgrade = false
+  enable_http_endpoint        = true
 
   db_cluster_parameter_group_name = aws_rds_cluster_parameter_group.rds_query_logging.name
 
@@ -49,8 +50,8 @@ resource "aws_rds_cluster" "db" {
   deletion_protection = !var.is_temporary
 
   serverlessv2_scaling_configuration {
-    max_capacity = 1.0
-    min_capacity = 0.5
+    max_capacity = var.serverless_max_capacity
+    min_capacity = var.serverless_min_capacity
   }
 
   db_subnet_group_name   = var.database_subnet_group_name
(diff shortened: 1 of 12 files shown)


infra/modules/database/main.tf
```diff
@@ -34,6 +34,7 @@ resource "aws_rds_cluster" "db" {
   storage_encrypted           = true
   kms_key_id                  = aws_kms_key.db.arn
   allow_major_version_upgrade = false
+  enable_http_endpoint        = true
```
daphnegold, line 37: Do we want this? I added it to our Terraform to stop it being reverted. At some point, it was set via the AWS console, I believe.

infra/modules/service/async_workers.tf
```diff
@@ -7,10 +6,6 @@ resource "aws_ecs_service" "solid_queue" {
   desired_count          = var.desired_solidqueue_instance_count
   enable_execute_command = var.enable_command_execution ? true : null
 
-  lifecycle {
```
daphnegold, line 10: I'm guessing we added this at some point so we could control the number of instances from the AWS console without Terraform intervening, but that's an anti-pattern kinda, and we should let Terraform manage this for both our app and workers.

infra/modules/service/main.tf
```diff
@@ -46,12 +46,6 @@ resource "aws_ecs_service" "app" {
   desired_count          = var.desired_instance_count
   enable_execute_command = var.enable_command_execution ? true : null
 
-  # Allow changes to the desired_count without differences in terraform plan.
-  # This allows autoscaling to manage the desired count for us.
```
daphnegold, line 50: We don't have autoscaling configured.

daphnegold, line 50: We don't have autoscaling configured. :P It seems like we somehow pulled this in from the template-infra without the autoscaling implementation?

infra/modules/database/main.tf
iannorriswork: @tdooner ? i don't think we set this during me/daphne's round of testing, so if you want that in

iannorriswork (approved):
Looks good to me!

We probably need to do followup with the infra team around the main.tf changes, since that could conflict later if they make changes to service/main.tf, but seems very reasonable for this go around.

daphnegold:
I ran our service and database `make` update commands against dev and prod, worked great. Checked results are as expected in console for both environments and that app is still up and running.

infra/modules/database/main.tf
daphnegold: We can turn this off whenever, so just leaving it on for now! :)

PR #678, opened 2025-05-14
Increase RAILS_MAX_THREADS from 5 -> 10

tdooner:
## Ticket

Related to [FFS-2873](https://jiraent.cms.gov/browse/FFS-2873).

## Changes

* **Increase RAILS_MAX_THREADS from 5 -> 10**

## Context for reviewers

In deployed environments, having 10 threads allows us to serve much more
load reliably.

See - https://cmsgov.slack.com/archives/C08BBUALWQN/p1747185382465289?thread_ts=1747086327.646649&cid=C08BBUALWQN

## Acceptance testing

### DIFF
--- a/infra/app/app-config/env-config/environment-variables.tf
+++ b/infra/app/app-config/env-config/environment-variables.tf
@@ -3,9 +3,10 @@ locals {
   # This is a map rather than a list so that variables can be easily
   # overridden per environment using terraform's `merge` function
   default_extra_environment_variables = {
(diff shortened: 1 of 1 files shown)


daphnegold (approved):
Nice, just FYI these environment variables do not show up and therefore can't be configured from the task definition JSON, but they are available to the container. If I remember correctly. I have no problem modifying the Terraform and applying, just saying.

PR #679, opened 2025-05-14
FFS-2861: Implement Report Validation Changes for LA Launch

PatriciaPerozoUSDS:
FFS-2861: Implement Report Validation Changes for LA Launch
## Ticket

Resolves [FFS-2861](https://jiraent.cms.gov/browse/FFS-2861).

## Changes
Updates the report validation with the following changes:
- removes the hours check for all reports
- defines a "not active employment" concept 
- removes paystub level validation for people no longer actively employed. The goal is to be permissive so as to allow easy verification of termination.

## Context for reviewers
 - I rewrote the boolean logic approx 20 times. This was the clearest way I could get the right validations through.
 - There could be more edge case testing in the unit tests. Please call out any additional cases you recommend I write.

## Acceptance testing

  * NOTE: MY LOCAL DEV ENV IS NOT WORKING. NEED TO PAIR WITH REVIEWER TO GET THIS STOOD UP ON THEIRS BEFORE ACCEPTANCE TESTING.

### DIFF
--- a/app/app/services/aggregators/validators/useful_report_validator.rb
+++ b/app/app/services/aggregators/validators/useful_report_validator.rb
@@ -8,15 +8,12 @@ module Aggregators::Validators
 
       report.errors.add(:employments, "No employments present") unless report.employments.present?
       report.employments.each { |e| validate_employment(report, e) }
-      report.employments.each do |employment|
-        employment_paystubs = paystubs_for_account(report, employment.account_id)
-        # If not actively employed we don't want to run paystub level validations.
-        # Its important that we share any confirmation of loss of employment with the state agency
-        unless is_unemployed(employment, employment_paystubs)
-          is_w2_worker = employment.employment_type == :w2
-          if report.paystubs.any?
-            validate_paystubs(report, is_w2_worker, employment_paystubs)
-          end
+      # If not actively employed we don't want to run paystub level validations.
+      # Its important that we share any confirmation of loss of employment with the state agency
+      unless is_unemployed(report)
+        is_w2_worker = report.employments.none? { |e| e.employment_type == :gig }
+        if report.paystubs.any?
+          validate_paystubs(report, is_w2_worker)
         end
       end
     end
@@ -27,20 +24,18 @@ module Aggregators::Validators
     # Not actively employed means status is not 'employed' OR
     # termination_date is not empty OR
     # The acccount has empty employment status, no termination date, no gross pay and no hours
-    def is_unemployed(employment, employment_paystubs)
-      # assume they're employed
-      unemployment_status = false
-      # TODO turn employment status into a constant or a symbol
-      # only want to set employment status to true don't want to have it set to true and then unset later
-      unemployment_status = true if employment.status.present? && employment.status != "employed"
-      unemployment_status = true if employment.termination_date.present?
-      hours_total = 0
-      employment_paystubs.each { |p| hours_total += p.hours.to_f if p.hours.present? }
-
-      if hours_total == 0 && employment_paystubs.sum { |paystub| paystub.gross_pay_amount.to_f } == 0
-        unemployment_status = true
+    def is_unemployed(report)
+      report.employments.each do |employment|
+        employment_paystubs = paystubs_for_account(report, employment.account_id)
+        # TODO turn employment status into a constant or a symbol
+        return true if employment.status.present? && employment.status != "employed"
+        return true if employment.termination_date.present?
+        hours_total = employment_paystubs.filter { | p| p.hours.present? }.sum { |p| p.hours.to_f }
+        gross_pay_total = employment_paystubs.sum { |paystub| paystub.gross_pay_amount.to_f }
+        return true if hours_total == 0 && gross_pay_total == 0
       end
-      unemployment_status
+      # We know the person is employed if we hit this point
+      false
     end
 
     def paystubs_for_account(report, account_id)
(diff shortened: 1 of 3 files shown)


app/spec/controllers/cbv/payment_details_controller_spec.rb
```diff
@@ -197,10 +197,11 @@
         pinwheel_stub_request_end_user_no_hours_response
       end
 
-      it "redirects to the synchronization failure page" do
-        get :show, params: { user: { account_id: account_id } }
-        expect(response).to redirect_to(cbv_flow_synchronization_failures_path)
-      end
+      # Removing hours check for LA launch - FFS-2866 ticket to add back logic for SNAP only pilots
```
jeffcatania-usds, line 200: instead of commenting out, I would recommend using xit to skip the test.

app/app/services/aggregators/validators/useful_report_validator.rb
```diff
@@ -8,15 +8,45 @@ def validate(report)
...
+      unemployment_status = true if employment.status.present? && employment.status != "employed"
+      unemployment_status = true if employment.termination_date.present?
+      hours_total = 0
+      employment_paystubs.each { |p| hours_total += p.hours.to_f if p.hours.present? }
+
+      if hours_total == 0 && employment_paystubs.sum { |paystub| paystub.gross_pay_amount.to_f } == 0
+        unemployment_status = true
+      end
+      unemployment_status
+    end
```
jeffcatania-usds, line 44:
```suggestion
    def is_unemployed(employment, employment_paystubs)
      # TODO turn employment status into a constant or a symbol
      return false unless employment.status.present? && employment.status != "employed"
      return false unless employment.termination_date.present?
      hours_total = employment_paystubs.filter { | p| p.hours.present? }.sum { |p| p.hours.to_f }
     gross_pay_total = employment_paystubs.sum { |paystub| paystub.gross_pay_amount.to_f } 

      return hours_total == 0 && gross_pay_total == 0
    end
```

jeffcatania-usds: Suggested refactor to make it easier to read.

jeffcatania-usds (changes requested):
LGTM.

app/spec/services/aggregators/validators/useful_report_validator_spec.rb
```diff
@@ -89,6 +156,22 @@
...
+        pay_period_end: nil,
+        pay_date: nil,
+        deducßions: [],
+        hours_by_earning_category: {},
+        hours: 0.0
+      )
+    end
+
+
     let(:invalid_paystub) do
```
jeffcatania-usds, line 175: this should no longer be called invalid_paystub

app/spec/services/aggregators/validators/useful_report_validator_spec.rb
```diff
@@ -89,6 +156,22 @@
...
+    let(:empty_paystub) do
+      Aggregators::ResponseObjects::Paystub.new(
+        account_id: "123",
+        gross_pay_amount: 0.0,
+        net_pay_amount: 0.0,
+        gross_pay_ytd: 0.0,
+        pay_period_start: nil,
+        pay_period_end: nil,
+        pay_date: nil,
+        deducßions: [],
```
jeffcatania-usds, line 168: das macht Spaß.

app/spec/services/aggregators/validators/useful_report_validator_spec.rb
```diff
@@ -298,5 +382,61 @@
         expect(report.errors).to be_empty
       end
     end
+
+    context 'with fully terminated employment' do
+      let(:identities) { [ valid_identity ] }
+      let(:employments) { [ fully_terminated_employment ] }
+      let(:paystubs) { [ invalid_paystub ] }
+
+      it 'is valid even with invalid paystubs' do
```
jeffcatania-usds, line 391: this test does not make sense to me.  Why would it be valid with invalid_paystub?  The issue here is that the invalid_paystub is now a valid paystub.

jeffcatania-usds: On second pass, this does make sense because it is testing the terminated employment case.

app/spec/services/aggregators/validators/useful_report_validator_spec.rb
jeffcatania-usds: I see this is actually an invalid_paystub still because pay_date = nil.  I am realizing naming testing elements "valid" and "invalid" is not great as the underlying requirements of what is valid or invalid changes.

jeffcatania-usds (changes requested):

app/spec/services/aggregators/validators/useful_report_validator_spec.rb
PatriciaPerozoUSDS: It still is an invalid paystub because it doesn't have a pay date

PatriciaPerozoUSDS:
Thanks Jeff - will up up another commit with the new changes

app/app/services/aggregators/validators/useful_report_validator.rb
PatriciaPerozoUSDS: I like the idea of your refactor but I think that would have us break too early. One of the cases David, Eryn, and I discussed, is a job where employment status is employed but everything is empty and they have no paystubs. We would want that to hit the `hours_total == 0 && gross_pay_total == 0` case and be marked as terminated and therefore have laxer validation. 

Tl;dr we can return early if we know the person is unemployed, we can't return early if we think the person is employed. I took your refactor and made some clearer changes to my code that I'll push in a minute.

jeffcatania-usds: You are right! I realize that I got mixed up in the inverses. This is a good time to mention I struggled with distinguishing my left from my right growing up.  I like the new changes!

app/spec/services/aggregators/validators/useful_report_validator_spec.rb
```diff
@@ -11,6 +11,10 @@
   let(:gigs) { [] }
 
   before do
+    # This is trying to isolate this test so that it only handles this unit of code
+    # WARNING!!! If you print the report in the debugger the underlying properties will be empty because
+    # we're overwriting the response to each of the properties of the object via these calls.
+    # TODO WE SHOULD CLEAN THIS UP.
```
tdooner, line 17: Sorry to hear you got bit by the rough edges here of these stubs. Comments like this are more of clutter than helpful, in my opinion, though, since the next person to get bit by it will probably not read this comment (or the comment will fall out-of-date with its surroundings). And it just makes this already large test file even bigger. But this is just an aesthetic preference, so not a blocker.

tdooner (approved):
Looks good to me, although this could be cleaner. Given our deadline, and that this file is pretty self-contained, it seems fine to me. to ship

I couldn't find an Argyle sandbox user that had the specific combination of data that we're trying to allow with this change, so I tested it locally with other users to verify existing behavior:
* Argyle Bob (✅ Verified that sync works)
* Argyle Joe (✅  Verified that sync works, although for some reason it didn't give me back multiple employments)
* John Loanseeker (✅  Verified that it still allows an empty report)
* Ann Loanseeker (✅ Verified it's still a /synchronization_failure)

PR #681, opened 2025-05-14
FFS-2868: Fix three metric bugs

tdooner:
## Ticket

Resolves [FFS-2868](https://jiraent.cms.gov/browse/FFS-2868).

## Changes

**Fix sending ApplicantSharedIncomeSummary when transmitting**
> The `event_logger` object was not defined (nor was `request`). The
> exception was being silenced. And, we removed the test of it during the
> refactor.

After the fix, here is an example of this event sending again:
image

**Fix ArgyleSuccess event to not set `user_id` attribute**
> This attribute is special, it is used by the MixpanelEventTracker to
> associate an event with a caseworker's session of that ID.

> Let's rename this to `argyle_user_id` so it doesn't collide there.

After the fix, here is an example of this event sending with the proper `distinct_id` (not caseworker-[uuid]):

image

**Remove duplicate event for ApplicantCreatedArgyleAccount**
> This event is being tracked twice for Argyle - once by the modal's user
> action and once when receiving the webhook.

> Let's simplify the EventsController by deduplicating in favor of the
> Argyle modal event, since the Argyle modal event has more useful
> attributes.

After the fix, here is an example showing only a single ApplicantCreatedArgyleAccount:
image

## Context for reviewers

[This Mixpanel user](https://mixpanel.com/project/3512948/view/4014736/app/profile#distinct_id=applicant-48) is an example of going through the flow taking into account these fixes.

## Acceptance testing

### DIFF
--- a/app/spec/jobs/case_worker_transmitter_job_spec.rb
+++ b/app/spec/jobs/case_worker_transmitter_job_spec.rb
@@ -1,5 +1,6 @@
 require 'rails_helper'
 require 'csv'
+
 RSpec.describe CaseWorkerTransmitterJob, type: :job do
   include PinwheelApiHelper
   include_context "gpg_setup"
@@ -10,6 +11,7 @@ RSpec.describe CaseWorkerTransmitterJob, type: :job do
   let(:errored_jobs) { [] }
   let(:current_time) { Date.parse('2024-06-18') }
   let(:pinwheel_report) { build(:pinwheel_report, :with_pinwheel_account) }
+  let(:fake_event_logger) { instance_double(GenericEventTracker, track: nil) }
 
   let(:cbv_flow) do
     create(:cbv_flow,
@@ -44,6 +46,10 @@ RSpec.describe CaseWorkerTransmitterJob, type: :job do
     allow(mock_client_agency).to receive(:id).and_return(mocked_client_id)
     allow(mock_client_agency).to receive(:transmission_method).and_return(transmission_method)
     allow(mock_client_agency).to receive(:transmission_method_configuration).and_return(transmission_method_configuration)
+
+    allow_any_instance_of(described_class)
+      .to receive(:event_logger)
+      .and_return(fake_event_logger)
   end
 
   context "#transmit_to_caseworker" do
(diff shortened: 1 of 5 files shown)


millerti:
Thank you for the excellent documentation!

app/spec/jobs/case_worker_transmitter_job_spec.rb
```diff
@@ -117,8 +123,14 @@
         expect(email.body.encoded).to include(cbv_flow.cbv_applicant.case_number)
       end
 
-      # Note that we are not testing events here because doing so requires use of expect_any_instance_of,
-      # which does not play nice since there are multiple instances of the event logger.
```
millerti, line 121: So it seems like we are stubbing away the entire event logger and just checking that for the events now, whereas before we were allowing the controllers to create any number of them and asking if any such object had received the event. Seems like a great way around this problem. Is this a pattern we should apply to all our other event rspec tests (eventually)?

millerti:
Unfortunately, I'm still not sending Mixpanel events on my local machine, so it's going to be difficult for me to verify this locally. I'd suggest getting someone else to do that verification, or maybe relegating it to AC testing. 

Otherwise, this LGTM.

millerti (approved):

GeorgeCodes19:
> Unfortunately, I'm still not sending Mixpanel events on my local machine, so it's going to be difficult for me to verify this locally. I'd suggest getting someone else to do that verification, or maybe relegating it to AC testing.
> 
> Otherwise, this LGTM.

I was able to test this and confirm that things are being properly reported in Mixpanel 👍🏾

app/spec/jobs/case_worker_transmitter_job_spec.rb
tdooner: It's not perfect though, because it wouldn't have caught the last problem where `event_logger` was actually just not defined in that class. I do think this is slightly more controllable than stubbing any instance of GenericEventTracker, since there are a lot of rspec gotchas when trying to make assertions on calls when you've stubbed `any_instance_of`.

The main thing we should be trying to test is whether all the event attributes are computed correctly (and that the event is tracked at the right time), so I'm glad that this does that.

tdooner:
Synced up with Eryn and she's going to acceptance test in demo.

PR #685, opened 2025-05-15
[FFS-2863] set up telemetry for newrelic solid queue

iannorriswork:
## Ticket

Resolves [FFS-2863](https://jiraent.cms.gov/browse/FFS-2863).

## Changes
surprisingly, there isnt a way to plug into activejob directly for this--> they only support telemetry plugins around retrying jobs over seeing them failed.

## Acceptance testing

- [ X ] Acceptance testing after merge
  * Check that newrelic is reporting the new metric by deliberately enqueing a failed job


app/app/jobs/application_job.rb
```diff
@@ -4,4 +4,23 @@ class ApplicationJob < ActiveJob::Base
 
   # Most jobs are safe to ignore if the underlying records are no longer available
   # discard_on ActiveJob::DeserializationError
+  rescue_from(Exception) do |error|
```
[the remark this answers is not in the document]
jeffcatania-usds: I'm also not understanding how this block gets triggered only under an exception.

app/config/initializers/newrelic_solidqueue.rb
```diff
@@ -0,0 +1,23 @@
+require "active_support/notifications"
+
+require "active_support/notifications"
+
+# Subscribe to when a job fails
+ActiveSupport::Notifications.subscribe("solid_queue.job_failed") do |name, start, finish, id, payload|
```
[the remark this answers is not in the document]
iannorriswork: woops, let me yank that. thought i uncommitted this file..

app/app/jobs/application_job.rb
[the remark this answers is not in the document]
iannorriswork: https://edgeguides.rubyonrails.org/active_job_basics.html#error-reporting-on-jobs

jeffcatania-usds (approved):

jeffcatania-usds:
@iannorriswork, it was decided at standup today to not put this into the LA release.

app/app/jobs/application_job.rb
```diff
@@ -1,11 +1,27 @@
...
-  # discard_on ActiveJob::DeserializationError
-
   def event_logger
     @event_logger ||= GenericEventTracker.new
   end
+
+  # Uses https://edgeguides.rubyonrails.org/active_job_basics.html#error-reporting-on-jobs as a pattern
+  # in order to send information to newrelic that we had a failed job and enable alerting on said failed job.
+  rescue_from(Exception) do |error|
+    ActiveSupport::Notifications.subscribe("") do |name, start, finish, id, payload|
```
tdooner, line 9: This looks weird to me - why are we creating a notification subscriber within an exception handler? Shouldn't it be calling `subscribe` with the name of an event to subscribe to?

Is this really the only way to get the `finish` timestamp? It's not a property of `job` or something else accessible within the handler?

tdooner (changes requested):

app/app/jobs/application_job.rb
tdooner: If this is really intentional, I think a comment explaining it would be helpful. I would prefer ditching the event subscription if its only purpose is to get the `finish` timestamp.

iannorriswork: woops, let me remove the subscription bit --> i think that got in from a bad rebase

app/app/jobs/application_job.rb
```diff
@@ -6,22 +6,18 @@ def event_logger
   # Uses https://edgeguides.rubyonrails.org/active_job_basics.html#error-reporting-on-jobs as a pattern
   # in order to send information to newrelic that we had a failed job and enable alerting on said failed job.
   rescue_from(Exception) do |error|
-    ActiveSupport::Notifications.subscribe("") do |name, start, finish, id, payload|
-      job = self.class
-      queue_name = job.queue_name || "default"
-      job_class = job.job_class || "UnknownJob"
+    puts "what the cluck"
```
tdooner, line 9: This still looks like it was included by accident.

FeminismIsAwesome: got rid of the activesupport, forgot a puts. fun times.

PR #687, opened 2025-05-15
Do not remap argyle's "inactive" unemployment status

jeffcatania-usds:
## Ticket

Resolves Issue:

Employment Status was marked as “Furloughed”, even though she is unemployed.
* **Cause**: This is due to how we are normalizing the Argyle employment status ( active / inactive / terminated ) to Pinwheel’s definitions ( employed / furloughed / terminated). In particular, the issue here is that Argyle’s inactive is being translated to “furloughed”. According to Argyle, “inactive: typically indicates the employee left voluntarily, is on leave, or has not been active for a certain period of time.” This definition of inactive does not fully match the situation of being RIF’d, but “inactive” is still better than “furloughed”. 
* **Suggested resolution**: do not remap Argyle’s inactive as furloughed.

## Changes

* Un-maps Argyle's inactive definition to pinwheel's "furloughed".  It will now show up in the report as "inactive".

## Acceptance testing

### DIFF
--- a/app/app/services/aggregators/format_methods/argyle.rb
+++ b/app/app/services/aggregators/format_methods/argyle.rb
@@ -7,8 +7,6 @@ module Aggregators::FormatMethods::Argyle
     case employment_status
     when "active"
       "employed"
-    when "inactive"
-      "furloughed"
     else
       employment_status
     end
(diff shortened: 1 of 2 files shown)


iannorriswork (approved):
LGTM

PR #688, opened 2025-05-15
FFS-2690: Send a report weekly to LA stakeholders

daphnegold:
## Ticket

Resolves [FFS-2690](https://jiraent.cms.gov/browse/FFS-2690).

## Changes
- Weekly report config for LA LDH
- Parameter store config for email recipients in both dev and prod
- Test

## Context for reviewers
I ran the rake task locally just to see it did what we expected without errors.

## Acceptance testing

Screenshot 2025-05-15 at 4 14 46 PM
Screenshot 2025-05-15 at 4 02 04 PM


app/app/mailers/weekly_report_mailer.rb
```diff
@@ -57,6 +57,10 @@ def weekly_report_data(current_agency, report_range)
             agency_id_number: applicant.agency_id_number,
             beacon_id: applicant.beacon_id
           )
+        when "la_ldh"
+          base_fields.merge(
+            case_number: applicant.case_number,
```
daphnegold, line 62: I only merged case number here based on what is present in the `CbvApplicant` model for LA LDH. I figured we wouldn't want to add in date of birth for security reasons.

GeorgeCodes19 (approved):
Looks great. I've also confirmed that AWS contains the proper env vars.

app/app/mailers/weekly_report_mailer.rb
```diff
@@ -41,21 +41,27 @@ def weekly_report_data(current_agency, report_range)
...
+            snap_application_date: applicant.snap_application_date,
+            email_address: invitation.email_address
           )
         when "ma"
           base_fields.merge(
             agency_id_number: applicant.agency_id_number,
-            beacon_id: applicant.beacon_id
+            beacon_id: applicant.beacon_id,
+            snap_application_date: applicant.snap_application_date,
+            email_address: invitation.email_address
```
millerti, line 60: Similar question as above -- do we ever show anything for this?

app/lib/tasks/weekly_reports.rake
```diff
@@ -1,5 +1,5 @@
 namespace :weekly_reports do
-  desc "Send weekly reports (NYC and MA)"
+  desc "Send weekly reports (NYC, MA, and LA)"
```
millerti, line 2: I think we can get rid of MA support since that pilot is over.

app/spec/mailers/weekly_report_mailer_spec.rb
```diff
@@ -140,4 +140,21 @@
...
+
+  context "for the LA LDH client agency" do
+    let(:client_agency_id) { "la_ldh" }
+
+    it "renders the CSV data with LA-specific columns" do
+      expect(mail.attachments.first.filename).to eq("weekly_report_20240902-20240908.csv")
+      expect(mail.attachments.first.content_type).to start_with('text/csv')
+
+      expect(parsed_csv[0]).to match(
+        "case_number" => cbv_flow_invitation.cbv_applicant.case_number,
```
millerti, line 152: Similar question as above -- should we reconstruct this test to not depend on the presence of an invitation?

app/lib/tasks/weekly_reports.rake
daphnegold: How do you feel about making tech debt follow-up tickets to deal with post-pilot cleanup?

app/app/mailers/weekly_report_mailer.rb
tdooner: Yes - for MA that invitation field would have been provided by the caseworker

app/lib/tasks/weekly_reports.rake
daphnegold: @tdooner Says we already have one, woo!

PR #689, opened 2025-05-16
FFS-2690: Send a report weekly to LA stakeholders

daphnegold:
## Ticket

Resolves [FFS-2690](https://jiraent.cms.gov/browse/FFS-2690).

## Changes
Now that we have the generic links option which LA LDH uses, we should not generate weekly flows based on `CbvFlowInvitation` query, use `CbvFlow` instead and don't reference invitations at all for LA LDH

## Acceptance testing

### DIFF
--- a/app/app/mailers/weekly_report_mailer.rb
+++ b/app/app/mailers/weekly_report_mailer.rb
@@ -39,7 +39,6 @@ class WeeklyReportMailer < ApplicationMailer
         applicant = invitation.cbv_applicant
 
         base_fields = {
-          invited_at: invitation.created_at,
           transmitted_at: cbv_flow&.transmitted_at,
           completed_at: cbv_flow&.consented_to_authorized_use_at
         }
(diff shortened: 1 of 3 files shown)


millerti:
Seems like the mailer tests at http://localhost:3000/rails/mailers/ would be a great way to test this, but this one is broken (probably not your fault -- they've been in disrepair lately). Would you consider fixing it? Or is there a better way for me to test?

millerti (approved):
Given we plan to fix the mailer test page later, and given we AC test by waiting for the email, I'm ok with merging this.

app/app/mailers/weekly_report_mailer.rb
```diff
@@ -30,41 +30,47 @@ def report_filename(report_range)
...
+    flows = CbvFlow.where(client_agency_id: client_agency_id)
+                   .includes(:cbv_applicant, :cbv_flow_invitation)
+
+    results = flows.map do |flow|
+      applicant = flow.cbv_applicant
+      invitation = flow.cbv_flow_invitation
+
+      next if invitation && !report_range.cover?(invitation.created_at)
+
+      base_fields = {
```
tdooner, line 44: Sorry, one more thing. Could we add

```suggestion
      base_fields = {
        started_at: flow.created_at,
```

app/app/mailers/weekly_report_mailer.rb
```diff
@@ -30,41 +30,47 @@ def report_filename(report_range)
...
-          )
-        when "la_ldh"
-          base_fields.merge(
-            case_number: applicant.case_number,
-          )
-        end
+    client_agency_id = current_agency.id
+
+    flows = CbvFlow.where(client_agency_id: client_agency_id)
+                   .includes(:cbv_applicant, :cbv_flow_invitation)
```
tdooner, line 36: Let's filter these down to only the CbvFlow's that were completed. I think that solves the uniqueness problem too, since (at least until we implement https://jiraent.cms.gov/browse/FFS-2664) the invitation link can only be used once.

daphnegold: This is such a smart simple solution. Working on it now and undoing things I regressed in oversimplifying.

tdooner (approved):

PR #699, opened 2025-05-20
FFS-2924: Fix missing_results page view render logic

GeorgeCodes19:
## Ticket

Resolves [FFS-2924](https://jiraent.cms.gov/browse/FFS-2924).

## Changes
Replace the `@has_payroll_account` view variable with `@cbv_flow.has_account_with_required_data?` in the conditional for rendering the `/applicant_information` link. 

## Context for reviewers
During our LA pilot an applicant encountered an [error](https://nava.slack.com/archives/C0792AR4570/p1747748131396709) which is a result of a bug in our view rendering logic whereas the `Continue to review my report` `(/applicant_information)` link was being rendered on the `/missing_results` page even if an applicant does not have a fully synced payroll account. 

The applicant clicked this button despite not having a `fully_synced` payroll account (or any payroll account linked for that matter) which resulted in the error above.

The view logic should first determine if the `cbv_flow` has the required data e.g. at least one fully synced payroll account before rendering the link to the `/applicant_information` page.

## Acceptance testing

### DIFF
--- a/app/app/controllers/cbv/missing_results_controller.rb
+++ b/app/app/controllers/cbv/missing_results_controller.rb
@@ -2,7 +2,6 @@ class Cbv::MissingResultsController < Cbv::BaseController
   before_action :track_missing_results_event, only: :show
 
   def show
-    @has_payroll_account = @cbv_flow.payroll_accounts.any?
   end
 
   def track_missing_results_event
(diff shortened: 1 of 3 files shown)


tdooner (approved):
Thanks for the regression test!

PR #702, opened 2025-05-21
filter a list of employers from search results

millerti:
## Changes
Now, we filter a list of blocked employers from the search results from Argyle and Pinwheel.

## Context for reviewers
See [this thread](https://cmsgov.slack.com/archives/C066XS5543S/p1747861309595409?thread_ts=1747856345.708169&cid=C066XS5543S). We decided to block certain employers who will always send us data in a format we don't expect (and which causes a crash). These employers were chosen because they represent unearned income, which is not what LA wants us to be reporting.

## Testing Instructions
0. Set up to use argyle and pinwheel
1. Search for Social Security Administration, Veteran's Affairs, and Louisiana Workforce Commission (this never showed up anyway for some unknown reason, but we're blocking it at Patricia's request since it falls under a common unearned income category.
- Before, you would have seen matches that looked like these exact queries. Now, they will have been removed. You may still see entries for these as _employers_ (eg, if you earn a paycheck (earned income) from working at one of these institutions), but the entries for them as benefits providers (unearned income) will not appear.

## Acceptance testing

### DIFF
--- a/app/app/services/provider_search_service.rb
+++ b/app/app/services/provider_search_service.rb
@@ -3,8 +3,12 @@ class ProviderSearchService
 
   # item_000136007 - SSA
   # item_000042186 - Veterans Affairs
-  BLOCKED_ARGYLE_EMPLOYERS = %w[item_000136007 item_000042186]
-  BLOCKED_PINWHEEL_EMPLOYERS = %w[]
+  # item_000023897 - Louisiana Workforce Commission, which administers Unemployment Insurance
+  BLOCKED_ARGYLE_EMPLOYERS = %w[item_000136007 item_000042186 item_000023897]
+
+  # 0bab321e-64ea-4047-8d8a-44ba283f6cd8 - Social Security Administration - Retirement, Survivors, Disability benefits
+  # 3f7b6ce9-6371-4bb8-8c2f-051c446dc55c - Veterans Affairs benefits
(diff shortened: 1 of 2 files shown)


app/app/services/provider_search_service.rb
```diff
@@ -27,11 +34,17 @@ def search(query = "")
...
       results = pinwheel_results if results.length == 0 || any_exact_matches?(pinwheel_results, query)
+
+      results = filter_results(results, BLOCKED_PINWHEEL_EMPLOYERS)
     end
 
     results
   end
 
+  def filter_results(results, blocked_employers)
+    results.select { |result| !result.provider_options.provider_id.in? blocked_employers }
```
tdooner, line 45: Here's a slightly more readable alternative:

```suggestion
    results.select { |result| blocked_employers.exclude?(result.provider_options.provider_id) }
```

tdooner (commented):
Seems like a good approach. Tests, of course, would be nice.

danielladevera-nava:
AC testing: 

- Did not find a match for SSA 
- Did not find a match for LA Workforce Commission 
- Did not find an exact match for Veterans Affairs, but two non-exact matches returned: "Disability" and "Retirement." The modals for those options seem to allow the user to connect unearned income sources via SSA or Veterans Affairs 

image
image

danielladevera-nava:
Passing AC now! LGTM

PR #703, opened 2025-05-22
FFS-2572: Our mailer testers are broken

daphnegold:
## Ticket

Resolves [FFS-2572](https://jiraent.cms.gov/browse/FFS-2572).

## Changes
I fixed the date error in the weekly report that was keeping it from loading. Removed some unused nyc/ma code, although that does hide some of our patterns for how to create mailers. I propose we have a ticket to update these every time we take on a new pilot (with all the rest of cleanup). Each agency should have its own clickable button on the mailer page for each case.

## Context for reviewers
Not perfect, but an improvement?

## Acceptance testing


iannorriswork (approved):
lgtm!

PR #704, opened 2025-05-22
apply standard retry for jobs that fail

iannorriswork:
## Changes

Changes jobs to try and rerun on initial failure.

## Context for reviewers

assumptions:

1) your background job should be idempotent (safe to rerun) 2) most of the background jobs will have some level of jitter

in that case, lets apply this to ApplicationJob rather than require each job to specify a default retry strategy.

## Acceptance testing

  * Kick off TestQueueingJob in a way that fails. see it rerun.


iannorriswork:
superseded by https://github.com/DSACMS/iv-cbv-payroll/pull/705

PR #709, opened 2025-05-27
FFS-2664: Allow CBV links to be re-used by other household members

daphnegold:
## Ticket

Resolves [FFS-2664](https://jiraent.cms.gov/browse/FFS-2664).

## Changes
Agency-configurable reuse of invitation links for multi-household

## Acceptance testing

### DIFF
--- a/app/app/controllers/cbv/base_controller.rb
+++ b/app/app/controllers/cbv/base_controller.rb
@@ -18,8 +18,12 @@ class Cbv::BaseController < ApplicationController
         track_expired_event(invitation)
         return redirect_to(cbv_flow_expired_invitation_path(client_agency_id: invitation.client_agency_id))
       end
-      if invitation.complete?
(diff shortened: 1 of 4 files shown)


app/app/controllers/cbv/base_controller.rb
```diff
@@ -18,8 +18,12 @@ def set_cbv_flow
         track_expired_event(invitation)
         return redirect_to(cbv_flow_expired_invitation_path(client_agency_id: invitation.client_agency_id))
       end
-      if invitation.complete?
-        return redirect_to(cbv_flow_expired_invitation_path(client_agency_id: invitation.client_agency_id))
+
+      client_agency = agency_config[invitation.client_agency_id]
```
millerti, line 22: Can you use the application_controller's current_agency method here?

daphnegold: `base_controller` also has a `current_agency` method, but the problem is that it uses the `cbv_flow` which isn't created yet when we are doing this check

PR #713, opened 2025-05-27
Ruby upgrade

iannorriswork:
## Changes

Upgrading the ruby version so that we get base images that are being updated more frequently--the current ruby version's docker image was last updated seven months ago.

according to https://hub.docker.com/_/ruby, “Unless you are working in an environment where only the ruby image will be deployed and you have space constraints, we highly recommend using the default image of this repository.”, so that's why it's changing the docker base image as well.

Confirmed tests run locally and Jeff confirmed locally runs fine.

#### Install version 3.4.4 in rbenv 
Verify that rbenv has 3.4.4 available to install
```
> rbenv install --list | grep 3.4.4

3.4.4
```
Upgrade ruby-build to get the latest version index for rbenv. 
```
> brew update && brew upgrade ruby-build
```
Install Ruby 3.4.4
```
> rbenv install 3.4.4
```
#### Use rbenv's version 3.4.4
Open up a new terminal session. and verify the ruby version.
```
> which ruby
{HOME_DIR}/.rbenv/shims/ruby

> ruby --version
4.4.4

```

If the ruby version does not match, run the rbenv local command.  This should not be needed.
```
> cd {project-repo}/app
> rbenv local 3.4.4
(restart terminal after)
```
Your IDE may need special configuration to use the new Ruby version.  Rubymine asked me whether i wanted to use 3.4.4.

#### Install Ruby packages for new version
Run a bundle install
```
cd {project-repo}/app
bundle install
```

## Acceptance testing


app/Gemfile
```diff
@@ -128,3 +132,5 @@ gem "wkhtmltopdf-binary"
 
 gem "net-sftp"
 gem "net-ssh"
+
+gem "highline", "~> 3.1"
```
tdooner, line 136: Was this included intentionally?

PR #720, opened 2025-05-30
Add script for importing/exporting I18n translation CSVs

tdooner:
## Ticket

N/A - A couple random fixes I pulled off a different branch.

## Changes

- **Fix filename of ClientAgencyConfig spec**
- **Add script for importing/exporting I18n translation CSVs**

## Acceptance testing


tdooner:
> Can you update [our translations process document](https://confluenceent.cms.gov/display/SFIV/Product+Design+Processes+and+Norms) to include how and when to use these scripts? I think that would help us make sure that this is the one true way we do these tasks now.

@millerti Done, added under "When doing a holistic copy review/edit".

tdooner:
@millerti I found this command very useful to produce the translations sheet that April is currently reviewing, and so I'm checking it in for future reference. I haven't tested the import command; I believe it will replace all strings (en + es) with whatever the value is in the imported CSV.

The `csv-export` does not seem to be able to filter strings by whether they are missing, so I don't think it is a replacement for the existing process. I can rename if you have a better name in mind.

GeorgeCodes19 (approved):
I pulled the branch and kicked the tires. LGTM- though I haven't tested existing entries which I suppose is fine since our current workflow involves deleting any non-`en` entries prior to translation. What should we do with the [other translation import script?]( https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/lib/tasks/translations.rake)

millerti:
Played around and satisfied myself. I'm going to punt on what to do with any other script until I need to merge the finished Es strings tomorrow. If this script works fine, I won't do anything, and if it doesn't maybe I'll raise it in Slack.

millerti (approved):

tdooner:
Thanks. Merging!

PR #721, opened 2025-05-30
Track `accountId` and `userId` for Argyle modal events

tdooner:
## Ticket

N/A - Fix to improve investigation capability that would have helped me.

## Changes

* **Track `accountId` and `userId` for Argyle modal events**

## Context for reviewers

When we initially implemented these in FFS-2652, we were treating these
IDs as PII. After implementing that ticket, we redefined our definition
of PII to exclude non-sensitive identifiers such as these.

Since these will greatly help us investigate odd connection behavior in
the Argyle modal, let's stop sanitizing them.

## Acceptance testing


jeffcatania-usds (approved):
Good PR.  I just noticed this sanitation method yesterday and forgot why we sanitized these to begin with.

tdooner:
I tested it myself to verify that it does indeed track these properties. Going to merge.

image

PR #722, opened 2025-05-30
Fix two translations on /employer_search to render for AZ DES

tdooner:
## Ticket

N/A - Noticed this while reading through the code.

## Context for reviewers

For AZ, this would break. This commit updates it to use the variables
called for in the content reference.

## Acceptance testing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,13 @@ repos:
     files: '^app/config/locales/.*\.yml$'
     pass_filenames: false
 
+  - id: i18n-interpolations
+    name: i18n-interpolations
+    entry: bin/run-linter --cd app -- bundle exec i18n-tasks check-consistent-interpolations
+    language: ruby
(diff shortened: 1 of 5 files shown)


tdooner:
> This looks fine to me but I'm having a hard time figuring out what you caught in the code that would have broken. Is it that there's no default for two of these strings?

@millerti The "broken" comment referred to [this commit's fix](https://github.com/DSACMS/iv-cbv-payroll/pull/722/commits/e28945f4974c67deba79103a6317415bd93838fc) which hardcoded example.com as the agency contact URL for the "default", which would be used by default for new agencies like AZ.

The `to_continue_li_1_html` string would also break in a different way: since no default was specified for that string, it would have been a missing translation for AZ.

millerti (approved):

millerti:
> @millerti Going to wait on this branch until you merge Daphne's (#724) since there will be conflicts.

Merged #724 so you should be good to merge.

PR #723, opened 2025-05-30
FFS-2667: Multi-earner household share link

GeorgeCodes19:
## Ticket

Resolves [FFS-2667](https://jiraent.cms.gov/browse/FFS-2667).

## Changes

- [x] Add stimulus controller for copying invitation link
- [x] Updated link share input to use agency's production and demo domains
- [x] Add test to assert that the proper links are given
- [x] Link generation for tokenized and generic flow
- [x] Update i18n to add ignored missing Spanish translations
- [x] Track events via Mixpanel and new relic

## Acceptance testing

Screenshot 2025-05-30 at 3 12 00 PM
Screenshot 2025-05-30 at 3 12 09 PM

Screen Shot 2025-05-30 at 16 26 53

Screen Shot 2025-05-30 at 16 26 56


app/app/javascript/controllers/copy_link_controller.js
```diff
@@ -0,0 +1,46 @@
+import { Controller } from "@hotwired/stimulus"
+import { trackUserAction } from "../utilities/api"
+
+export default class extends Controller {
+  static targets = ["copyLinkButton", "input", "successButton"]
+
+  connect() {
+    if(this.hasCopyLinkButtonTarget){
```
GeorgeCodes19, line 8: USWDS has media queries that add margin based on the device screen size so we're resorting to JS to remove those margins. Otherwise things start to look weird.

app/app/views/cbv/successes/show.html.erb
```diff
@@ -1,23 +1,67 @@
...
+              class: "copy-link-button",
+              icon: "content_copy",
+              icon_position: :leading,
+              variant: :primary,
+              type: "button",
+              data: {
+                copy_link_target: "copyLinkButton",
+                action: "click->copy-link#copy"
+              }) %>
+        <%# success button variant %>
```
GeorgeCodes19, line 54: We're using multiple variants rendered server-side in order to support localization for the button with the success variant. I figured this was the path of least resistance as opposed to using Turbo to dynamically render a single button.

app/app/controllers/cbv/successes_controller.rb
```diff
@@ -3,6 +3,23 @@ class Cbv::SuccessesController < Cbv::BaseController
...
   def show
+    @invitation_link = invitation_link
+  end
+
+  private
+
+  def invitation_link
+    domain = Rails.env.production? ? current_agency.agency_production_domain : current_agency.agency_demo_domain
+
+    if %w[az_des sandbox].include?(@cbv_flow.client_agency_id)
```
GeorgeCodes19, line 14: I wonder if we should be storing the link type on the client agency config?

app/app/assets/stylesheets/cbv.scss
```diff
@@ -189,3 +189,33 @@ html {
...
+*/
+input#invitation_link{
+  margin-top: 0;
+  border-right: none;
+  height: 3rem;
+  
+  // Hide visually but keep in DOM for copying on smaller screens
+  position: absolute;
+  left: -9999px;
+  opacity: 0;
```
tdooner, line 204: Is it really necessary for the field to be visible on the page for its value to be used in copying? This seems a bit weird to me.

app/app/assets/stylesheets/cbv.scss
```diff
@@ -189,3 +189,33 @@ html {
...
+}
+.copy-link-button {
+  border-radius: 0 0.25rem 0.25rem 0;
+  border-left: none;
+  height: 3rem;
+}
+
+.usa-button--success,
+.usa-button--success:hover {
+  background-color: #00A91C;
```
tdooner, line 220: I suspect the button should be darker when hovered over. Might be good to clarify with design.

app/app/controllers/cbv/successes_controller.rb
```diff
@@ -3,6 +3,23 @@ class Cbv::SuccessesController < Cbv::BaseController
...
+      if (invitation_link = @cbv_flow.cbv_flow_invitation&.to_url)
+        URI.parse(invitation_link).tap { |uri| uri.host = domain; uri.scheme = "https"; uri.port = nil }.to_s
+      else
+        "https://#{domain}"
+      end
+    else # generate a generic link
+      "https://#{domain}/en/cbv/links/#{@cbv_flow.client_agency_id}"
+    end
```
tdooner, line 22: I think it would be better to do this solely based on the presence of a `@cbv_flow.cbv_flow_invitation`, so it's not based on the agency ID. If the invitation is present, render the Invitation's link. Otherwise, render the generic link.

app/app/controllers/cbv/successes_controller.rb
```diff
@@ -3,6 +3,23 @@ class Cbv::SuccessesController < Cbv::BaseController
...
+  end
+
+  private
+
+  def invitation_link
+    domain = Rails.env.production? ? current_agency.agency_production_domain : current_agency.agency_demo_domain
+
+    if %w[az_des sandbox].include?(@cbv_flow.client_agency_id)
+      if (invitation_link = @cbv_flow.cbv_flow_invitation&.to_url)
+        URI.parse(invitation_link).tap { |uri| uri.host = domain; uri.scheme = "https"; uri.port = nil }.to_s
```
tdooner, line 16: Would be nice to update the `CbvFlowInvitation#to_url` method to use the proper host for the relevant client agency. It would allow you to pass in `host: ...` directly into the Rails URL helper rather than having to parse/regenerate the URL here.

app/app/helpers/uswds_form_builder.rb
```diff
@@ -197,6 +197,54 @@ def yes_no(attribute, options = {})
...
+      icon_path = "#{icon_sprite_path}##{icon_name}"
+      icon_wrapper_class = "usa-button__icon--#{icon_position}"
+
+      icon_svg = @template.content_tag(:svg, class: "usa-icon", "aria-hidden": true, focusable: false, role: "img") do
+        @template.tag.use("", href: icon_path)
+      end
+
+      text_span = @template.content_tag(:span, value, class: "usa-button__text")
+
+      if icon_position == :leading
```
tdooner, line 234: YAGNI? It looks like both usages of this button have a "leading" `icon_position`, so we may not need this option?

app/app/helpers/uswds_form_builder.rb
```diff
@@ -197,6 +197,54 @@ def yes_no(attribute, options = {})
     end
   end
 
+  def button_with_icon(value = "Button", options = {})
+    button_classes = [ "usa-button" ]
+
+    icon_name = options.delete(:icon)
+    icon_position = options.delete(:icon_position) || :leading
+    variant = options.delete(:variant)
+    button_type = options.delete(:type) || "button"
```
tdooner, line 206: YAGNI? It looks like both `button_with_icon`'s have type = "button", so we may not need this option.

app/app/javascript/controllers/copy_link_controller.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+  
+    setTimeout(() => {
+      if(this.hasSuccessButtonTarget) {
+        this.successButtonTarget.classList.add('invisible')
+      }
+      if(this.hasCopyLinkButtonTarget) {
+        this.copyLinkButtonTarget.classList.remove('invisible')
+      }
+    }, 3000)
```
tdooner, line 42: Let's use stimulus lifecycle events to unregister this timeout if the component [disconnects](https://stimulus.hotwired.dev/reference/lifecycle-callbacks).

app/app/views/cbv/successes/show.html.erb
```diff
@@ -1,23 +1,67 @@
...
+    </li>
+  </ul>
+
+  <div class="border border-info-light bg-info-lighter padding-2 margin-top-3 radius-md">
+    <h3 class="margin-top-0"><%= t(".share_invitation_link_title") %></h3>
+    <p><%= t(".share_invitation_link_content") %></p>
+
+    <div class="display-flex flex-align-center" data-controller="copy-link">
+      <%= form_with(builder: UswdsFormBuilder, html: { style: "display: contents;" }) do |f| %>
+        <%= f.label :invitation_link, "Invitation link", class: "usa-sr-only" %>
```
tdooner, line 41: I18n

app/spec/controllers/cbv/successes_controller_spec.rb
```diff
@@ -24,6 +25,48 @@
...
         expect(response.body).to include(cbv_flow.confirmation_code)
       end
+
+      it "shows copy link button" do
+        get :show
+        expect(response.body).to include(I18n.t("cbv.successes.show.copy_link"))
+        expect(response.body).to have_selector('button[data-copy-link-target="copyLinkButton"]')
+      end
+
+      describe "invitation_link" do
```
tdooner, line 35:
```suggestion
      describe "#invitation_link" do
```

tdooner (commented):
Looking good in general. I think the only real blocker I have is the -9999px trick - it seems like there must be a better way. The rest should be little cleanups?

app/app/controllers/cbv/successes_controller.rb
GeorgeCodes19: Wow, you read my mind on this one! Glad that we're on the same page.

app/app/assets/stylesheets/cbv.scss
GeorgeCodes19: This is only for the success state. The issue that I ran into was that my mouse stayed hovered over the button so the green background change wasn't apparant. Design already gave the approval for the PR, but it might be worth running this specific scenario by them again.

app/app/views/cbv/successes/show.html.erb
```diff
@@ -1,23 +1,67 @@
...
+  <h2><%= t(".whats_next") %></h2>
 
-  <div>
-    <%= link_to agency_url, target: "_blank", rel: "noopener noreferrer", class: "usa-button margin-top-3 usa-button usa-button--outline" do %>
-      <%= t(".back_to_agency", agency_short_name: current_agency.agency_short_name) %>
-    <% end %>
+  <ul>
+    <li>
+      <strong><%= t(".whats_next_1_title") %></strong>
+      <ul style="list-style-type: disc;">
```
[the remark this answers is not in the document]
GeorgeCodes19: Unfortunately the `<ul class="usa-list"><li></li> ... </ul>` wasn't applying the styling that the (USWDS list docs)[https://designsystem.digital.gov/components/list/] suggests it does. However, adding the inline style attribute did work. I suppose we could bump this to class so that we can use it elsewhere.

app/app/assets/stylesheets/cbv.scss
GeorgeCodes19: I remember having this conversation before regarding another feature. Maybe it was the help modal related stuff. If I recall, having `display:none` or not having the element visible in the DOM prevented its value from being able to be read by the browser agent. I'll mess around with this some more to verify.

GeorgeCodes19:
> Looking good in general. I think the only real blocker I have is the -9999px trick - it seems like there must be a better way. The rest should be little cleanups?

Agreed. I've gone ahead and cleaned that up as well. We may need to spend some time pairing on my local setup. `display:none` does work, but I wasn't aware because I need to clobber, precompile my assets, and restart the Rails server with each CSS/JS change.

app/config/locales/es.yml
```diff
@@ -247,8 +247,6 @@ es:
         consent_to_authorize_warning: A fin de compartir su informe, marque la casilla de consentimiento legal.
     successes:
       show:
-        back_to_agency: Vaya al sitio web de %{agency_short_name}
-        caseworker_received: Hemos enviado su información de pago a %{agency_acronym}. Ellos lo contactarán si se necesita alguna información adicional.
         confirmation_code_html: "<strong>Código de confirmación </strong>: %{confirmation_code}"
```
millerti, line 250: Looks like this too needs to be removed (you added "your" to the English variant).

GeorgeCodes19:
> In the mobile view, the left two corners of the copy link button should be rounded, but aren't.
> 
> image

Ahh, good eye! There's some missing conditional styling that should be applied to mobile when the input field is no longer visible.

app/app/controllers/cbv/successes_controller.rb
```diff
@@ -3,6 +3,19 @@ class Cbv::SuccessesController < Cbv::BaseController
...
+
+  private
+
+  def invitation_link
+    host = Rails.env.production? ? current_agency.agency_production_domain : current_agency.agency_demo_domain
+
+    if @cbv_flow.cbv_flow_invitation.present?
+      @cbv_flow.cbv_flow_invitation.to_url(host: host, protocol: "https")
+    else # generate a generic link
+      "https://#{host}/en/cbv/links/#{@cbv_flow.client_agency_id}"
```
jeffcatania-usds, line 17: I'm surprised this isn't defined somewhere else in the codebase. This should at least be refactored into its own method.

app/app/helpers/uswds_form_builder.rb
```diff
@@ -197,6 +197,100 @@ def yes_no(attribute, options = {})
...
 
+  def button_with_icon(value = "Button", options = {})
+    if value.is_a?(Hash) && options.empty?
+      options = value
+      value = "Button"
+    end
+
+    local_options = options.dup
+
+    button_classes = [ "usa-button" ]
```
jeffcatania-usds, line 208: please include link to docmentation of the possible variants

jeffcatania-usds (changes requested):

app/app/controllers/cbv/successes_controller.rb
GeorgeCodes19: I've added a [to_generic_url method](https://github.com/DSACMS/iv-cbv-payroll/blob/63fd373831ec0baedd68e925119e7a8530e79f64/app/app/models/cbv_flow.rb#L41) to the `cbv_flow` model. Originally I figured that the `ClientAgencyConfig` would be a well suited place for this logic, but that required passing the desired `client_agency`. The `cbv_flow` model already has knowledge via the `client_agency_id` column/attribute.

app/app/views/cbv/successes/show.html.erb
```diff
@@ -1,23 +1,76 @@
...
 <% content_for :title, title %>
 <div class="grid-col tablet:grid-col-10">
-<h1><%= title %></h1>
-</div>
+  <h1><%= title %></h1>
 
-<div class="usa-prose">
-  <p><%= t(".confirmation_code_html", confirmation_code: @cbv_flow.confirmation_code) %></p>
+  <div class="usa-prose">
+    <p><%= t(".confirmation_code_html", confirmation_code: @cbv_flow.confirmation_code) %></p>
```
jeffcatania-usds, line 7: Please verify with product that this is a default feature for the platform and will go live for both AZ and LA

jeffcatania-usds (approved):

PR #724, opened 2025-05-30
FFS-2648: Implement content updates for AZ DEZ

daphnegold:
## Ticket

Resolves [FFS-2648](https://jiraent.cms.gov/browse/FFS-2648).

## Changes
Updates for /entry and /success to work with AZ_DES

## Context for reviewers
This is a follow up PR to fix a few small things that were caught in acceptance testing

## Acceptance testing

Screenshot 2025-05-30 at 1 46 32 PM
Screenshot 2025-05-30 at 4 28 32 PM

### DIFF
--- a/app/config/locales/en.yml
+++ b/app/config/locales/en.yml
@@ -350,7 +350,7 @@ en:
         what_if_i_cant_use_this_body_1: If you need to report your income and can’t use this tool, then you’ll need to share your income information for this job directly with %{agency_short_name}.
         what_if_i_cant_use_this_body_2_html:
           default: Visit <a href="%{agency_url}" target="_blank" rel="noopener noreferrer">%{agency_acronym}’s website</a> for more information on how to submit documents.
-          la_ldh: 'You can: <ul><li>Upload documents through the Self Service Portal online at <a href="%{agency_url}" target="_blank" rel="noopener noreferrer">%{agency_url}</a></li><li>Send documents through US Mail: Louisiana Medicaid/LaCHIP, P.O. Box 91283, Baton Rouge, LA 70821-9278</li><li>Email documents: <a href="mailto:MyMedicaid@la.gov">MyMedicaid@la.gov</a></li><li>Bring your documents to the Medicaid office of choice.</li></ul>'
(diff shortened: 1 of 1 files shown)


app/app/views/cbv/entries/show.html.erb
```diff
@@ -55,7 +55,7 @@
   <div id="what_if_i_cant_use_this_accordion" class="usa-accordion__content usa-prose">
     <p>
       <%= t(".what_if_i_cant_use_this_body_1", agency_short_name: current_agency.agency_short_name) %>
-      <%= agency_translation(".what_if_i_cant_use_this_body_2_html", agency_url: "https://mymedicaid.la.gov/") %>
+      <%= agency_translation(".what_if_i_cant_use_this_body_2_html", agency_url: current_agency.agency_contact_website, agency_acronym: agency_translation("shared.agency_acronym")) %>
```
millerti, line 58: According to Makaela in early May, Design wanted this link to go to mymedicaid.la.gov instead of the agency_contact_website. Did they change their mind? I can't see whatever conversation happened during AC testing.

app/app/views/cbv/successes/show.html.erb
```diff
@@ -9,7 +9,7 @@
 
   <p><%= t(".caseworker_received", agency_acronym: current_agency.agency_short_name) %></p>
   <p><%= t(".if_it_didnt_work", agency_acronym: current_agency.agency_short_name) %></p>
-  <p><%= t(".follow_up", agency_acronym: current_agency.agency_short_name) %></p>
+  <p><%= t(".follow_up", agency_acronym: current_agency.agency_short_name, benefit: agency_translation("shared.benefit"), reporting_purpose: agency_translation("shared.reporting_purpose"), agency_url: current_agency.agency_contact_website).html_safe %></p>
```
millerti, line 12: I thought we were just hardcoding benefit and reporting_purpose into the strings per state, since they won't change per state. I am also worried because I don't know how this will translate (eg if the setting is "benefits", it'll just put "benefits" into a Spanish string, which is certainly wrong).

tdooner: We clarified during Eng Sync - these will be translated because they're translated before interpolation due to the `agency_translation("shared.benefit")` call.

millerti (approved):

PR #728, opened 2025-06-04
[FFS-2686] introduce measurements for household level analytics

iannorriswork:
## Ticket

Resolves [FFS-2686](https://jiraent.cms.gov/browse/FFS-XXXX).

## Changes

Introduces metrics to count number of household members, number of reports completed, and number of links started. 

## Context for reviewers

This is likely a hack for count_unique_members, but its a starting point before doing further investigation and negotiation with product.

## Acceptance testing
  * Check mixpanel events are firing as expected while walking through as an AZ user, including using the same invitation to do multiple report flows.

  * Check mixpanel events are firing as expected while walking through as an AZ user, including using the same invitation to do multiple report flows.

### DIFF
--- a/app/app/controllers/cbv/base_controller.rb
+++ b/app/app/controllers/cbv/base_controller.rb
@@ -28,6 +28,7 @@ class Cbv::BaseController < ApplicationController
         end
       end
 
+
       @cbv_flow = CbvFlow.create_from_invitation(invitation)
       session[:cbv_flow_id] = @cbv_flow.id
       track_invitation_clicked_event(invitation, @cbv_flow)
@@ -143,9 +144,18 @@ class Cbv::BaseController < ApplicationController
       cbv_flow_id: cbv_flow.id,
       cbv_applicant_id: cbv_flow.cbv_applicant_id,
       client_agency_id: current_agency&.id,
-      seconds_since_invitation: (Time.now - invitation.created_at).to_i
+      seconds_since_invitation: (Time.now - invitation.created_at).to_i,
+      number_household_members:  count_unique_members(invitation, cbv_flow),
+      number_reports_completed: invitation.cbv_flows.where.not(consented_to_authorized_use_at: nil).count,
+      number_links_started: invitation.cbv_flows.count
     })
-  rescue => ex
-    Rails.logger.error "Unable to track event (ApplicantClickedCBVInvitationLink): #{ex}"
+    # rescue => ex
+    # Rails.logger.error "Unable to track event (ApplicantClickedCBVInvitationLink): #{ex}"
+  end
+
+  def count_unique_members(invitation, cbv_flow)
+    return 1 if invitation.cbv_applicant.income_changes.blank?
+
+    invitation.cbv_applicant.income_changes.map { |income_change| income_change.with_indifferent_access[:member_name] }.uniq.count
   end
 end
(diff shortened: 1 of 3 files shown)


app/app/controllers/cbv/base_controller.rb
```diff
@@ -143,9 +143,18 @@ def track_invitation_clicked_event(invitation, cbv_flow)
       cbv_flow_id: cbv_flow.id,
       cbv_applicant_id: cbv_flow.cbv_applicant_id,
       client_agency_id: current_agency&.id,
-      seconds_since_invitation: (Time.now - invitation.created_at).to_i
+      seconds_since_invitation: (Time.now - invitation.created_at).to_i,
+      number_household_members:  count_unique_members(invitation),
+      number_reports_completed: invitation.cbv_flows.where.not(consented_to_authorized_use_at: nil).count,
+      number_links_started: invitation.cbv_flows.count
```
tdooner, line 149: Naming nitpick, we use `_count` suffix in other event attributes, so I would name these:
* `household_member_count`
* `completed_reports_count`
* `flows_started_count`

app/app/controllers/cbv/base_controller.rb
```diff
@@ -143,9 +143,18 @@ def track_invitation_clicked_event(invitation, cbv_flow)
       cbv_flow_id: cbv_flow.id,
       cbv_applicant_id: cbv_flow.cbv_applicant_id,
       client_agency_id: current_agency&.id,
-      seconds_since_invitation: (Time.now - invitation.created_at).to_i
+      seconds_since_invitation: (Time.now - invitation.created_at).to_i,
+      number_household_members:  count_unique_members(invitation),
+      number_reports_completed: invitation.cbv_flows.where.not(consented_to_authorized_use_at: nil).count,
```
tdooner, line 148: `invitation.cbv_flows.completed.count`

app/spec/controllers/cbv/entries_controller_spec.rb
```diff
@@ -45,6 +45,36 @@
         )
       end
 
+      context "with multiple cbv flows" do
+        it "does the thing" do
```
tdooner, line 49: ^ reminder to write something here

app/app/controllers/cbv/base_controller.rb
```diff
@@ -143,9 +143,18 @@ def track_invitation_clicked_event(invitation, cbv_flow)
...
-  rescue => ex
-    Rails.logger.error "Unable to track event (ApplicantClickedCBVInvitationLink): #{ex}"
+    rescue => ex
+      Rails.logger.error "Unable to track event (ApplicantClickedCBVInvitationLink): #{ex}"
+  end
+
+  def count_unique_members(invitation)
+    return 1 if invitation.cbv_applicant.income_changes.blank?
+
+    invitation.cbv_applicant.income_changes.map { |income_change| income_change.with_indifferent_access[:member_name] }.uniq.count
```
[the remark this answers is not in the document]
tdooner: Yeah, it's unfortunate that this base controller has gotten so loaded with this stuff. Potentially we could extract some of the logic in `set_cbv_flow` into a service object or something, but some logic does have to still remain in the controller (the ultimate `redirect_to` and/or setting the `@cbv_flow`) so it's a bit tricky. Hmmm. Although Ian's work here perpetuates an unfortunate pattern, I don't see it as a blocker for now.

tdooner (approved):

PR #736, opened 2025-06-06
FFS-2953: AZ string token translations

millerti:
## Changes
Strings translated for 14.1 and AZ go-live. This PR includes translations for all known English strings; missing Spanish tokens should now be treated as bugs. This incorporates Patricia's feedback on all strings, but Yuleidy has not yet weighed in on the strings we just got back from Lionbridge. There will be a subsequent PR with touchups.

## Testing Instructions
* Go through the flow in Spanish. You should not see any strings in English, except for the 3 missing strings on the success page.
* Open the help modal in both locales and click on every help topic. The site shouldn't crash!
* Someone who worked on the DES updates: please make sure your copy changes are now reflected.

## Acceptance testing

### DIFF
--- a/app/config/locales/es.yml
+++ b/app/config/locales/es.yml
@@ -405,7 +405,6 @@ es:
     - viernes
     - sábado
     formats:
-      day_of_month: "%-d/%-m"
       default: "%-d/%-m/%Y"
       long: "%-d de %B de %Y"
       month_year: "%B de %Y"


app/app/views/cbv/missing_results/show.html.erb
```diff
@@ -4,7 +4,7 @@
 <div class="usa-prose">
   <p class="line-height-sans-5"><%= t(".not_listed_p1", agency_acronym: current_agency.agency_short_name) %></p>
   <% if @has_payroll_account %>
-    <p class="line-height-sans-5"><%= agency_translation(".not_listed_p2_link_html", agency_url: agency_url, agency_short_name: current_agency.agency_short_name) %></p>
+    <p class="line-height-sans-5"><%= t(".not_listed_p2_link_html", agency_url: agency_url, agency_short_name: current_agency.agency_short_name) %></p>
```
millerti, line 7: No longer need agency_translation for this string because it's generic now!

app/config/locales/en.yml
```diff
@@ -568,28 +566,23 @@ en:
           - Search your email for keywords like "payroll," "[Payroll Provider Name]," or "company ID."
           title: Check your welcome materials
         step2:
-          content:
-          - Your company ID might be on old pay stubs, tax forms, or other payroll papers. Check there first if you have them.
+          content: Your company ID might be on old pay stubs, tax forms, or other payroll papers. Check there first if you have them.
```
millerti, line 569: This is an example of the way the i18n-normalizer would screw up the help modal. I've went ahead and put all single-line content nodes back to single-lines pre-emptively.

app/app/views/help/_help_topic_content.html.erb
```diff
@@ -10,7 +10,7 @@
     <li>
       <strong><%= t("help.show.#{topic}.#{step}.title") %></strong>
       <ul class="main-bullets">
-        <% t("help.show.#{topic}.#{step}.content", default: []).each do |content| %>
+        <% Array(t("help.show.#{topic}.#{step}.content", default: [])).each do |content| %>
```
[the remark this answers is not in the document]
tdooner: Sigh. We should move away from string arrays in the translations, IMHO, since they're pretty hard to work with.

app/config/locales/es.yml
```diff
@@ -302,6 +402,7 @@ es:
     - viernes
     - sábado
     formats:
+      day_of_month: "%-d/%-m"
```
tdooner, line 405: Hmm, I saw this on Jeff's branch but I thought he removed it after I commented that I wasn't sure it made sense. Can you double check where it came from?

For example, rendering the report on June 5th, it would render the income summary table as:

> Mes parcial: desde el 1/6 y hasta el 5/6

which is, I guess, possible to understand, but I think pretty ambiguous. I know that in other countries, they're used to doing day/month/year, but I'm not sure whether Spanish-speaking Americans would be used to day/month/year or month/day/year.

tdooner (approved):

app/config/locales/es.yml
tdooner: Stand by, I'm asking April, who's asking Patricia.

jeffcatania-usds: i had removed them from the translation PR, but it looks like they got re-added by the monthly summary PR.

tdooner: Were you going to remove this line? If you don't, I can update my other PR to remove it.

tdooner (approved):

millerti:
My bad. I thought that your PR deleted it, but I now see that it is a different line.

PR #747, opened 2025-06-10
FFS-2918: fetch gig mileage expenses

jeffcatania-usds:
## Ticket

Resolves [FFS-2918](https://jiraent.cms.gov/browse/FFS-2918).

## Changes

* Implement the calculation underlying the "Verified Mileage Expenses" column in the monthly summary
* Extends the Gig response object with new 'mileage' property.  Implements this property for Argyle.  Pinwheel does not support mileage, so it will always be nil.
* Calculate 'total_mileage' in the monthly summary calculation of the aggregator_report
* Adds a 'gigs_mileage_present_count' field to the argyle finish sync analytics call. 

## Acceptance testing


app/app/services/aggregators/aggregator_reports/aggregator_report.rb
```diff
@@ -106,12 +106,20 @@ def summarize_by_month(from_date: nil, to_date: nil)
...
           end
         end
     end
 
+    def total_miles(gigs)
+      mileages = gigs.map { |g| g.mileage }.compact
+      # if there are no gigs with miles, return nil
+      return nil if mileages.empty?
+      mileages.sum.round(2)
+    end
```
tdooner, line 121: You could golf this down as `gigs.sum { |g| g.mileage || 0 }`, similar to how we do it with `accrued_gross_earnings`, but no big deal either way.

And, I think we should leave the rounding to the `ReportViewHelper`, since it's mostly a presentation concern rather than something we actually want to do with the data.

app/app/services/aggregators/format_methods/argyle.rb
```diff
@@ -12,6 +12,15 @@ def self.format_employment_status(employment_status)
     end
   end
 
+  def self.format_mileage(distance_string, distance_unit = "miles")
+    return nil if distance_string.blank?
+    distance = distance_string.to_f
+    if distance_unit == "km"
+      distance = distance * 0.62
```
tdooner, line 19: Let's extract to a constant. It feels a bit weird to use only two significant figures here when we could easily use more. Can we use `0.62137` instead?

app/app/services/aggregators/format_methods/argyle.rb
```diff
@@ -12,6 +12,15 @@ def self.format_employment_status(employment_status)
...
+    end
+    distance.round(2)
```
tdooner, line 21: Same comment about rounding.

app/spec/services/aggregators/format_methods/argyle_spec.rb
```diff
@@ -98,6 +98,29 @@
...
+      expect(subject).to be_nil
+    end
+
+    it "'3.4' miles is 3.40 miles" do
+      subject = described_class.format_mileage("3.4", "miles")
+      expect(subject.class).to eq(Float)
+      expect(subject).to eq(3.4)
+    end
+
+    it "'3.4' km is 5.47 miles" do
```
tdooner, line 117:
```suggestion
    it "'3.4' km is 2.11 miles" do
```

tdooner (approved):

app/app/services/aggregators/aggregator_reports/aggregator_report.rb
jeffcatania-usds: i wanted total_miles to be nil'able to know whether it should be displayed or not. But I think i'm over complicating this and should be fine with just letting 0 == nil in this case.

PR #748, opened 2025-06-10
Fix build by allowing workflows to inherit secrets

tdooner:
## Ticket

N/A - Fixes the build, hopefully.

## Context for reviewers

Commit caa7a6a7 added our first secret to the GH Actions build.
According to [the docs][1]:

> Secrets are not automatically passed to reusable workflows.

We need to explicitly allow the reusable workflows to inherit the
secrets from the main actions file. See [these docs][2] for where I got
this implementation from.

[1]: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions
[2]: https://docs.github.com/en/actions/sharing-automations/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow

## Acceptance testing

### DIFF
--- a/.github/workflows/ci-app-vulnerability-scans.yml
+++ b/.github/workflows/ci-app-vulnerability-scans.yml
@@ -24,3 +24,4 @@ jobs:
     uses: ./.github/workflows/vulnerability-scans.yml
     with:
       app_name: "app"
+    secrets: inherit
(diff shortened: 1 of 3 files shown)


tdooner:
Going to self-merge this to hopefully fix the build. Happy to address any followups.

PR #749, opened 2025-06-10
FFS-2979: Deprecating multi-domain configuration

GeorgeCodes19:
## Ticket

Resolves [FFS-2979](https://jiraent.cms.gov/browse/FFS-2979).

## Changes

- [ ] Removed agency specific domain configurations in favor of `{client_agency_id}_DOMAIN_NAME`

## Acceptance testing


app/config/client-agency-config.yml
```diff
@@ -32,9 +31,8 @@
   agency_name: Department of Transitional Assistance
   agency_short_name: DTA
   agency_contact_website: https://www.mass.gov/guides/how-to-contact-dta
+  agency_domain_name: <%= ENV["MA_DTA_DOMAIN_NAME"] %>
```
GeorgeCodes19, line 34: This pilot is deprecated, but I've added this for continuity.

tdooner: I know they're deprecated but we should not risk taking a shortcut here, IMHO, since I am not sure what to expect if these end up being null and we try generating links with a null `host`.

Let's create these env vars in Terraform as well, same as you did for the supported pilots.

tdooner (changes requested):

PR #751, opened 2025-06-10
FFS-2953: implemented final Spanish translations

millerti:
## Changes
Implemented final internal pass on Spanish translations for AZ go-live

## Acceptance testing

### DIFF
--- a/app/app/views/cbv/successes/show.html.erb
+++ b/app/app/views/cbv/successes/show.html.erb
@@ -27,7 +27,7 @@
       <strong><%= t(".whats_next_1_title") %></strong>
       <ul style="list-style-type: disc;">
         <li><%= t(".whats_next_1_li_1", agency_acronym: current_agency.agency_short_name) %></li>
-        <li><%= t(".whats_next_1_li_2", agency_acronym: current_agency.agency_short_name, agency_website: current_agency.agency_contact_website).html_safe %></li>
+        <li><%= t(".whats_next_1_li_2_html", agency_acronym: current_agency.agency_short_name, agency_website: current_agency.agency_contact_website).html_safe %></li>
       </ul>
     </li>
   </ul>
@@ -36,7 +36,7 @@
     <li>
       <strong><%= t(".whats_next_2_title") %></strong>
       <ul style="list-style-type: disc;">
-        <li><%= t(".whats_next_2_li_1", agency_feedback_form: feedback_form_url).html_safe %></li>
+        <li><%= t(".whats_next_2_li_1_html", agency_feedback_form: feedback_form_url).html_safe %></li>
       </ul>
     </li>
   </ul>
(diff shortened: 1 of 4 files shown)


app/app/views/cbv/successes/show.html.erb
```diff
@@ -27,7 +27,7 @@
       <strong><%= t(".whats_next_1_title") %></strong>
       <ul style="list-style-type: disc;">
         <li><%= t(".whats_next_1_li_1", agency_acronym: current_agency.agency_short_name) %></li>
-        <li><%= t(".whats_next_1_li_2", agency_acronym: current_agency.agency_short_name, agency_website: current_agency.agency_contact_website).html_safe %></li>
+        <li><%= t(".whats_next_1_li_2_html", agency_acronym: current_agency.agency_short_name, agency_website: current_agency.agency_contact_website).html_safe %></li>
```
tdooner, line 30: The `.html_safe` at the end of the string is the same thing as what the `_html` suffix does. The `_html` suffix is the preferred way to do it.

```suggestion
        <li><%= t(".whats_next_1_li_2_html", agency_acronym: current_agency.agency_short_name, agency_website: current_agency.agency_contact_website) %></li>
```

tdooner (approved):

app/app/views/cbv/successes/show.html.erb
```diff
@@ -27,7 +27,7 @@
       <strong><%= t(".whats_next_1_title") %></strong>
       <ul style="list-style-type: disc;">
         <li><%= t(".whats_next_1_li_1", agency_acronym: current_agency.agency_short_name) %></li>
-        <li><%= t(".whats_next_1_li_2", agency_acronym: current_agency.agency_short_name, agency_website: current_agency.agency_contact_website).html_safe %></li>
+        <li><%= t(".whats_next_1_li_2_html", agency_acronym: current_agency.agency_short_name, agency_website: current_agency.agency_contact_website) %></li>
```
tdooner, line 30: Oops, accidentally deleted a previous comment that said something like:

The `_html` suffix does effectively the same thing as the `.html_safe` method call. The only difference is that the `_html` suffix allows the I18n library to manage the HTML-safety of the string, which means it will look at the variables being interpolated to the string, and escape them if they are not HTML safe themselves.

E.g.: with the string:

```
string:
  non_safe: %{num} animals
  safe_html: %{num} animals
```

This interpolation is unsafe:

```erb
<%= t("string.non_safe", num: "<script>alert('hi');</script>").html_safe %>
```

The same interpolation is safe when done via the I18n library, because I18n will sanitize the value of `num`:
```erb
<%= t("string.safe_html", num: "<script>alert('hi');</script>") %>
```

See: https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

millerti: Thanks for explaining!

PR #752, opened 2025-06-10
FFS-2917: mileage in monthly summaries

jeffcatania-usds:
## Ticket

Resolves [FFS-2917](https://jiraent.cms.gov/browse/FFS-2917).

## Changes

* Introduces the Verified Mileage Expenses within the Monthly Summary Table Component
* Uses the MonthlySummaryTableComponent in the Summaries view (deprecating the MonthlySummaryCompactComponent)
* On MonthlySummaryTableComponent
      * Introduces is_responsive property to use uswds's "stacked" table display option for mobile.  
      * Introduces show_footnote and show_payments attributes that controls whether to display the expended sections of the summary. These are hidden on the /summaries view.
      * The new "Verified Mileage Expenses" column will only be shown if there are any miles detected.
* On ReportViewHelper, introduce `#get_federal_cents_per_mile` helper method to create a lookup for 2024 and 2025 federal mileage expense rates.
   

image

https://github.com/user-attachments/assets/5392a56d-2271-4f5d-bb62-e90e027f0888

## Acceptance testing


app/app/components/report/monthly_summary_table_component.erb
```diff
@@ -1,59 +1,73 @@
...
+      <% if has_mileage_data? %>
+        <%= row.with_data_cell(is_header: true)
+               .with_content(I18n.t("components.report.monthly_summary_table.verified_mileage_expenses")) %>
+      <% end %>
       <%= row.with_data_cell(is_header: true)
              .with_content(I18n.t("components.report.monthly_summary_table.total_gig_hours")) %>
     <% end %>
 
-    <% unless @has_monthly_summary_results %>
+    <% if not has_monthly_summary_results? %>
```
tdooner, line 19: Best practice is to put the positive case first so we can remove the `not`

app/app/components/table_component.rb
```diff
@@ -15,6 +15,13 @@ class TableComponent < ViewComponent::Base
     }
   }
 
+  def initialize(is_responsive: false, class_names: "", thead_class_names: "border-top-05")
+    @class_names = "usa-table usa-table--borderless width-full"
+    @class_names = [ @class_names, class_names ].join(" ") if class_names.present?
+    @class_names = [ @class_names, " usa-table--stacked" ].join(" ") if is_responsive
```
tdooner, line 21: Interesting - is there a visual difference caused by the responsive styling? I'm kind of surprised that we need to be able to toggle this on/off since I would expect the styling to be the same on desktop (and only change for mobile breakpoints?)

app/app/helpers/report_view_helper.rb
```diff
@@ -4,9 +4,13 @@ def format_hours(hour)
     hour.to_f.round(1)
   end
 
-  def format_miles(miles)
-    return miles unless Float(miles, exception: false).present?
-    miles.to_f.round(1)
+  def get_federal_cents_per_mile(year)
```
tdooner, line 7: nit: `get_` prefix for method names is usually wasted bytes, since `federal_cents_per_mile` is just as good a name.

tdooner (approved):

app/app/components/table_component.rb
jeffcatania-usds: this was implemented specifically to force PDF to be non-responsive.

app/app/components/report/monthly_summary_table_component.erb
jeffcatania-usds: fixed.

PR #761, opened 2025-06-12
FFS-2995: Add import script for translation CSV files

millerti:
## Changes
Creates a more robust tool to import csv files that contain translation updates into their corresponding locale yaml files.

## Testing instructions
Check out the Context for Reviewers section for an idea of how using this tool works in general.
Create a few test csv files and put them into the import folder. Data like this should help you evaluate whether this is working:

```
key,en,es
new.key.here,"whatever","test string to appear in yml file"
existing.key.here,"whatever","test string to override an existing string in yml file"
...etc
```

## Context for reviewers
In step 6 of our recommended translation process ([bottom of this document](https://confluenceent.cms.gov/display/SFIV/Translation+Process+while+Live+in+AZ)), each engineer working on a ticket with string changes will need to import the approved and reviewed strings into our software. This tool will help with this task. Here's what the workflow will look like:
1. After development is complete, the engineer receives a spreadsheet from whoever reviewed the content. This contains all the new locale-specific strings that need to be merged into locale.yml.
2. Verify that it's in the proper format (first row is key,en,* where * is the two letter abbreviation for the locale being updated, probably es)
3. Verify that the content is formatted appropriately (key column has key paths in it that correspond to nodes in the locale.yml file, en is the English string (though this doesn't really matter), and the locale column has a value anywhere that string needs to be edited; blank values here will be skipped)
4. Put the csv file(s) into /app/lib/import/ (this tool supports any number of files, and will infer the destination locale.yml file from the first row in the file; note that we have NOT tested support for updates for anything but es.yml, but this should help a lot if we ever have more locales than es and en)
5. On the command line, run `i18n-tasks csv-import`
6. On the command line, run `i18n-tasks normalize`
7. At this point, the locale.yml file should be ready to commit and test.

## Acceptance testing


millerti:
Note that a few scans are failing because of the current google outage. If they turn out to be actually failing, I'll take a look.

millerti:
Also, apologies, I feel like this script is kind of messy, but I wanted to get it working so that I can move on to the other script that will help devs next week (something that outputs the review csv for a finished ticket).

app/lib/i18n_csv_tasks.rb
```diff
@@ -1,9 +1,15 @@
 # see: https://github.com/glebm/i18n-tasks/wiki/Custom-CSV-import-and-export-tasks
 require "i18n/tasks/commands"
 require "csv"
+require "yaml"
 
 module I18nCsvTasks
   include ::I18n::Tasks::Command::Collection
+
+  IMPORT_FOLDER = File.join(File.dirname(__FILE__), "import")
```
tdooner, line 9: So this script will expect the files in `app/lib/import`? Might be slightly preferable to put them in `tmp` since the intent will be that they don't get committed.

tdooner (approved):

app/lib/i18n_csv_tasks.rb
millerti: Great suggestion. FIxed.

PR #765, opened 2025-06-12
Fix missing "host" ArgumentError on /submits

tdooner:
## Ticket

N/A - Dev bug.

## Context for reviewers

In dev, we don't always have a value defined for
`client_agency.agency_domain`. To allow the `CbvFlow#to_generic_url`
method to succeed, let's remove the `host` option when the value is
`nil`. This will fall back to
`Rails.application.routes.default_url_options[:host]`, which is always
defined.

## Acceptance testing

### DIFF
--- a/app/app/models/cbv_flow_invitation.rb
+++ b/app/app/models/cbv_flow_invitation.rb
@@ -61,7 +61,12 @@ class CbvFlowInvitation < ApplicationRecord
     client_agency = Rails.application.config.client_agencies[client_agency_id]
     raise ArgumentError.new("Client Agency #{client_agency_id} not found") unless client_agency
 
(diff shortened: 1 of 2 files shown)


app/app/models/cbv_flow_invitation.rb
```diff
@@ -61,7 +61,12 @@ def to_url(**options)
     client_agency = Rails.application.config.client_agencies[client_agency_id]
     raise ArgumentError.new("Client Agency #{client_agency_id} not found") unless client_agency
 
-    Rails.application.routes.url_helpers.cbv_flow_entry_url({ token: auth_token, locale: language, host: client_agency.agency_domain, **options }.compact)
+    Rails.application.routes.url_helpers.cbv_flow_entry_url({
+      token: auth_token,
+      locale: language,
+      host: client_agency.agency_domain,
+      **options
+    }.compact)
```
tdooner, line 69: no change, just formatting

tdooner:
Duplicate of #766

PR #769, opened 2025-06-13
FFS-2981: Inconsistently reporting data as blank or N/A

daphnegold:
## Ticket

Resolves [FFS-2981](https://jiraent.cms.gov/browse/FFS-2981).

## Changes
Rows that can be nullable will show N/A

## Acceptance testing

Screenshot 2025-06-13 at 3 25 10 PM


GeorgeCodes19 (approved):
LGTM

GeorgeCodes19 (approved):
My only concern (and I'm not sure how valid it is) is that this doesn't seem to work when a value is an empty string i.e. `""` or just white-space `"   "`. If the likelihood of that occurring is none then this looks great. I did white a couple tests to verify.

PR #772, opened 2025-06-16
FFS-2932: Adding "other_jobs" page

GeorgeCodes19:
## Ticket

Resolves [FFS-2932](https://jiraent.cms.gov/browse/FFS-2932).

## Changes

- [x] Add new `/other_job` route, applicable controller, and view
- [x] Update the `add_jobs_controller` to handle the new navigation flow
- [x] Update site copy
- [x] Update the `/summary` page with a new table for **Other jobs**
- [x] Update staff-facing report with other jobs info
- [x] Update client-facing report with other jobs info
- [x] Implement tracking `ApplicantContinuedFromAddJobsPage`
- [x] Implement tracking `ApplicantContinuedFromOtherJobsPage`
- [x] Write tests for `/other_jobs` controller
- [x] Update tests for `/add_jobs` controller 
- [x] Update Confluence events docs with the new event `ApplicantContinuedFromAddJobsPage`
- [x] Update Confluence events docs with the new event `ApplicantContinuedFromOtherJobsPage`
- [x] Design review 
- [x] Gather and send missing translation entries for translation
- [x] Import Spanish translations

## Generic flow
Screen Shot 2025-06-17 at 16 33 45

## Tokenized flow 
Screen Shot 2025-06-17 at 16 35 08-fullpage

## Client-facing Generic Flow Report
Screenshot 2025-06-18 at 13-50-59 Verify gov - 310 pdf

## Staff facing tokenized
Screenshot 2025-06-18 at 14-48-14 Verify gov - 311 pdf

## Summary table with Other jobs info
Screenshot 2025-06-18 at 12-41-15 Review your income report SNAP Income Pilot

## Metric logging
```bash
16:35:13 web.1    | [ActiveJob] Enqueued EventTrackingJob (Job ID: 42dbaed2-6517-4924-9c0b-7ff89b882b3a) to
 SolidQueue(default) with arguments: "ApplicantContinuedFromOtherJobsPage" 👈🏾, {headers: {"User-Agent": 
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:138.0) Gecko/20100101 Firefox/138.0"}, remote_ip: "127.0.0.1"},
 {timestamp: 1750192513, ip: "127.0.0.1", cbv_flow_id: 305, client_agency_id: "az_des", 
locale: "en", 
user_agent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:138.0) Gecko/20100101 Firefox/138.0", 
has_other_jobs: false} 👈🏾
```

## Acceptance testing

### DIFF
--- a/app/config/locales/es.yml
+++ b/app/config/locales/es.yml
@@ -58,29 +58,17 @@ es:
   applicant_mailer:
     invitation_email:
       body_1:
-        default: "%{agency_acronym} quiere que usted verifique sus ingresos como parte
-          de su solicitud o recertificación del Programa de Asistencia Nutricional
-          Suplementaria (SNAP, por sus siglas en inglés). El SNAP Income Pilot (Piloto
-          de ingresos) es una nueva herramienta diseñada para ayudarle a compartir
-          información de sus ingresos directamente con %{agency_acronym} iniciando
-          sesión en su proveedor de nóminas."
+        default: "%{agency_acronym} quiere que usted verifique sus ingresos como parte de su solicitud o recertificación del Programa de Asistencia Nutricional Suplementaria (SNAP, por sus siglas en inglés). El SNAP Income Pilot (Piloto de ingresos) es una nueva herramienta diseñada para ayudarle a compartir información de sus ingresos directamente con %{agency_acronym} iniciando sesión en su proveedor de nóminas."
       body_2_html:
-        default: Este proceso es solo para usted. Por favor, no comparta esta invitación
-          con nadie más. <strong>Tiene hasta el %{deadline}, para realizar su verificación.</strong>
-          De lo contrario, puede solicitar una nueva invitación.
+        default: Este proceso es solo para usted. Por favor, no comparta esta invitación con nadie más. <strong>Tiene hasta el %{deadline}, para realizar su verificación.</strong> De lo contrario, puede solicitar una nueva invitación.
       body_3:
-        default: Esta es una herramienta opcional. Si decide que no desea usarla,
-          puede proveer los documentos de sus ingresos a través de la aplicación %{app_name},
-          fax, correo postal o en persona.
+        default: Esta es una herramienta opcional. Si decide que no desea usarla, puede proveer los documentos de sus ingresos a través de la aplicación %{app_name}, fax, correo postal o en persona.
       button: Verifique sus ingresos
-      button_caption: 'Para verificar sus ingresos con el SNAP Income Pilot, haga
-        clic en el siguiente botón:'
-      footer: Este es un mensaje generado automáticamente desde su agencia de SNAP.
-        Por favor no conteste a este correo electrónico.
+      button_caption: 'Para verificar sus ingresos con el SNAP Income Pilot, haga clic en el siguiente botón:'
+      footer: Este es un mensaje generado automáticamente desde su agencia de SNAP. Por favor no conteste a este correo electrónico.
       greeting: "¡Hola!"
       header:
-        sandbox: La Agencia de pruebas CBV le ha enviado una invitación para verificar
-          sus ingresos
+        sandbox: La Agencia de pruebas CBV le ha enviado una invitación para verificar sus ingresos
       subject:
         default: Verifique sus ingresos para su solicitud o renovación de SNAP
   caseworker:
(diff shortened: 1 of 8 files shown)


app/config/locales/es.yml
```diff
@@ -545,6 +542,8 @@ es:
       extend_button: Sí, extender el tiempo de mi sesión.
       heading: "¿Necesita más tiempo en la sesión?"
   shared:
+    next_path:
```
GeorgeCodes19, line 545: Moved to shared since multiple controller are now relying on this entry.

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -55,45 +55,48 @@
...
-    <% @cbv_flow.cbv_applicant.applicant_attributes.each do |item| %>
-      <% if @cbv_flow.cbv_applicant[item].present? %>
-        <tr>
-          <td><%= t("cbv.applicant_informations.#{@cbv_flow.client_agency_id}.fields.#{item}.prompt") %></td>
-          <td>
-            <% if item == :date_of_birth %>
-              <%= format_date(@cbv_flow.cbv_applicant[item]) %>
-            <% else %>
-              <%= @cbv_flow.cbv_applicant[item] %>
+  <table class="usa-table usa-table--borderless width-full margin-top-2 margin-bottom-0">
```
GeorgeCodes19, line 58: The USWDS utility class `margin-bottom-0` has been added to the table.

app/app/views/cbv/summaries/_other_jobs_table.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="<%= local_assigns[:wrapper_class] %>">
+    <% if local_assigns[:header] %>
```
GeorgeCodes19, line 2: Even though there are attempts to render the [header within the table element elsewhere in the codebase](https://github.com/DSACMS/iv-cbv-payroll/pull/772/files#diff-00417506b741a1c59995d779503563e191a9a7611130508488f0f5bd16d92b29R59), it will ultimately be compiled and hoisted outside of the table. I think placing it where it will ultimately render helps eliminate some confusion attempting to inspect the page's source code. Also, I didn't feel as though this needed to be a `ViewComponent`, but I'm open to feedback.

app/app/views/cbv/add_jobs/show.html.erb
```diff
@@ -1,37 +1,21 @@
 <% content_for :title, t(".header") %>
 <h1><%= t(".header") %></h1>
-<p><%= t(".subheader") %></p>
+<p><%= t(".header_sub_text") %></p>
 
 <div data-controller="cbv-add-job">
```
tdooner, line 5: I noticed that this page's `form_with`, and its duplication of it on /other_jobs, have a `data: { "add-job-target": "addJobForm" }` attribute. Looks like it was added originally in 5871a059. They indicate the existence of a stimulus controller for this page, but I don't think we've ever had one.

I think you can remove this attribute from those pages, along with this instantiation of the controller.

app/app/views/cbv/summaries/show.html.erb
tdooner: Thanks for fixing the indenting and calling out the actual diff so I don't drive myself mad trying to actually find it.

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -55,45 +55,48 @@
 <% end %>
 
 <% if @cbv_flow.cbv_flow_invitation.blank? %>
-<table class="usa-table usa-table--borderless width-full margin-top-2">
-  <h3><%= t(".your_information") %></h3>
```
tdooner, line 59: Wait, wtf, we have an `<h3>` and a `<div>` within a `<table>`???? That can't be valid HTML....

I'll make a note of this in our future ticket to refactor tables that Daphne's writing now.

app/db/migrate/20250617183425_add_has_other_jobs_to_cbv_flow.rb
```diff
@@ -0,0 +1,5 @@
+class AddHasOtherJobsToCbvFlow < ActiveRecord::Migration[7.1]
+  def change
+    add_column :cbv_flows, :has_other_jobs, :boolean, default: false
```
tdooner, line 3: Let's omit the `default` -- there is a semantic difference between `null` (the user hasn't answered the question yet) and `false` (the user has answered the question with a "no" answer).

app/app/views/cbv/summaries/_other_jobs_table.html.erb
```diff
@@ -0,0 +1,23 @@
...
+                    <h4 class="margin-0"><%= t(".other_jobs") %></h4>
+                </th>
+                <th class="padding-3" style="background-color: #d9e8f6;">
+                    <%# intentionally empty for equal header spacing %>
+                </th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td><%= strip_tags(t("cbv.other_jobs.show.header_html")) %></td>
```
tdooner, line 18: Why `strip_tags` here? Usually that is used to sanitize user input. I think it will remove the `<em>`, and the designs look like we should have some italic text in there.

app/app/views/cbv/summaries/_other_jobs_table.html.erb
tdooner: Agreed that it doesn't need to be a ViewComponent, since those are useful for cross-page reusability or if there is a lot of view logic we want to extract out of the ERB. A partial seems like a preferable call.

My ultimate preference, though, if possible, would be to not need the partial at all by just making the rendering conditional in the view itself, e.g. in the show.html.erb:

```erb
<% if @cbv_flow.cbv_flow_invitation.blank? %>
  <table ...>
    <%# ... existing table stuff for tokenized links ... %>
  </table>
<% end %>

<% if @cbv_flow.cbv_flow_invitation.blank? %>
  <%# ... render the <h3> with the margin-top-5 class %>
<% end %>

<%# ... render the rest of the other jobs table ... %>
```

Not sure if this will work with the different spacing requirements, so I think the partial is ok if it works. But it seems like the partial is not really simplifying much since we still have to reason about the conditional logic of which classes get applied and which elements get rendered in both cases.

tdooner (commented):
Looking good so far. Some early feedback on whether to use a partial and the HTML structure of the form, but good start.

app/app/controllers/cbv/other_jobs_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  def show
+  end
+
+  def create
+    @cbv_flow.update!(has_other_jobs: params[:additional_jobs] == "true")
+
+    begin
+      event_logger.track("ApplicantContinuedFromOtherJobsPage", request, {
+        timestamp: Time.now.to_i,
+        referer: request.referer,
```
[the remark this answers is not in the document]
GeorgeCodes19: Correct, I copied this over and haven't gone through things with a fine toothed comb.

app/app/controllers/cbv/other_jobs_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        cbv_flow_id: @cbv_flow&.id,
+        client_agency_id: @cbv_flow&.client_agency_id,
+        has_other_job: @cbv_flow.has_other_jobs
```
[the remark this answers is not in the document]
GeorgeCodes19: I have this on my to do list of things of things to cleanup. Nonetheless, thanks for the suggestion.

app/app/views/cbv/summaries/_other_jobs_table.html.erb
GeorgeCodes19: Yeah this was the path that I originally took and wrestled with the table spacing. The partial works fine here so I'll keep this rendering logic as-is.

app/app/views/cbv/summaries/show.html.erb
GeorgeCodes19: Yup, that didn't seem right. I changed this in the partial that I created so that we're not building on this practice.

app/app/controllers/cbv/other_jobs_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Cbv::OtherJobsController < Cbv::BaseController
+  def show
+    @cbv_flow.has_other_jobs = nil if @cbv_flow.has_other_jobs == false
```
GeorgeCodes19, line 3: Prevents a radio value from being selected by default.

app/app/views/cbv/summaries/_other_jobs_table.html.erb
tdooner: Ah, I see, but I don't think we should programmatically transform I18n strings. If there's no italics in this version of the string, let's just make it a separate string.

app/app/controllers/cbv/add_jobs_controller.rb
```diff
@@ -3,17 +3,32 @@ def show
...
+      })
+    rescue => ex
+      raise unless Rails.env.production?
+      Rails.logger.error "Unable to track ApplicantContinuedFromAddJobsPage event: #{ex}"
+    end
+
     redirect_to next_path
   end
 
   def next_path
```
GeorgeCodes19, line 26: `next_path` should only be handling the path resolution. I've moved the logic that appends a flash method to the `create` method of the controller. This also ensures that we're only logging the event once the user has selected a valid option rather than always logging the event.

app/app/helpers/report_view_helper.rb
```diff
@@ -31,6 +31,19 @@ def format_money(dollars_in_cents)
     number_to_currency(dollars_in_cents.to_f / 100)
   end
 
+  def format_boolean(boolean_value)
```
GeorgeCodes19, line 34: The PR required a means to convert a boolean value from the db to a human readable string so I've added this method.

app/app/services/client_agency/report_fields.rb
GeorgeCodes19, line 1: Definitely open to feedback on this... 

Instead of modifying the `submit.pdf.erb` view to explicitly have this field- I opted to stick with the existing pattern of specifying report fields via `ClientAgency::ReportFields`. Inheritance > an explicit entry and code duplication made sense to me.

It's a requirement of this PR that `cbv_flow.has_other_jobs` visible on both the client-facing and caseworker report.

app/app/services/client_agency/report_fields.rb
```diff
@@ -0,0 +1,23 @@
+class ClientAgency::ReportFields
+  extend ReportViewHelper
+
+  def self.caseworker_fields_for(_cbv_flow)
+    []
+  end
+
+  def self.applicant_fields_for(cbv_flow)
+    [
+      [ additional_jobs_to_report_string(cbv_flow), format_boolean(cbv_flow.has_other_jobs) ]
```
GeorgeCodes19, line 10: This is the field that should appear on all reports.

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -55,44 +55,68 @@
...
-  </tbody>
+    </tbody>
+    <tbody>
+      <tr>
+        <th class="padding-3" style="background-color: #d9e8f6;" colspan="2">
+          <h4 class="margin-0"><%= t("cbv.summaries.show.other_jobs") %></h4>
+        </th>
+      </tr>
+      <tr>
+        <td><%= strip_tags(t("cbv.other_jobs.show.header_html")) %></td>
```
tdooner: Let's make this a separate string as well.

app/app/views/cbv/summaries/show.html.erb
```diff
@@ -55,44 +55,68 @@
...
+        <td><%= @cbv_flow.has_other_jobs ? t("us_form_with.boolean_true") : t("us_form_with.boolean_false") %></td>
```
tdooner: I think we have `ReportHelper` included here, so you could probably re-use the `format_boolean` method?

tdooner (approved):
Looks great.

PR #773, opened 2025-06-17
No ticket: create translations:generate rake task

millerti:
## Changes
This ticket creates a `bin/rails translations:generate` task that will create a csv file ready to be sent to our translators.

## Context for reviewers
This `bin/rails translations:generate` task will be used in [step 3 of our translations process](https://confluenceent.cms.gov/display/SFIV/Translation+%28T9n%29+Process). It looks at all the changes made to en.yml when compared to the branch's creation and creates a csv in the right format to send to translators.

## Testing instructions
0. Copy this new file to a new branch.
1. Make some changes to en.yml and commit them. Try doing this several times so that there are multiple commits with changes.
2. Run `bin/rails translations:generate` and verify that the resulting csv file contains exactly the changes you made.

## Acceptance testing


millerti:
I've just realized that this script will not work very well if we're using the feature flag strategy. If there are multiple branches that have been merged in, we will need to run this in each of them to get a complete picture of the strings that need to be translated. Do you think that's OK? Or should I go back to the drawing board?

app/lib/tasks/generate_translations.rake
```diff
@@ -0,0 +1,166 @@
...
+    puts "4. Translating strings to `#{TARGET_LANGUAGE_CODE}` and generating CSV..."
+    create_csv(keys_to_translate, flat_current_en)
+
+    puts Rainbow("\n🎉 Success! CSV file created at: #{OUTPUT_CSV_PATH}").green
+    puts "Please review the file and send it to your translation service."
+  # rescue Google::Cloud::Error => e
+  #   puts Rainbow("\n💥 Google Cloud Error:").red
+  #   puts Rainbow("   Could not authenticate or communicate with the Translation API.").red
+  #   puts Rainbow("   Please ensure you have run `export GOOGLE_APPLICATION_CREDENTIALS=/path/to/key.json`").red
+  #   puts "   Error details: #{e.message}"
```
millerti, line 70: I plan to leave these comments in since we will presently be investigating whether these automated translations are useful.

app/services/locale_diff_service.rb
```diff
@@ -0,0 +1,89 @@
+require 'yaml'
```
millerti, line 1: This code will eventually be called by our GitHub Actions script to perform a very similar task.

app/lib/tasks/generate_translations.rake
```diff
@@ -0,0 +1,125 @@
...
+
+# The branch to compare against (e.g., main, develop).
+# The script will find what has changed in `en.yml` since this branch.
+BASE_BRANCH = "main"
+EN_LOCALE_PATH = "app/config/locales/en.yml"
+ES_LOCALE_PATH = "app/config/locales/es.yml"
+TARGET_LANGUAGE_CODE = "es"
+OUTPUT_CSV_PATH = "translation_update.csv"
+
+class TranslationDiffGenerator
```
tdooner, line 21: Let's put this in its own file, just so it behaves the same way as all our other classes (with regards to what has been autoloaded, and testing, etc.)

app/lib/tasks/generate_translations.rake
```diff
@@ -0,0 +1,125 @@
...
+    old_en_hash = YAML.safe_load(old_en_yaml_content) || {}
+
+    puts "2. Loading current locale files from your branch..."
+    current_en_hash = @locale_diff_service.load_yaml_file(File.join(@project_root, EN_LOCALE_PATH))
+    # current_es_hash = @locale_diff_service.load_yaml_file(File.join(@project_root, ES_LOCALE_PATH))
+
+    # Flatten the hashes to make them easy to compare (e.g., { "en.users.show.title" => "User Profile" }).
+    flat_old_en = @locale_diff_service.flatten_hash(old_en_hash)
+    flat_current_en = @locale_diff_service.flatten_hash(current_en_hash)
+    # flat_current_es = @locale_diff_service.flatten_hash(current_es_hash)
```
tdooner, line 50: All this commented-out code is pretty messy - does it all pertain to the desired machine-translation future?

tdooner (approved):
Looks good. No real blockers since this is a dev tool, I don't think it's worthy of blocking for any of these nitpicks. I pulled it and it worked for me locally.

app/lib/tasks/generate_translations.rake
millerti: Yeah, that's why I left it in -- see my pre-emptive comment on this PR. But I can just remove it. I don't think it's actually saving me a ton of rework.

PR #785, opened 2025-06-23
FFS-2325: Remove New Relic event tracking, update events

patsier-cms:
## Ticket

Resolves [FFS-2325](https://jiraent.cms.gov/browse/FFS-2325).

## Changes

- Removes `NewRelicEventTracker` class and any uses
- Updates all uses of older New Relic event names (from `NewRelicEventTracker::NEW_RELIC_EVENT_MAP`) to use current Mixpanel names
- Removes mentions of logging "NewRelic event" when it no longer applies in logs

## Context for reviewers

- Leaves in `ApplicationController::add_newrelic_metadata` in case that's being used for general metrics, but that's quick to remove
- Leaves in the custom event for `ApplicationJob` since that seems to be used for metrics and alerts rather than general product tracking

## Acceptance testing


app/app/controllers/cbv/base_controller.rb
```diff
@@ -122,18 +122,18 @@ def track_timeout_event
       client_agency_id: current_agency&.id
     })
   rescue => ex
-    Rails.logger.error "Unable to track NewRelic event (ApplicantTimedOut): #{ex}"
+    Rails.logger.error "Unable to track event (ApplicantTimedOut): #{ex}"
   end
 
   def track_expired_event(invitation)
-    event_logger.track("ApplicantLinkExpired", request, {
+    event_logger.track("ApplicantAccessedExpiredLinkPage", request, {
```
jeffcatania-usds, line 129: Please make updates to the CBV Event tracker. https://confluenceent.cms.gov/display/SFIV/CBV+Events

jeffcatania-usds (approved):
Looks good!  Just update the confluence page and you should be good to go.

app/app/controllers/cbv/base_controller.rb
millerti: Yep!

PR #786, opened 2025-06-23
FFS-2687: Add E2E test framework that replays modal callbacks/webhooks

tdooner:
## Ticket

Resolves [FFS-2687](https://jiraent.cms.gov/browse/FFS-2687).

## Changes

**Update ArgyleWebhooksManager to return all created subscriptions**

**Add `rspec` binstub**
> This is an alternative to using `bundle exec rspec`. It was generated
> with `bundle binstubs rspec-core`.

**Create E2eCallbackRecorder to record/replay SDK callback events**

**Upgrade puma 6.4.3 -> 6.6.0**
> This is necessary or else we get an error when running E2E tests.

**Add E2E test for caseworker login**

**Move `with_ngrok_tunnel` context to E2e::NgrokManager**
> This will help us dynamically manage it via a E2e record/replay service
> in a future commit.

**Make NgrokManager retry if one is already running**
> This helps running E2e tests in quick succession.

**Move E2eTestHelpers -> E2e::TestHelpers**

**Move SUPPORTED_PROVIDERS onto Rails.application.config**
> This has become globally useful, so let's put it in the application
> config, and make a test helper to change its value.

**Finish initial attempt at getting E2e tests working**
> * Create a E2e::MockingService that handles all mocking, and a bunch of
>   helper classes in `spec/support/e2e`.
> * Add documentation of the whole enchilada to the `README.md`

## Context for reviewers

**I recommend reviewing the commits in order**. All but the last commit are incremental cleanup/changes necessary for the ultimate framework, and then the last commit ties it all together in an example.

This does not actually enable the tests to run: i.e., the `E2E_RUN_TESTS` env var must still be manually provided.

After this PR is merged, I will do a new PR that adds the recordings for these tests and configures them to run in CI.

## Acceptance testing


app/spec/e2e/caseworker_login_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    fill_in "Case Number", with: "Some kinda case"
+    fill_in "Your WELID", with: "12345"
+    fill_in "Client's email address", with: "hank@example.com"
+    click_on "Send Invitation"
+    expect(page).to have_content "In what language should we send the invitation?"
+    find(".usa-radio-group .usa-radio__label[for='language-en-0']").click
+    click_on "Send Invitation"
+    expect(page).to have_content "Successfully delivered invitation to hank@example.com"
+  end
+end
```
tdooner, line 41: This came from #549 , and is an example of a normal Selenium test that doesn't need to use the recording/replaying framework since it doesn't interact with any aggregators.

app/spec/e2e/cbv_flow_spec.rb
```diff
@@ -1,112 +1,71 @@
...
+    # /cbv/summary
+    verify_page(page, title: I18n.t("cbv.summaries.show.header"))
+    click_on "Continue"
+    find(:css, "label[for=cbv_flow_consent_to_authorized_use]").click
+    click_on "Share my report with CBV"
 
-    it_behaves_like "proceeding through the flow normally"
+    # /cbv/success
+    verify_page(page, title: I18n.t("cbv.successes.show.header", agency_acronym: ""))
+    # TODO: Test PDF rendering by writing it to a file
```
tdooner, line 69: Hard to read this diff, but it's pretty similar to the pinwheel file above.

app/spec/e2e/cbv_flow_spec.rb
```diff
@@ -1,112 +1,71 @@
 require "rails_helper"
 
 RSpec.describe "e2e CBV flow test", type: :feature, js: true do
-  include E2eTestHelpers
```
tdooner, line 4: It's a bit hard to read this diff, but it is pretty much identical to the pinwheel spec above (except for within the Argyle modal, of course).

app/spec/e2e/cbv_flow_spec.rb
```diff
@@ -1,112 +1,71 @@
...
+    raise "Argyle not in supported_providers!" unless Rails.application.config.supported_providers.include?(:argyle)
 
-    # TODO: Remove these when we stub out Pinwheel usage:
-    page.quit
-    WebMock.disable_net_connect!
-  end
+    # /cbv/entry
+    visit URI(root_url).request_uri
+    visit URI(cbv_flow_invitation.to_url).request_uri
+    verify_page(page, title: I18n.t("cbv.entries.show.header"))
```
jeffcatania-usds, line 21: A broader observation that I've made in our testing is the usage of the I18n.t method to test string equality. If test using the translation method instead of the actual expected contents, we will not catch errors within the strings itself.  The advantage of using I18n.t is that it makes it easier to copy-paste tests between languages and reduces the need to update tests every time the text changes.  But now that we are doing actual end-to-end testing, it feels like we should be testing the true "ends" of our system which includes strings post-translation.

app/spec/support/e2e/test_helpers.rb
```diff
@@ -0,0 +1,31 @@
+module E2e
+  module TestHelpers
+    def verify_page(page, title:, wait: Capybara.default_max_wait_time)
```
jeffcatania-usds, line 3: maybe a comment that makes it obvious that the wait is in seconds?

tdooner: Good idea - Just added documentation to this method.

app/spec/spec_helper.rb
```diff
@@ -58,7 +58,7 @@
   config.filter_run_when_matching :focus
 
   # Don't run E2E tests with JS for now
-  if ENV["RUN_E2E_TESTS"].nil?
+  if ENV["E2E_RUN_TESTS"].nil?
```
iannorriswork, line 61: can we add them to CI?

app/app/javascript/utilities/E2eCallbackRecorder.ts
```diff
@@ -0,0 +1,109 @@
+const SESSION_STORAGE_CALLBACKS_KEY = "e2eInvokedCallbacks" // Keep in sync with E2e::MockingService
+const SESSION_STORAGE_REPLAY_KEY = "e2eCallbacksToInvoke" // Keep in sync with E2e::MockingService
+
```
GeorgeCodes19, line 3:
```suggestion

type Indexable = Record<string, unknown>
```

app/app/javascript/utilities/E2eCallbackRecorder.ts
GeorgeCodes19, line 1: I'm having difficulties running the tests, but I have a solution for some awkwardness around conventions in this file such as a [leading semicolon](https://github.com/DSACMS/iv-cbv-payroll/pull/786/files?short_path=b335630#diff-b3e1a1d6cef562f7b6f80601f59c52eda5f4fe33964114e7a67bb9221d0f0dafR73) and replacing `any` with `unknown` (which will force explicit type checking). I've stashed my changes locally, but I don't want to hold this up so I can follow up with a small PR to resolve some of the TS issues later.

GeorgeCodes19 (approved):

app/app/javascript/utilities/E2eCallbackRecorder.ts
```diff
@@ -0,0 +1,103 @@
...
+  /*
+   * Intercept callbacks by replacing them with a "recorder" that saves the
+   * invocation's arguments into sessionStorage and then calls the original
+   * callback method.
+   */
+  interceptCallbacks(params: ArgyleInitializationParams | InitializationParams) {
+    // Assume that all functions in the params object's values are callback
+    // methods we should be proxying. This assumes that `params` is flat (i.e.
+    // contains no callbacks nested under another key).
+    for (const [key, value] of Object.entries(params)) {
```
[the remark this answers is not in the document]
jeffcatania-usds: If you do turn this into an each, then you could do the type definition within the callback function to avoid the awkwardness of the (params as any) on line 65 and 67.   The other thing to consider for type definition is the use of generics instead of ArgyleInitializationParams | InitializationParams.

app/spec/spec_helper.rb
tdooner: #787 does this!

app/spec/e2e/caseworker_login_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe 'Caseworker Login and Invitation', type: :system do
+  before do
+    driven_by(:selenium_chrome_headless) # Change to :selenium_chrome if you want to see the browser
```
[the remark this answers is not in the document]
tdooner: I updated #787 to make headless the default, since that's how it runs in CI, and how we probably want to run it locally unless we're explicitly trying to debug it during local dev (in which case we can pass `E2E_SHOW_BROWSER`)

app/app/javascript/utilities/E2eCallbackRecorder.ts
tdooner: Fixed in: [`14a1c53` (#786)](https://github.com/DSACMS/iv-cbv-payroll/pull/786/commits/14a1c5360ca3d8f1d5d7a053bc64f669da6a23ae)

patsier-cms (approved):
I don't have strong thoughts on the use of I18n here, so this looks good to me! Glad to see it running on the other branch, and will defer on the most straightforward merge order

app/spec/e2e/cbv_flow_spec.rb
tdooner: Agreed - there are upsides/drawbacks to both: Using i18n prevents the need to change the tests whenever the headings change, but of course we run into interpolation issues and wouldn't be catching any underlying issues with I18n.

jeffcatania-usds (approved):
LGTM!

PR #789, opened 2025-06-24
Expire sessions in Rails rather than via cookie `Expires`

tdooner:
## Ticket

Related to [FFS-2687](https://jiraent.cms.gov/browse/FFS-2687).

## Context for reviewers

This is motivated by the E2E test branch - in those tests we will be
stubbing the Rails Time.now (the basis for the cookies' `Expires`
attribute), but Chrome will happily invalidate the cookie based on the
real time the test is being run.

To eliminate this issue, this commit moves the session expiration into
the cookie's contents instead, by storing an `expires_at` value, where
Rails can exclusively determine whether 30 minutes has passed.

I tested this manually by advancing my system clock, and even once
waiting 30 minutes the long way.

## Acceptance testing

### DIFF
--- a/app/app/controllers/application_controller.rb
+++ b/app/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   before_action :redirect_if_maintenance_mode
   before_action :enable_mini_profiler_in_demo
   before_action :check_help_param
+  before_action :validate_session_expiration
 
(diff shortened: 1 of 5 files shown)


tdooner:
As [discussed in Slack](https://cmsgov.slack.com/archives/C066XS5543S/p1750787038350379) I will reimplement using an environment variable to eliminate the default cookie timeout.

PR #788, opened 2025-06-24
Cache `bundle install` when running Rspec

tdooner:
## Ticket

N/A - Noticed this while doing the E2E test branch.

## Context for reviewers

This should shave off ~1 minute from our test runs whenever we're not changing
the Gemfile.

## Acceptance testing

- [x] No acceptance testing needed
  * This change will not affect the user experience (bugfix, dependency updates, etc.)
- [ ] Acceptance testing prior to merge
  * This change can be verified visually via screenshots attached below or by sending a link to a local development environment to the acceptance tester
  * Acceptance testing should be done by **design** for visual changes, **product** for behavior/logic changes, **or both** for changes that impact both.
- [ ] Acceptance testing after merge
  * This change is hard to test locally, so we'll test it in the demo environment (deployed automatically after merge.)
  * Make sure to notify the team once this PR is merged so we don't inadvertently deploy the unaccepted change to production. (e.g. `:alert: Deploy block! @ffs-eng I just merged PR [#123] and will be doing acceptance testing in demo - please don't deploy until I'm finished!`)

PR #791, opened 2025-06-24
Fix erb_lint inadvertently running on vendor/bundle files

tdooner:
## Ticket

N/A - Test failure introduced by #788, I think.

Erb-lint seems to be failing on a bunch of lint errors in vendor/bundle.

## Changes

I suppose we need to configure erb_lint to ignore that directory.

## Acceptance testing

### DIFF
--- a/app/app/views/cbv/summaries/show.html.erb
+++ b/app/app/views/cbv/summaries/show.html.erb
@@ -27,13 +27,13 @@
             <th class="padding-3" style="background-color: #d9e8f6;" colspan="2">
               <h4 class="margin-0" data-testid="paystub-total-income">
                 <% if summary[:total] > 0 %>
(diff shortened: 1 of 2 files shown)


app/app/views/cbv/summaries/show.html.erb
```diff
@@ -27,13 +27,13 @@
             <th class="padding-3" style="background-color: #d9e8f6;" colspan="2">
               <h4 class="margin-0" data-testid="paystub-total-income">
                 <% if summary[:total] > 0 %>
-                <% if employer_name %>
```
tdooner, line 30: This is a dummy change just to force `erb_lint` to run on this PR.

tdooner:
Going to self-merge so my other PR will not fail when these tests run on it.

PR #796, opened 2025-06-26
FFS-2849: Implement wind-down content for LA

GeorgeCodes19:
## Ticket

Resolves [FFS-2849](https://jiraent.cms.gov/browse/FFS-2849).

## Changes

- [ ] Added `pilot_ended` attribute `client-agency-config.yml` which redirects the `generic_links_controller` to the `root_path`
- [ ] Updated the `application_controller` to set a `pilot_ended` view helper method which is inspected and used by the `root_path` view
- [ ] Added a dynamic view partial pattern for ended pilots
- [ ] Updated `pages_controller_spec.rb` to test if agency pilot has ended
- [ ] Export translations and attach to Jira ticket
- [ ] Import returned Spanish translations

## Acceptance testing


app/config/client-agency-config.yml
GeorgeCodes19, line 1: Even though several of these pilots have ended, we'll set `pilot_ended: false` since we don't want to break tests. It's likely that we'll clean up the codebase and remove these configuration entries anyway.

rutvigupta-design:
LGTM; reviewed and accepted by design! 🎉

app/app/controllers/application_controller.rb
```diff
@@ -82,6 +88,10 @@ def detect_client_agency_from_domain
     end
   end
 
+  def pilot_ended?
+    @pilot_ended.nil? ? current_agency&.pilot_ended : @pilot_ended
+  end
```
tdooner, line 93: This could be simpler - instead of setting an instance variable here to communicate across both methods, you could have made both methods pure by making `pilot_ended?` just return `current_agency&.pilot_ended` and then just call that in the `check_if_pilot_ended` rather than doing this instance variable back-and-forth.

app/app/controllers/cbv/generic_links_controller.rb
```diff
@@ -9,9 +10,18 @@ def show
...
 
     redirect_to root_url, flash: { info: t("cbv.error_invalid_link") }
   end
+
+  def check_if_pilot_ended_for_agency
+    agency = agency_config[params[:client_agency_id]]
+    if agency&.pilot_ended
+      redirect_to root_url
+    end
+  end
```
tdooner, line 26: Do we need this? Shouldn't the ApplicationController redirect get this?

app/config/locales/en.yml
```diff
@@ -604,10 +604,19 @@ en:
       header: It looks like something went wrong
       refresh: Refresh page
     home:
-      description_1: Report My Income is a new tool designed to help you connect your income details from your employer or payroll provider directly to your benefits agency. We’re currently testing this tool to ensure it works effectively.
+      description_1: Report My Income is a new tool designed to help you connect your income details from your employer or payroll provider directly to your benefits agency. We're currently testing this tool to ensure it works effectively.
```
tdooner, line 607: Death to smart quotes. Thanks.

tdooner (approved):
Could be simpler, but since this is time-sensitive, we can simplify that stuff later.

app/app/controllers/cbv/generic_links_controller.rb
GeorgeCodes19: This is needed. Without it the user is redirected to the `cbv/entry` page since the ApplicationController doesn't check the params to resolve the `client_agency_id` in a before hook. It's only done in a [helper method](https://github.com/DSACMS/iv-cbv-payroll/pull/796/files#diff-cba1c47c831f809d988acb1619f556a26a53c3f23ad92f795ed627766a17ca79L3). You can recreate the scenario by commenting out  `before_action :check_if_pilot_ended_for_agency`. I didn't want to introduce side-effects by modifying a base class.

app/app/controllers/application_controller.rb
GeorgeCodes19: That was my original approach, but I [switched things up last minute](https://github.com/DSACMS/iv-cbv-payroll/pull/796/commits/f6f99241dbbdb1291dcb08db37088428fc295a50) to eliminate some redundancy in favor of memoization. I can revert and ensure that `check_if_pilot_ended` calls `pilot_ended?` rather than `current_agency&.pilot_ended`

PR #797, opened 2025-06-26
FFS-2996: Clean up strings to use correct variables

daphnegold:
## Ticket

Resolves [FFS-2996](https://jiraent.cms.gov/browse/FFS-2996).

## Changes
- Replaced `current_agency.agency_short_name` with `agency_translation("shared.agency_acronym")`
- Updated locale files: `%{agency_short_name} --> %{agency_acronym}`
- Updated ERB parameter names to match new locale keys
- Removed unused `agency_short_name` property from Client Agency config

## Context
It wasn't specified in this ticket, but there are other variables we might consider doing this with, like `agency_name`. As a side note `agency_acronym` was exactly the same as `agency_short_name`. I caused myself some fun merge conflicts for sure with #795. 🤦‍♀️ 

## Acceptance testing

### DIFF
--- a/app/spec/controllers/caseworker/entries_controller_spec.rb
+++ b/app/spec/controllers/caseworker/entries_controller_spec.rb
@@ -12,11 +12,10 @@ RSpec.describe Caseworker::EntriesController do
       end
 
       it "should show sandbox specific copy with a link to /sso/sandbox" do
-        agency_short_name = client_agency_config["sandbox"].agency_short_name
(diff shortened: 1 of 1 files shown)


jeffcatania-usds (approved):
this ticket is FUNKY. "Replace usages of current_agency.agency_short_name with agency_translation("shared.agency_full_name") or similar." but it looks like the work to be done is to map agency_short_name to shared.agency_acronym - which is what is done in the PR.

PR #798, opened 2025-06-27
No ticket: fix i18n-tasks health github action + cleanup

millerti:
## Changes
The i18n-health check github action was mysteriously not working anymore. Now, it will.
I've also removed the rspec tests that run i18n-tasks health, since they're now in their own github action.

## Context for reviewers
To trigger this check, change either en.yml or es.yml.

## Acceptance testing


app/config/locales/en.yml
```diff
@@ -8,6 +8,7 @@ en:
             date_of_birth:
               future_date: Date of birth must be today or in the past
               invalid_date: Enter a valid date of birth
+              timotest: TIMOTEST
```
millerti, line 11: This is just here to make sure the github action runs; I will remove this change before merge.

jeffcatania-usds (approved):
LGTM!

PR #799, opened 2025-06-27
FFS-2973: Investigate 'undefined method 'paystubs' for nil' error on summary page

daphnegold:
## Ticket

Resolves [FFS-2973](https://jiraent.cms.gov/browse/FFS-2973).

## Changes
- Added `check_aggregator_report` `before_action` to `SummariesController` to handle nil aggregator report
- Test case for scenario when user has no successfully synced accounts
- Redirect to synchronization failures page instead of crashing with `NoMethodError` (mimics `SubmitsController`)
- Helpful logging

### Root cause

The flow's next_path logic allows users to go directly from `/applicant_information` to `/summary page`, even when they've never connected any employers. This leaves us without successful data to generate a report from.

Users shouldn't be able to reach the `/summary page` when there's no income data to summarize. The summary page assumes there's at least one successfully synced employer account to display paystubs from. 👍 

To replicate:

- Visit http://localhost:3000/cbv/links/la_ldh
- Click "I Agree", then manually navigate to `/cbv/applicant_information`
- Fill out form and click "Continue" (crashes before fix, redirects after)

## Acceptance testing

### DIFF
--- a/app/app/controllers/cbv/summaries_controller.rb
+++ b/app/app/controllers/cbv/summaries_controller.rb
@@ -12,6 +12,7 @@ class Cbv::SummariesController < Cbv::BaseController
 
   def check_aggregator_report
     if @aggregator_report.nil?
+      Rails.logger.error "Aggregator report nil for #{@cbv_flow.id}. User reached summary page without successfully synced payroll accounts."
       redirect_to cbv_flow_synchronization_failures_path
     end
   end
--- a/app/app/controllers/cbv/summaries_controller.rb
+++ b/app/app/controllers/cbv/summaries_controller.rb
@@ -2,9 +2,18 @@ class Cbv::SummariesController < Cbv::BaseController
   include Cbv::AggregatorDataHelper
 
   before_action :set_aggregator_report, only: %i[show]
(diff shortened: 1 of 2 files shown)


app/app/controllers/cbv/summaries_controller.rb
```diff
@@ -2,11 +2,21 @@ class Cbv::SummariesController < Cbv::BaseController
...
 
   def show
     track_accessed_income_summary_event(@cbv_flow, @aggregator_report.paystubs)
   end
 
+  private
+
+  def check_aggregator_report
+    if @aggregator_report.nil?
+      Rails.logger.error "Aggregator report nil for #{@cbv_flow.id}. User reached summary page without successfully synced payroll accounts."
```
iannorriswork, line 15: assuming we'll be committing to checking in on whether this is happening later? i like it though!

iannorriswork (approved):
seems reasonable. as part of this or a followup, could we fix the next_path logic that's sending them a page that won't work and bounce them to the right part of the flow?

danielladevera-nava:
passed acceptance testing 👍

PR #806, opened 2025-07-01
FFS-3007-1: Fix CbvApplicant instantiation and add first/last names to Identity object

tdooner:
## Ticket

Precursor to [FFS-3007](https://jiraent.cms.gov/browse/FFS-3007).

## Changes

Two unrelated things necessary for FFS-3007. (I plan to rebase this on main since these are unrelated).

1. **Fix factories not instantiating CbvApplicant STI subclasses**
Due to the lack of `initialize_with` in the CbvApplicant factory,
FactoryBot always was instantiating a base CbvApplicant rather than the
desired subclass.
There was one spec failure, in the WeeklyReportMailer spec. Somewhere
deep within ActiveRecord, objects for this spec were getting created
with a `client_agency_id` being produced by the `.sti_name` method. This
method was not symmetric with `.sti_class_for`, as it should be, when
the name has a capital letter within it.
This also makes the same change to our other usage of STI, in
PayrollAccount, although that should be a noop since we don't have any
subclasses of PayrollAccount with capital letters within them.
2. **Add `first_name` / `last_name` to ResponseObjects::Identity**
This ticket calls for using these fields, which are only provided by
Argyle. Since Argyle is more common, this implements a hacky backfill
for Pinwheel, assuming single-part first names.

## Context for reviewers

Plan is to merge this one first, without acceptance testing, and then get acceptance testing on the overall feature subsequently.

## Acceptance testing

### DIFF
--- a/app/app/models/cbv_applicant.rb
+++ b/app/app/models/cbv_applicant.rb
@@ -16,8 +16,8 @@ class CbvApplicant < ApplicationRecord
   self.inheritance_column = "client_agency_id"
 
   def self.sti_name
-    # "CbvApplicant::Sandbox" => "sandbox"
-    name.demodulize.downcase
+    # "CbvApplicant::AzDes" => "az_des"
+    name.demodulize.underscore
   end
 
   def self.sti_class_for(type_name)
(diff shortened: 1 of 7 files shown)


GeorgeCodes19 (approved):
No concerns here.

PR #807, opened 2025-07-01
FFS-3007-2: Match names between report and indexing data

tdooner:
## Ticket

Resolves [FFS-3007](https://jiraent.cms.gov/browse/FFS-3007).

## Changes

**FFS-3007: Match names between report and indexing data**
> This new event (IncomeSummaryMatchedAgencyNames) tracks, upon delivery,
> the overlap of names from the agency's original data to the names we get
> from Argyle/Pinwheel.

> Currently, only AZ provides us the expected name(s) of who will be using
> CBV, but this is implemented generically so that we can compute this for
> a future agency as well.

> This creates a Service object to do the comparison, a Job class that
> orchestrates fetching the report data and everything, and some tests to
> make sure it works somewhat well.

## Acceptance testing

### DIFF
--- a/app/app/services/agency_name_matching_service.rb
+++ b/app/app/services/agency_name_matching_service.rb
@@ -1,0 +1,67 @@
+# This class compares two sets of names as outlined in:
+#   https://confluenceent.cms.gov/x/iDy8PQ
+#
+# A match result will be computed for each name in `source_names_list`
+# against all the names in the `target_names_list`.
+class AgencyNameMatchingService
+  # Sorted worst -> best:
+  NAME_MATCH_LEVELS = %i[
+    none
+    approximate
+    close
+    exact
+  ].freeze
+
+  def initialize(source_names_list, target_names_list)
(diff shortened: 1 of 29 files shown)


app/app/services/agency_name_matching_service.rb
```diff
@@ -0,0 +1,67 @@
...
+        results[:none_match_count] += 1
+      else
+        raise "Unknown match type"
+      end
+    end
+  end
+
+  private
+
+  def match_names(name_one, name_two)
```
GeorgeCodes19, line 48: This chunk comparison approach seems a lot more straight forward than using Levenshtein Distance. 

Does the result in the image attached to this comment not occur as often as we'd anticipated? I was surprised to find that the whitespace did not result in an exact match.

Screenshot 2025-07-02 at 1 46 40 PM

GeorgeCodes19 (approved):
This approach is easy to reason with and seems effective enough to produce results that allow us to draw some reasonable conclusions with name matching.

PR #810, opened 2025-07-02
Silence 3 deprecation warnings

tdooner:
## Ticket

N/A - Personal annoyance.

## Changes

- **Set Devise `secret_key` to silence a deprecation warning**
- **Remove RSpec `fixture_path` to silence deprecation warning**
- **Upgrade newrelic_rpm 9.12.0 -> 9.19.0**

## Acceptance testing

### DIFF
--- a/app/Gemfile.lock
+++ b/app/Gemfile.lock
@@ -305,7 +305,7 @@ GEM
     net-smtp (0.5.0)
       net-protocol
     net-ssh (7.3.0)
-    newrelic_rpm (9.12.0)
+    newrelic_rpm (9.19.0)
     nio4r (2.7.4)
     nokogiri (1.18.8-aarch64-linux-gnu)
       racc (~> 1.4)


app/config/initializers/devise.rb
```diff
@@ -21,7 +21,7 @@
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-  # config.secret_key = 'f41eba739bc87c147a84c7dc9b72f0a138b36fd8feb6323a954d75743d89ab61654b691fa73a2c488865cb06bf363510c5abd52ba699bc9e4980f0b6bd9b827d'
+  config.secret_key = Rails.application.secret_key_base
```
tdooner, line 24: The docs say changing this value will result in invalidating sessions, but I don't think that's a risk because Devise is only used for caseworker sessions. And the value is probably not even being changed, since I'd guess it was falling back to this value anyway.

patsier-cms: It looks like if this isn't set it creates a new value in `/app/tmp/local_secret.txt` which persists between local runs, and I'm seeing `SECRET_KEY_BASE` in our deployed conf so this should be alright. If caseworker sessions aren't long-lasting they're likely expired now anyway. Mostly just repeating what I think you're saying here but this should be good!

patsier-cms (approved):
Looks good, and I'm not seeing any warnings with this locally

PR #815, opened 2025-07-03
FFS-2797: Register webhooks automatically on deploy

GeorgeCodes19:
## Ticket

Resolves [FFS-2797](https://jiraent.cms.gov/browse/FFS-2797).

## Changes

- [x] Add a `webhooks:register` Rake task to automate webhook registration for environments (e.g., demo, prod).
- [x] Introduces a new `webhook-registration` job to the deploy.yml workflow, which runs after database migrations.
- [x] The new job executes the Rake task in a one-off ECS container using a dedicated `webhook_registrar` IAM role.
- [x] Adds the necessary Terraform configuration for this role to the `infra/modules/service` module (see context below on why)

## Context for reviewers
Originally, I thought this would be siloed to just a rake task, but seeing that we want it to run during deployment I figured it should mirror the database migrations CI workflow as much as possible. I see this as being an application life-cycle dependency rather than something that operates within the confines of the Rails app. 

The new `webhook_registrar` role is intentionally placed within the service module (rather than its own module) to treat it as a deployment lifecycle task, analogous to the migrator_task used to run ActiveRecord migrations during deployment.

## Acceptance testing


app/lib/tasks/webhooks.rake
```diff
@@ -0,0 +1,27 @@
+namespace :webhooks do
+  desc "Register webhooks"
+  task register: :environment do
+    puts "Registering webhooks..."
+
+    subscription_name = Rails.env
```
GeorgeCodes19, line 6: Is this suffice? I'm not sure. This might always resolve to `production` since demo is also run with `production` as the Rails.env.

millerti:
Closing this until we pick up the work again

PR #816, opened 2025-07-07
[FFS-2854] delete finished jobs from solidqueue on regular basis

iannorriswork:
## Ticket

Resolves [FFS-2854](https://jiraent.cms.gov/browse/FFS-2854).

## Changes
chose to use our technique of jobs from the background rake tasks, though we may want to consider activejob in the future (see https://github.com/rails/solid_queue/blob/e04c1f969e3c9775abf85472b04622f21fa5ef27/lib/generators/solid_queue/install/templates/config/recurring.yml#L12)

## Acceptance testing

- [ X ] Acceptance testing after merge
  * Will confirm that the background job runner is running after, though did run the rake task locally.

### DIFF
--- a/app/config/initializers/solid_queue.rb
+++ b/app/config/initializers/solid_queue.rb
@@ -1,0 +1,1 @@
+SolidQueue.clear_finished_jobs_after = 7.days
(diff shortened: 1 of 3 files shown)


tdooner:
@iannorriswork We did discuss doing this via ActiveJob at the last Eng Sync meeting. (it's in the ticket) Do you have capacity to update this PR or should we merge this and change it later?

daphnegold (approved):
lgtm

PR #817, opened 2025-07-07
[FFS-2496] remove sync status fields from PayrollAccount and begin prepping pinwheel_account_id migration

iannorriswork:
## Ticket

Resolves [FFS-2496](https://jiraent.cms.gov/browse/FFS-2496).

## Changes
Removes columns that are no longer in use as well as begins preparing a column for a more descriptive rename. 

## Acceptance testing

  * Check that the data migrates appropriately and fields are deleted--tested this locally but feel free to test yourself.
  * Also reviewed the newrelic production setup and didn't see any alerts or metrics that were directly tied to this--was there a spot i should've looked other than alerts and dashboards?


app/db/migrate/20250707143214_prepare_pinwheel_account_id_rename.rb
```diff
@@ -0,0 +1,9 @@
+class PreparePinwheelAccountIdRename < ActiveRecord::Migration[7.1]
+  def change
+    add_column :payroll_accounts, :remote_account_id, :string
+
+    PayrollAccount.where.not(pinwheel_account_id: nil).find_in_batches(batch_size: 1000) do |batch|
+     PayrollAccount.where(id: batch).update_all("remote_account_id = pinwheel_account_id")
```
tdooner, line 6: Hmm, naming is hard. How would you feel about `aggregator_account_id` instead of `remote_account_id` (to be slightly more obvious that it's an Account ID issued by the aggregators themselves.)

app/db/migrate/20250707143214_prepare_pinwheel_account_id_rename.rb
```diff
@@ -0,0 +1,9 @@
...
+   end
```
tdooner, line 7: Slight preference for doing data manipulations like this in a rake task instead, in case there's some bug that crashes it here. See: `git log -- lib/tasks/backfills.rake`

tdooner (approved):

PR #829, opened 2025-07-11
Remove update aria-label to be the employer name instead of the descr…

jeffcatania-usds:
…iption text.

## Ticket

## Changes

Defects:

### Employer Search:
- [ ] Tab selection of employers does not announce the employers name, label is also very long
        * https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html
        * [https://stackoverflow.com/questions/71695002/form-control-labels-must-be-unique-for-button](https://stackoverflow.com/questions/71695002/form-control-labels-must-be-unique-for-buttons)s

- [ ]  Move focus into the Argyle / Pinwheel Modal when employer has been selected.
        * https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html
- [ ]  On closing the argyle modal, focus should be set to the new help text or use aria live. 

        * Will need to confirm if this works already in Safari
- [ ] Searching reloads the page, but there's no indication that results have changed
- [ ] Choose from popular providers tabs aren't semantic tab groups

### Synchronizations

- [ ]  [Spike required] Do loading times get announced on synchronizations?

### Payment details

- [ ] Label is too long on additional comments - we should break between label and hint.

        * https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html

## Acceptance testing

### DIFF
--- a/app/app/views/cbv/employer_searches/_employer.html.erb
+++ b/app/app/views/cbv/employer_searches/_employer.html.erb
@@ -33,7 +33,7 @@
               class="usa-button usa-button--outline"
               type="button"
               tabindex="0"
-              aria-label="<%= t("cbv.employer_searches.show.select_button_aria_label") %>"
+              aria-label="<%= employer.name %>"
             >
               <%= t("cbv.employer_searches.show.select") %>
             </button>


jeffcatania:
@millerti @patsier-cms Somebody please close this PR.

PR #831, opened 2025-07-14
No ticket: parameterize accordion heading level

millerti:
## Changes
E2E tests were breaking because the accordion component put h4s into the HTML regardless of the heading structure. Now, you can pass the heading level you want.

## Acceptance testing

### DIFF
--- a/app/app/components/accordion_component.rb
+++ b/app/app/components/accordion_component.rb
@@ -9,6 +9,7 @@ class AccordionComponent < ViewComponent::Base
   def initialize(id:, heading_level: 4, data_action: "", data_section_identifier: "", expanded: false)
     @base_id = id
     @heading_level = heading_level.to_i.clamp(1, 6)
+    @data_action = data_action
     @data_section_identifier = data_section_identifier
     @expanded = expanded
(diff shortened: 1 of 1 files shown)


app/app/components/accordion_component.rb
```diff
@@ -6,9 +6,9 @@ class AccordionComponent < ViewComponent::Base
   renders_one :title
   renders_many :accordion_items
 
-  def initialize(id:, data_action: "", data_section_identifier: "", expanded: false)
+  def initialize(id:, heading_level: 4, data_action: "", data_section_identifier: "", expanded: false)
     @base_id = id
-    @data_action = data_action
```
tdooner, line 11: Looks like removing this variable will break the event tracking on the /entries page since we do pass in a value for this parameter.

tdooner (changes requested):

app/app/components/accordion_component.rb
millerti: Man, I must be tilted today or something.

tdooner (approved):

PR #832, opened 2025-07-15
working accordion component

millerti:
## Ticket

## Acceptance testing


app/app/components/table_component.html.erb
```diff
@@ -1,7 +1,11 @@
 <%= tag.table(**@attributes.merge(class: @class_names)) do %>
   <thead class="<%= @thead_class_names %>">
     <tr>
+      <% if header.present? %>
         <%= header %>
+      <% elsif header_two_column.present? %>
+        <%= header_two_column %>
+      <% end %>
```
millerti, line 8: I think this approach may be controversial. I also think that Jeff may have implemented something that replaces this strategy, but I can't see it yet because he doesn't have github access yet. So FYI this may be replaced in this PR with whatever Jeff's solution is.

app/spec/components/report/payments_deductions_monthly_summary_component_spec.rb
GeorgeCodes19, line 1: There's a lot of code duplication between these test setups I suspect that there's a way to cut down on code duplication here? Maybe using a shared context. At the minimum- a comment denoting what's different in the context setup would be nice.

I think a shared context would look something like this:

```ruby
RSpec.shared_context "argyle gig worker setup" do |include_paystubs: true|
  let(:argyle_report) { 
    Aggregators::AggregatorReports::ArgyleReport.new(
      payroll_accounts: [payroll_account], 
      argyle_service: argyle_service, 
      days_to_fetch_for_w2: 90, 
      days_to_fetch_for_gig: 182
    ) 
  }
  
  before do
    argyle_stub_request_identities_response("bob")
    argyle_stub_request_paystubs_response("bob") if include_paystubs
    argyle_stub_request_gigs_response("bob")
    argyle_stub_request_account_response("bob")
    argyle_report.fetch
  end
end

context "a" do
  include_context "argyle gig worker setup"
  it "does something" do
	...
  end
end

context "b" do
  include_context "argyle gig worker setup", include_paystubs: false
  it "does something" do
	...
  end
end
```

app/app/components/report/payments_deductions_monthly_summary_component.html.erb
```diff
@@ -0,0 +1,34 @@
...
+<% if has_monthly_summary_results? %>
+  <% @monthly_summary_data.each_with_index do |(month_string, summary), index| %>
+    <%= render AccordionComponent.new(id: "#{month_string}-accordion") do |accordion| %>
+      <% accordion.with_title do %>
+        <%= format_month_string(month_string, summary) %>
+      <% end %>
+      <% accordion.with_accordion_item do %>
+        <% summary[:paystubs].each do |paystub| %>
+          <%= render(TableComponent.new(is_responsive: @is_responsive)) do |table| %>
+            <%= table.with_header_two_column(column1_title: t("cbv.payment_details.show.pay_information"), column2_title: t("cbv.payment_details.show.your_details")) %>
```
GeorgeCodes19, line 11: Big fan of the `with_header_two_column`, but the rendered results varies from the Figma mock

Current
Screenshot 2025-07-23 at 4 36 09 PM

Figma
Screenshot 2025-07-23 at 5 16 26 PM

GeorgeCodes19 (changes requested):

GeorgeCodes19:
This mostly looks good! I had a couple nit things that you can take or leave, but I requested changes to ensure that we're aligned with our approaches to styling and to ensure that the end result matches what design expects. Aside from those changes- once design approves then I'll approve.

app/app/assets/stylesheets/uswds-overrides.scss
```diff
@@ -10,6 +10,10 @@
   max-width: none;
 }
 
+.usa-accordion .usa-accordion__button {
```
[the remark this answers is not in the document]
millerti: Maybe I don't understand the approach we're going for, but these are class selector overrides aren't they? Are we avoiding these too?

app/app/components/table_header_two_column_component.html.erb
```diff
@@ -0,0 +1,2 @@
+<th scope="col" class="padding-3" style="background-color: #d9e8f6;"><%= @column1_title %></th>
```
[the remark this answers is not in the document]
millerti: I was surprised to find no other examples of this to fix. If you know of any I missed, lmk!

PR #833, opened 2025-07-15
Add Mixpanel JS and use it to record anonymized user sessions

tdooner:
## Ticket

N/A - Realized we could do this during this morning's meeting about feedback.

## Changes

Mixpanel supports recording user sessions in order to replay a user's
journey through the flow. Since we are investigating user dropoff, let's
enable this feature and see if we can glean anything useful.

## Context for reviewers

No text is included in the PII (either the webpage text nor any
information entered by the user) so these recordings will not include
any PII.

Example:

https://github.com/user-attachments/assets/7bdeb741-ed62-4595-9b21-eb86319065b7

## Acceptance testing

### DIFF
--- a/app/.erb_lint.yml
+++ b/app/.erb_lint.yml
@@ -7,6 +7,9 @@ linters:
   # ErbSafety:
   #   enabled: true
   #   better_html_config: .better-html.yml
+  RequireScriptNonce:
+    # Required to comply with CSP.
+    enabled: true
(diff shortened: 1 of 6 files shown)


tdooner:
Per the conversation [in this Slack thread](https://cmsgov.slack.com/archives/C08B42218F9/p1752613783280849) we do not wish to roll this out at this time.

PR #837, opened 2025-07-16
FFS-2430: Remove inline HTML styles

patsier-cms:
## Ticket

Resolves [FFS-2430](https://jiraent.cms.gov/browse/FFS-2430).

## Changes

- Enables `!important` for USWDS utilities so that we can use them for overrides instead of inline styles
- Adds two utility classes for setting specific properties not included in USWDS
- Replaces inline styles with USWDS or custom utility classes
- Removes a small amount of CSS previously used only because `!important` wasn't enabled for utilities

## Context for reviewers

- The ticket mentions eventually updating the content security policy to disallow inline styles, but I think that could be more complicated here and might require some digging
- The [sessions controller for Pinwheel](https://github.com/DSACMS/iv-cbv-payroll/blob/595407fc0a86ea99283e190e567b68c982e8b76d/app/app/javascript/controllers/cbv/sessions_controller.js#L50-L66) also relies on inline styles, but I left that alone since we're [disabling the CSP for Pinwheel](https://github.com/DSACMS/iv-cbv-payroll/blob/f73a3e5f1f7c74c2d8c9eb8c8e8fc5ef68935930/app/app/controllers/cbv/employer_searches_controller.rb#L2-L3) anyway

## Acceptance testing

Screenshot 2025-07-16 at 9 54 24 AM
Screenshot 2025-07-16 at 9 47 21 AM

### DIFF
--- a/app/app/assets/stylesheets/cbv.scss
+++ b/app/app/assets/stylesheets/cbv.scss
@@ -138,13 +138,6 @@ html {
   @include u-bg('yellow-5v');
 }
 
-// Addressing USWDS utilities specificity
-.usa-table {
-  td.text-bold {
-    font-weight: bold;
-  }
-}
-
 .rotate {
   animation: 1s linear infinite rotate-con;
 }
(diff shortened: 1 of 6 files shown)


GeorgeCodes19 (approved):
I still find it interesting that we have to explicitly allow the CSS `!important` clause with USWDS. Better safe, than sorry 🤷🏾‍♂️

PR #842, opened 2025-07-21
FFS-3082: Distinguish SMS and RFI links for LA users

daphnegold:
## Ticket

Resolves [FFS-3082](https://jiraent.cms.gov/browse/FFS-3082).

## Changes
- `/start` for LA with source "mail"
- Source is "sms" otherwise
- Test

## Context
After discussing with Eryn, we decided to use the events `ApplicantViewedAgreement` and `ApplicantAgreed`. We also discussed and agreed source can be nil for other tracking events that are not triggered by LA.

### Testing
You might want to set the pilot ended config in `client-agency-config.yml` to false for testing the agreement.

1. Visit http://la.localhost:3000/start & check for `mail` as source in events for page view and agreement

1. Visit http://la.localhost:3000 & check for `sms` as source in events for page view and agreement

1. Visit http://localhost:3000/cbv/links/sandbox and check source is `nil`

## Acceptance testing

Screenshot 2025-07-21 at 1 12 51 PM
Screenshot 2025-07-21 at 1 12 38 PM

## Preview environment
♻️ Environment destroyed ♻️

### DIFF
--- a/app/app/controllers/pages_controller.rb
+++ b/app/app/controllers/pages_controller.rb
@@ -32,8 +32,8 @@ class PagesController < ApplicationController
     client_agency_id = detect_client_agency_from_domain
     if client_agency_id.present?
       agency = agency_config[client_agency_id]
-      source = params[:source] || agency&.default_source
-      redirect_to cbv_flow_new_path(client_agency_id: client_agency_id, source: source)
+      session[:cbv_source] = params[:source] || agency&.default_source
+      redirect_to cbv_flow_new_path(client_agency_id: client_agency_id)
     end
   end
 end
--- a/app/app/controllers/cbv/generic_links_controller.rb
+++ b/app/app/controllers/cbv/generic_links_controller.rb
@@ -6,7 +6,6 @@ class Cbv::GenericLinksController < Cbv::BaseController
   def show
     @cbv_flow = CbvFlow.create_without_invitation(params[:client_agency_id])
     session[:cbv_flow_id] = @cbv_flow.id
-    session[:cbv_source] = params[:source]
 
     redirect_to next_path
   end
--- a/app/spec/controllers/cbv/generic_links_controller_spec.rb
+++ b/app/spec/controllers/cbv/generic_links_controller_spec.rb
@@ -8,16 +8,6 @@ RSpec.describe Cbv::GenericLinksController do
         expect(response).to redirect_to(cbv_flow_entry_path)
         expect(assigns(:cbv_flow).client_agency_id).to eq("sandbox")
       end
-
-      it "sets source to SMS when provided" do
-        get :show, params: { client_agency_id: "sandbox", source: "sms" }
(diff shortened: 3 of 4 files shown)


app/.env
```diff
@@ -15,7 +15,7 @@ SUPPORTED_PROVIDERS=pinwheel
 
 DOMAIN_NAME=localhost
 AZ_DES_DOMAIN_NAME=localhost
-LA_LDH_DOMAIN_NAME=localhost
+LA_LDH_DOMAIN_NAME=la.localhost
```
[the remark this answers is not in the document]
GeorgeCodes19: Sorry, this may be on me. I introduced the *_DOMAIN_NAME configuration pattern, but didn't suggest that people update their local DNS resolution. I've updated my `/private/etc/hosts` file on my Mac to have `127.0.0.1	localhost dev.local la.cbv.local sandbox.cbv.local az.cbv.local` as a line item.

app/config/routes.rb
```diff
@@ -25,6 +25,9 @@
     # Feedback CTA
     get "/feedback", to: "feedbacks#show", as: :feedbacks
 
+    # RFI (mail) source tracking route for LA
+    get "/start", to: "pages#home", defaults: { source: "mail" }
```
GeorgeCodes19, line 29: Interesting, TDIL that routes have a `defaults` property.

app/app/controllers/pages_controller.rb
```diff
@@ -31,7 +31,8 @@ def redirect_to_client_agency_entries
 
     client_agency_id = detect_client_agency_from_domain
     if client_agency_id.present?
-      redirect_to cbv_flow_new_path(client_agency_id: client_agency_id)
+      source = params[:source] || ("sms" if client_agency_id == "la_ldh")
```
GeorgeCodes19, line 34: Why have the source default to sms only for LA if it's not present as a query parameter? Couldn't we just do a global default to SMS irregardless of the agency? Also, is this the proper place for agency specific config? It might seem like an overkill, but the `client-agency-config.yml` seems like a sensible place for this default if it's going to be agency specific.

GeorgeCodes19, line 34: The BaseController, EntriesController, and GenericLinksController all extend the `ApplicationController`. Instead of persisting things to the session I wonder if it would make more sense to set an instance variable on the `ApplicationController` i.e. `@source = params[:source]` rather than storing and retrieving the `source` from the session. I don't have an issue with the current approach though.

GeorgeCodes19 (approved):
I had a thought about how we can rework this a bit. However, I'm not sure if my approaches are an overkill since this pattern does accomplish the goal.

I did pull and run the code. The events are all working as expected.

app/.env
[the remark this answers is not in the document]
daphnegold: No worries, maybe I'll just update the rest in this PR.

PR #843, opened 2025-07-22
[Snyk] Fix for 1 vulnerabilities

decause-gov:
snyk-top-banner

#### Snyk changed the following file(s):

- `app/Gemfile`

<details>
<summary>⚠️ <b>Warning</b></summary>

```
Failed to update the Gemfile.lock, please update manually before merging.
```

</details>

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
high severity | OS Command Injection <br/>[SNYK-RUBY-THOR-10843853](https://snyk.io/vuln/SNYK-RUBY-THOR-10843853) | &nbsp;&nbsp;**123**&nbsp;&nbsp; 

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/ospo/project/8530cdfd-48a9-4e1a-84d2-15c64e1ba960?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates?utm_source=github&utm_content=fix-pr-template)
🛠 [Adjust project settings](https://app.snyk.io/org/ospo/project/8530cdfd-48a9-4e1a-84d2-15c64e1ba960?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://docs.snyk.io/scan-with-snyk/snyk-open-source/manage-vulnerabilities/upgrade-package-versions-to-fix-vulnerabilities?utm_source=github&utm_content=fix-pr-template)

---

**Learn how to fix vulnerabilities with free interactive lessons:**

🦉 [OS Command Injection](https://learn.snyk.io/lesson/malicious-code-injection/?loc&#x3D;fix-pr)

[...]

PR #847, opened 2025-07-23
No ticket: Update thor to address vulnerability

patsier-cms:
## Ticket

Fixes vulnerability that's currently failing the build

## Changes

- More granular update of `thor` than #843 if we want to get that in sooner without addressing as many changes

## Context for reviewers

Wanted to open a quick fix to hopefully get this addressed with less review than automated PRs currently

## Acceptance testing

### DIFF
--- a/app/Gemfile
+++ b/app/Gemfile
@@ -81,6 +81,7 @@ gem "gpgme", "~> 2.0", ">= 2.0.12"
 gem "pdf-reader", "~> 2.12.0"
 gem "net-imap", "0.4.20"  # Fixing GHSA-j3g3-5qv5-52m
 gem "cgi", ">= 0.4.2"     # Fixing GHSA-mhwm-jh88-3gjf
+gem "thor", ">= 1.4.0"    # Fixing GHSA-mqcp-p2hv-vw6x
 
 group :development, :test do
   gem "brakeman"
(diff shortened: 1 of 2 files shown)


app/Gemfile
```diff
@@ -81,6 +81,7 @@ gem "gpgme", "~> 2.0", ">= 2.0.12"
 gem "pdf-reader", "~> 2.12.0"
 gem "net-imap", "0.4.20"  # Fixing GHSA-j3g3-5qv5-52m
 gem "cgi", ">= 0.4.2"     # Fixing GHSA-mhwm-jh88-3gjf
+gem "thor", ">= 1.4.0"    # Fixing GHSA-mqcp-p2hv-vw6x
```
GeorgeCodes19, line 84: I don't think is necessary anymore. I've been tracking solid_queue and seems that the Rails team has already [patched this within the last day](https://github.com/rails/solid_queue/releases/tag/v1.2.1).

millerti (approved):
Even if thor isn't necessary, I did a quick runthrough on sandbox and didn't see any issues. LGTM

app/Gemfile
patsier-cms: That's possible, but `bundle update thor` returns "Bundler attempted to update thor but its version stayed the same" and we don't want to update the lockfile directly

PR #853, opened 2025-07-25
FFS-3206: fix aria-label for employer search buttons

jeffcatania:
## Ticket

Resolves [FFS-3206](https://jiraent.cms.gov/browse/FFS-3206).
Fix Issue: Tab selection of employers does not announce the employers name, label is also very long
* https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html
* [https://stackoverflow.com/questions/71695002/form-control-labels-must-be-unique-for-button](https://stackoverflow.com/questions/71695002/form-control-labels-must-be-unique-for-buttons)s

## Changes

* Updates the aria-label to announce employer name instead of long string.

## Context for reviewers

https://github.com/user-attachments/assets/b42231f7-7db7-4bb6-ab9b-f1e5ad217e00

## Acceptance testing


patsier-cms:
Making a note here that we'll need to figure out how we want to handle the vulnerability scan workflows that require access to GitHub secrets that forks can't access. Do those need a Docker Hub login? I think there was some mention of it that resolving a separate issue?

PR #854, opened 2025-07-25
Jeffcatania/ffs 3027 payment details monthly summary

jeffcatania:
## Ticket

## Acceptance testing


daphnegold (approved):
Nice! Shipping pending fixing linter and rubocop, etc. I think there's some opportunity for refactoring shared bits of the new components in the future to dry up the code, using inheritance and composition. But the current implementation is great as is! 👍

A side note for myself, I think I should have broken employment details into a separate ticket... Thanks for taking on this huge update.

PR #855, opened 2025-07-25
FFS-3139: Save CBV Applicant ID into cookie

daphnegold:
## Ticket

Resolves [FFS-3139](https://jiraent.cms.gov/browse/FFS-3139).

## Changes
Who put the cookie in the cookie jar. Me, it was me.

- Add cookie with cbv applicant ID for both generic and token flows
- Test!

## Context for reviewers
It's a permanent cookie for now, based on the spec, until we decide on something we like better. 

## Acceptance testing

# Receipts for cookies in the cookie jar
Screenshot 2025-07-25 at 2 26 34 PM
Screenshot 2025-07-25 at 2 25 53 PM


iannorriswork (approved):
looks reasonable especially with the tests! 

I do wonder if at some point we should think about capturing all the session logging in/out in one set of objects/concerns, but that's for a future refactor :)

PR #857, opened 2025-07-28
CHORE Finish renaming field

iannorriswork:
## Ticket

Resolves the renaming of a field from pinwheel_account_id to aggregator_account_id. 

## Changes

This is under the assumption the [backfill](https://github.com/DSACMS/iv-cbv-payroll/blob/main/app/lib/tasks/database_backfills.rake) has already been run in sandbox and production.

## Acceptance testing

 * Run the db backfill locally, make sure that a previous cbv applicant does not have any issues.

### DIFF
--- a/app/app/components/report/monthly_summary_table_component.rb
+++ b/app/app/components/report/monthly_summary_table_component.rb
@@ -11,7 +11,7 @@ class Report::MonthlySummaryTableComponent < ViewComponent::Base
     @is_pdf = is_pdf
 
     # Note: payroll_account may either be the ID or the payroll_account object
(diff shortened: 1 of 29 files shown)


millerti (approved):
Your documentation answered the only questions I had. I approve. **Note that I do NOT know if this backfill has been run, so please check in demo and prod before you merge!**

tdooner (changes requested):
There are a lot of test failures here that aren't showing up because, for some reason, rspec did not run on the latest commit. I'm going to "Request Changes" on this PR so we don't accidentally merge it in this state.

PR #858, opened 2025-07-28
SPIKE show some options for caching

iannorriswork:
Don't merge this, but it's a way to show some of our options available for caching.

Both of these options require the minimal amount of massaging assuming we're calling the same query API methods over consistently--assumes we're not trying to massage the data or have a common data store modeling of say, the paystubs.

* future iteration on improvements:
right now there's a minor performance implication by storing the requests--could kick that off to a background job instead.

took two approaches:

redis:
advantage in approach one: redis naturally expires things out of the box

postgres via solid_cache:
advantages:
only one database

downsides:
solid cache does not by default expire things until they are called, so if we want to make sure we're not holding onto data unnecessarily, we will need to investigate our own background jobs to evict (see https://github.com/rails/solid_cache?tab=readme-ov-file)


app/db/migrate/20250728143621_create_solid_cache_entries.solid_cache.rb
```diff
@@ -0,0 +1,12 @@
+# This migration comes from solid_cache (originally 20230724121448)
+class CreateSolidCacheEntries < ActiveRecord::Migration[7.0]
+  def change
+    create_table :solid_cache_entries do |t|
+      t.binary   :key,        null: false,   limit: 1024
+      t.binary   :value,      null: false,   limit: 512.megabytes
```
jeffcatania, line 6: If we choose to go down this route, this table should NEVER appear in backups and we should also restrict permission to reading this :value in prod. Operations should be able to invalidate the caches, but inspecting this value exposes pii.

app/app/models/concerns/cacheable_external_requests.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  class_methods do
+    def cache_method(method_name, expires_in: 5.minutes)
+      alias_method "uncached_#{method_name}", method_name
+
+      define_method(method_name) do |**args|
+        if CACHING_STRATEGY == :redis
+          redis_cache_fetch(method_name, args, expires_in: expires_in)
+        elsif CACHING_STRATEGY == :postgres
```
jeffcatania, line 46: I highly recommend we choose Redis over Postgres so that our database remains clear of sensitive PII for the purposes of backups and troubleshooting.

iannorriswork:
i think the PR feedback i got was what i needed, closing in favor of getting a Redis set up properly.

PR #859, opened 2025-07-28
FFS-3115 LA Wind-down spanish translations

jeffcatania:
## Ticket

Resolves [FFS-3115](https://jiraent.cms.gov/browse/FFS-3115).

## Changes

* Added spanish translations for LA wind-down page
* Fixed a couple issues with the import translation script
   * It assumed that you were importing a file with Windows UTF-8 encodings which caused it to break. I turned that off and added a comment.
   * I forced the to_yaml to write single lines instead of multi-line to reduce diffs on PR's.
   
## Acceptance testing

image

### DIFF
--- a/app/config/locales/es.yml
+++ b/app/config/locales/es.yml
@@ -532,10 +532,19 @@ es:
       header: "¡Parece que algo salió mal!"
       refresh: Actualizar página
     home:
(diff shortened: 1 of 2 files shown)


millerti (approved):
LGTM!

PR #875, opened 2025-08-04
No ticket: Experimental: Provide optional sha to deploy

millerti:
## Changes
Adding a SHA field to our deploy action so that we're not limited to deploying the latest commit to main.

## Context for reviewers
I've found it annoying that, after completing smoke testing, I find that someone has committed to main (necessitating more smoke testing before a prod deploy). I would like to update our deploy action so that we can specify a SHA to deploy, falling back to the most recent if blank. I'm open to feedback on whether this is a wise move.

## Testing instructions
0. Go to the Deploy App action in GitHub Actions and click Run Workflow. Pick this branch. You should now see the optional SHA field.
1. Pick a SHA to deploy to dev. I recommend not going back beyond a migration, which might cause problems. 
2. Kick off the workflow.
3. Don't panic if the shortened SHA you see in the UI is not the one you specified -- that's the SHA of the branch you're on. You can open the workflows to make sure that the input is being passed around appropriately. You likely won't need to build and publish it, since that SHA's image is probably still around.
4. Go to the dev site and verify that the SHA you picked is reported: https://la-verify-demo.navapbc.cloud/health
5. Optionally, verify that some of the commits that are now not included are missing on the site (eg I verified that my fix to the help modal where I made sure the "report feedback" button showed up was no longer included).
6. Repeat these tests but _don't_ specify a SHA. Make sure that the system automatically picked the HEAD of the branch.

## Acceptance testing

### DIFF
--- a/app/app/controllers/cbv/payment_details_controller.rb
+++ b/app/app/controllers/cbv/payment_details_controller.rb
@@ -129,7 +129,7 @@ class Cbv::PaymentDetailsController < Cbv::BaseController
       client_agency_id: current_agency&.id,
       invitation_id: @cbv_flow.cbv_flow_invitation_id,
       aggregator_account_id: @payroll_account.id,
-      payments_length: @payroll_account_report.paystubs.length,
+      payments_length: @payroll_account_report.paystubs&.length,
       has_employment_data: has_employment_data?,
       has_paystubs_data: has_paystubs_data?,
       has_income_data: has_income_data?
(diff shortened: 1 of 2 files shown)


.github/workflows/cd-app.yml
```diff
@@ -21,6 +21,9 @@ on:
         options:
           - dev
           - prod
+      commit_sha:
+        required: false
+        type: string
```
tdooner, line 26: Hmm, how about adding a description here that says something like "Override commit SHA to deploy" to indicate it's optional?

PR #878, opened 2025-08-05
Update Rails to 7.2.2.1

patsier-cms:
## Ticket

No ticket

## Changes

- Bumps Rails to `~7.2.2` resolving to 7.2.2.1
- Updates some tests using the `include` matcher to strip out extra space, since it looks like this was changed and caused some tests to fail

## Context for reviewers

- I'm using `~7.2.2` to match what was there previously, but happy to pin to be more strict

## Acceptance testing


patsier-cms:
@daphnegold yep, that was to fix failing tests. It looks like the `include` matcher was updated so that it doesn't strip excess space. If you remove one of those lines and try to run that test file the test fails

PR #879, opened 2025-08-05
consolidating preview methods

GeorgeCodes19:
## Ticket

Resolves [FFS-2768](https://jiraent.cms.gov/browse/FFS-2768).

## Acceptance testing

### DIFF
--- a/app/app/components/memorable_date_component.rb
+++ b/app/app/components/memorable_date_component.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class MemorableDateComponent < ViewComponent::Base
+  def initialize(form:, attribute:, legend:, hint: nil)
+    @form = form
+    @attribute = attribute
+    @legend = legend
+    @hint = hint
+  end
+
+  private
+
+  attr_reader :form, :attribute, :legend, :hint
+
+  def field
+    form.object
+  end
+
+  def field_name
+    form.object_name
+  end
+
+  def date_values
+    raw_value = field.send(attribute)
+    return {} unless raw_value.is_a?(Date)
+
+    {
+      year: raw_value.year,
+      month: raw_value.month,
+      day: raw_value.day
+    }
+  end
+end
--- a/app/app/helpers/uswds_form_builder.rb
+++ b/app/app/helpers/uswds_form_builder.rb
@@ -125,6 +125,13 @@ class UswdsFormBuilder < ActionView::Helpers::FormBuilder
     text_field(attribute, options.merge(value: value, group_options: group_options))
   end
 
+  def memorable_date(attribute, options = {})
+    legend = options.delete(:label)
+    hint = options.delete(:hint)
+
+    @template.render(MemorableDateComponent.new(form: self, attribute: attribute, legend: legend, hint: hint))
+  end
+
   def field_error(attribute)
     return unless has_error?(attribute)
     error_messages = object.errors.messages_for(attribute)
--- a/app/app/models/cbv_applicant.rb
+++ b/app/app/models/cbv_applicant.rb
@@ -51,7 +51,14 @@ class CbvApplicant < ApplicationRecord
   }
 
   def date_of_birth=(value)
-    self[:date_of_birth] = parse_date(value)
+    if value.is_a?(Hash)
+      day = value["day"].to_i
+      month = value["month"].to_i
+      year = value["year"].to_i
+      self[:date_of_birth] = Date.new(year, month, day) rescue nil
+    else
+      self[:date_of_birth] = parse_date(value)
+    end
   end
 
   def snap_application_date=(value)
--- a/app/app/views/layouts/component_preview.html.erb
+++ b/app/app/views/layouts/component_preview.html.erb
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="<%= I18n.locale %>">
+  <head>
+    <title>Component Preview</title>
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <%= csrf_meta_tags %>
+    <%= csp_meta_tag %>
+    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
+    <%= javascript_include_tag "application", "data-turbo-track": "reload", type: "module" %>
+  </head>
+  <body>
+    <div style="padding: 2rem;">
+      <%= yield %>
+    </div>
+  </body>
+</html>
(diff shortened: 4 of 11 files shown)


app/app/models/cbv_applicant.rb
```diff
@@ -51,7 +51,14 @@ def self.valid_attributes_for_agency(client_agency_id)
   }
 
   def date_of_birth=(value)
-    self[:date_of_birth] = parse_date(value)
+    if value.is_a?(Hash)
+      day = value["day"].to_i
+      month = value["month"].to_i
+      year = value["year"].to_i
+      self[:date_of_birth] = Date.new(year, month, day) rescue nil
```
GeorgeCodes19, line 58: If we cannot construct a valid date from the hash then we'll return nil which will cause a validation error to bubble up. This ensures that invalid dates cannot be submitted.

app/spec/components/previews/application_preview.rb
```diff
@@ -0,0 +1,19 @@
...
+  # for the UswdsFormBuilder to function correctly in previews.
+  def view_context
+    # Using ActionController::Base.new.view_context gives us a context
+    # with all the standard Rails helpers loaded.
+    @view_context ||= ActionController::Base.new.view_context.tap do |view_context|
+      # We need to include our own ApplicationHelper to make methods like
+      # `uswds_form_with`, `current_agency`, etc. available.
+      view_context.class.include(ApplicationHelper)
+    end
+  end
```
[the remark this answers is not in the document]
GeorgeCodes19: No problem, I can slim them down. I wasn't aware of what a `view_context` entailed or why we needed any of this for the component previews to work so I leaned towards being a bit verbose.

app/app/views/layouts/component_preview.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <head>
+    <title>Component Preview</title>
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <%= csrf_meta_tags %>
+    <%= csp_meta_tag %>
+    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
+    <%= javascript_include_tag "application", "data-turbo-track": "reload", type: "module" %>
+  </head>
+  <body>
+    <div style="padding: 2rem;">
```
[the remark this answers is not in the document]
GeorgeCodes19: This is only used for previews (in test and development environments). I guess I hadn't considered that we can still use USWDS for our preview templates. The component themselves are using it, after-all. I'll update this to be a class.

app/app/helpers/uswds_form_builder.rb
```diff
@@ -125,6 +125,13 @@ def date_picker(attribute, options = {})
     text_field(attribute, options.merge(value: value, group_options: group_options))
   end
 
+  def memorable_date(attribute, options = {})
+    legend = options.delete(:label)
+    hint = options.delete(:hint)
```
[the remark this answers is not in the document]
GeorgeCodes19: Following convention used elsewhere in the file. 

`.delete` ensures that we're not using the same property twice without referring to the reference created by initially calling `delete`- in conjunction with returning the property value. 

So if we do `legend = options.delete(:label)` then we **can't** reference `options[:label]` again. So we're forced to reference `legend` variable. 

In short- it prevents multiple calls to `options[:legend]` and enforces the usage of destructured variables. A bit of defensive programming that says, "Hey, we're already using this property"

app/app/components/memorable_date_component.rb
```diff
@@ -0,0 +1,33 @@
...
+    @hint = hint
+  end
+
+  private
+
+  attr_reader :form, :attribute, :legend, :hint
+
+  def field
+    form.object
+  end
```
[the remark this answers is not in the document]
GeorgeCodes19: Ahh good catch it's no longer needed. I was using it to permit params in the [applicant_params method of the applicant_informations_controller.rb.](https://github.com/DSACMS/iv-cbv-payroll/blob/b7c852cba6fdda214da97092d45e8634e53f6f42/app/app/controllers/cbv/applicant_informations_controller.rb#L49) It's no longer needed.

app/app/models/cbv_applicant.rb
```diff
@@ -88,9 +95,18 @@ def set_snap_application_date
...
     @applicant_attributes = agency_config&.applicant_attributes&.compact&.keys&.map(&:to_sym) || []
+    @applicant_attributes << { date_of_birth: [ :day, :month, :year ] } if @applicant_attributes.include?(:date_of_birth)
+
     @required_applicant_attributes = get_required_applicant_attributes
   end
 
+  # Reset the applicant attributes to nil by removing any non-symbol keys i.e. { date_of_birth: [ :day, :month, :year ] }
+  # and then setting the attributes to nil.
+  def reset_applicant_attributes
+    clear_attributes = applicant_attributes.reject { |key| !key.is_a?(Symbol) }.index_with(nil)
```
millerti, line 106: I guess I'm a little confused about this. Looks like you're trying to clear out all the values for the symbols in applicant_attributes. But isn't date_of_birth a symbol in your example? What is the non-symbol key you're targeting here? 

Perhaps what I'm confused about is the name of the method, which makes it sound like we're clearing everything, when this line clearly doesn't clear everything. So should the name change, or should we actually clear everything?

app/spec/support/test_helpers.rb
```diff
@@ -110,4 +112,13 @@ def override_supported_providers(new_value)
...
   end
+
+  def extract_pdf_text(response)
+    pdf = PDF::Reader.new(StringIO.new(response.body))
+    pdf_text = ""
+    pdf.pages.each do |page|
+      pdf_text += page.text.gsub(/\s+/, ' ')
+    end
+    pdf_text
+  end
```
millerti, line 123: Nice.

app/app/models/cbv_applicant.rb
GeorgeCodes19: That's handled by `.index_with(nil)` in this instance. We first clear the non-symbol attributes (the hash) and then set all remaining values to `nil` for update the CbvApplicant.

millerti: Here's what I'm worried about, and maybe I'm fundamentally misunderstanding something:

1. rejecting based on |key| !key.is_a?(Symbol) filters out the date_of_birth attribute
2. we call .index_with(nil) on the result, which clears everything but date_of_birth
3. now, all the attributes _except_ dob are cleared... but shouldn't dob be cleared as well?

Sorry if this is getting frustrating but hopefully we can spot any miscommunication with this format

millerti: Based on a conversation we had in slack, I understand what's going on now.

millerti (approved):

PR #885, opened 2025-08-07
FFS-3220: Use most recent source for origin tracking

millerti:
## Changes
Origin tracking used to keep the first received origin and persist it throughout the session, rejecting any subsequent origins. This was found to be unacceptable for go-live, and now we will update the origin any time we receive a different one.

## Context for reviewers
I'm a little paranoid about this change because rejecting future origin differences seems to have been a purposeful decision made by Pat when he implemented [FFS-3138](https://github.com/DSACMS/iv-cbv-payroll/pull/850). Can anyone think of a reason this might have undesirable consequences? Maybe in other workflows? Pat is back tomorrow so I plan to ask for his input on this. I wonder why rejecting origin differences was so important.

### General testing
0. Use the preview environment link below to construct a generic link by appending /en/cbv/links/sandbox to the end of it
1. Then, add an origin by appending ?origin=email
2. Visit our site
3. Check Mixpanel (preview environment events file to the demo instance, not development). Check that your ApplicantClickedGenericLink event has an origin property that matches your origin parameter.
4. Now update your generic link with a different origin, like mfb_documents, and visit again.
5. Check Mixpanel to make sure you get a new ApplicantClickedGenericLink event on the same profile and that its origin matches your updated origin (it should NOT match the first origin from the first link -- previously, it would have).

### Defaults testing
0. Running locally, access http://la.localhost:3000/en/start and verify that the generated events have the origin "mail", not "sms".

## Acceptance testing

## Preview environment
- Service endpoint: http://p-885-app-dev-2118289484.us-east-1.elb.amazonaws.com
- Deployed commit: bd562394debbce0c2fa1e8cf4d12a4dfcdc09637

### DIFF
--- a/app/app/controllers/cbv/base_controller.rb
+++ b/app/app/controllers/cbv/base_controller.rb
@@ -46,9 +46,22 @@ class Cbv::BaseController < ApplicationController
   end
 
   def set_cbv_origin
-    # Running before set_cbv_flow so we need to use the domain
-    agency = agency_config[detect_client_agency_from_domain]
-    origin = params.fetch(:origin, agency&.default_origin)
+    origin_param = params.fetch(:origin, "")
+    if origin_param.present?
+      # If we get a param, use it to overwrite the origin.
+      # This helps us meet the state expectation that somebody who clicks a second link should switch to that origin in our tracking.
+      origin = origin_param
+    elsif origin_param.blank? and session[:cbv_origin].blank?
+      # If we don't get a param, and if we don't already have an origin, regress to the default.
+      # This preserves defaulting behavior.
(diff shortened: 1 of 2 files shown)


daphnegold (changes requested):
Documenting that we found an issue with LA origins because it passes its own `?origin=mail` with `/start` route.

patsier-cms (approved):
Since this has a separate review just flagging I have no concerns with this! If the original implementation is resulting in any unexpected behavior I think this makes sense

daphnegold (approved):
:shipit:

millerti:
Did AC testing with April and she gave the OK!

PR #895, opened 2025-08-14
FFS-3229: make page footer not scrolled off the end of the page

millerti:
## Changes
Footer used to be cut off at the bottom of the page. Now, it won't be.

## Acceptance testing

### DIFF
--- a/app/app/assets/stylesheets/cbv.scss
+++ b/app/app/assets/stylesheets/cbv.scss
@@ -8,6 +8,12 @@ html {
   min-width: 320px;
 }
 
+#root {
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+}
+
 .usa-button--outline-email {
   // Because css rendering in email is fraught, we must replace the default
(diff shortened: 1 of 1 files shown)


app/app/assets/stylesheets/cbv.scss
```diff
@@ -21,7 +27,7 @@ html {
   // at the bottom of the screen by default.
   $header-height: 4.5rem;
   $footer-height: 5rem;
-  min-height: calc(100vh - $header-height - $footer-height);
```
patsier-cms, line 24: It looks like we can get rid of these variables and the comment now that they're no longer being used

patsier-cms (approved):
Confirmed that this works! I left one small comment about cleaning up outdated variables but I verified the change locally

app/app/assets/stylesheets/cbv.scss
millerti: Sorry about that! I tidied up a bit.

PR #900, opened 2025-08-18
Ensure consistent margin-top-5 on "Continue" buttons

tdooner:
## Ticket

Part of FFS-2839 but broken out as it can be accepted separately.

## Changes

* Remove custom styling for `.usa-button.display-block` as it diverges
  from intended button styling in Figma
* Override button spacing to be 5 units to match Figma (instead of the
  default of 3) when inside a .usa-form.

## Screenshots
/help modal:
image

/entry:
image

/payment_details:
image

/add_job:
image

/other_jobs:
(Note: The extra spacing above the button will be fixed in the other PR having to do with usa-prose)
image

/applicant_informations:
image

/summary:
image

/submit:
image

Examples of buttons on mobile:
image
image
image

## Context for reviewers

I chatted with Daniella and it seems desired that we want the "Continue" button to always have 5 units of spacing above it.

## Acceptance testing

## Infrastructure Changes

  - [ ] Plan reviewed

**Risk / Downtime:**


danielladevera-nava:
design comment: might we increase the `margin-bottom` to be an equivalent (visually) the the `margin-top`? 

see tokens as labeled in the figma (spacing on bottom of button, then spacing of the container on mobile: 
image

tdooner:
@danielladevera-nava Ok sure. It looks like this on most pages.

https://github.com/user-attachments/assets/a12174b3-6493-4cb7-a73a-d8cf4c51a12c

I imagine you'll want me to tweak the styling on the /entry page divider so it looks even. Like this:

https://github.com/user-attachments/assets/636d7a56-9169-475b-9a1b-a582bbaf5282

PR #909, opened 2025-08-20
FFS-3283: Backdate events in Mixpanel by using "time" property

tdooner:
## Ticket

Resolves [FFS-3283](https://jiraent.cms.gov/browse/FFS-3283).

## Changes

**FFS-3283: Backdate events in Mixpanel by using "time" property**
> We currently track events with a "timestamp" property, in an attempt to
> capture the original event timestamp.

> For Mixpanel, in order to backdate events properly, we need the event
> name to be "time", as described [in the Mixpanel API docs][1].

> [1]: https://developer.mixpanel.com/reference/track-event

## Acceptance testing

## Infrastructure Changes

  - [ ] Plan reviewed

**Risk / Downtime:**

### DIFF
--- a/app/app/controllers/api/argyle_controller.rb
+++ b/app/app/controllers/api/argyle_controller.rb
@@ -71,6 +71,7 @@ class Api::ArgyleController < ApplicationController
 
   def track_event
     event_logger.track("ApplicantBeganLinkingEmployer", request, {
+      time: Time.now.to_i,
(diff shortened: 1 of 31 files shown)


millerti:
I did a search for .track( and noticed that several calls were missing time. Is that intentional?
Check out payment_details_controller.rb, events_controller.rb, and maybe more.

millerti (approved):
Looking good!

PR #910, opened 2025-08-20
FFS-3214: add w2 monthly summary to pdfs

millerti:
## Changes
* Adds the W2 Monthly Summary component to the pdfs
* Minor changes to the Payments and Deductions list
* Per convo with Rutvi, only show the footnote for the first of each type of employment

## Context for reviewers
We've decided _not_ to use the payments and deductions component on the PDF.

## Acceptance testing

## Screenshots
Example with 2x W2s and 2x gigs. Notice that only the first of each displays a footnote.
[2x w2 2x gig caseworker final final.pdf](https://github.com/user-attachments/files/21945252/2x.w2.2x.gig.caseworker.final.final.pdf)

## Preview environment
♻️ Environment destroyed ♻️

### DIFF
--- a/app/app/components/report/gig_monthly_summary_table_component.erb
+++ b/app/app/components/report/gig_monthly_summary_table_component.erb
@@ -62,7 +62,7 @@
 
     <% if has_monthly_summary_results? && show_payments? %>
       <div class="margin-top-5">
-        <h2><%= I18n.t("components.report.monthly_summary_table.payments_from_header", employer_name: @employer_name) %></h2>
+        <h3><%= I18n.t("components.report.monthly_summary_table.payments_from_header", employer_name: @employer_name) %></h3>
         <p><%= payments_from_text %></p>
         <% if @is_pdf %>
           <ul class="usa-list">


app/app/views/cbv/submits/show.pdf.erb
```diff
@@ -114,15 +115,18 @@
   <% end %>
 
   <% if is_w2_worker %>
+    <% w2_count += 1 %>
+    <%= render(Report::W2MonthlySummaryTableComponent.new(@aggregator_report, account_id, is_caseworker: is_caseworker, is_responsive: false, is_pdf: true, show_footnote: (w2_count == 1))) %>
```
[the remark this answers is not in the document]
millerti: Hmm, maybe I have something to learn here, but I'd prefer the VC not have to know anything about where it's hosted. It should offer the host the ability to turn the footnote on and off, and it's up to the host to decide whether that's appropriate.

Is this an outdated design philosophy? Why would it be better to, for example, pass the VC the aggregator_report?

app/spec/controllers/cbv/submits_controller_spec.rb
```diff
@@ -336,7 +336,7 @@
           pdf_text = extract_pdf_text(response)
 
           expect(response).to be_successful
-          expect(pdf_text).to include("Pay Date")
+          expect(pdf_text).to include("Pay date")
```
tdooner, line 339: Hmm, is there a test case we can make to test that the footnote appears only once?

millerti: Added two! Good suggestion.

PR #912, opened 2025-08-20
Increase solidqueue workers 2 -> 4 for LA launch week

tdooner:
## Ticket

N/A - Followup from minor incident this morning.

## Changes

We [learned this morning][1] that two are not sufficient to keep the
queue from growing. Let's up the count to 4 for the remainder of the LA
messaging week.

[1]: https://cmsgov.slack.com/archives/C08GB9BHSEB/p1755698014230989?thread_ts=1755696907.956919&cid=C08GB9BHSEB

## Acceptance testing

## Infrastructure Changes

  - [x] Plan reviewed
  - [x] Applied in dev before merge
  - [x] Applied in prod after merge (note any exceptions or special coordination below)

**Risk / Downtime:**

### DIFF
--- a/infra/app/app-config/prod.tf
+++ b/infra/app/app-config/prod.tf
@@ -20,7 +20,7 @@ module "prod_config" {
   service_cpu                       = 1024
   service_memory                    = 4096
   service_desired_instance_count    = 20
-  solidqueue_desired_instance_count = 2
+  solidqueue_desired_instance_count = 4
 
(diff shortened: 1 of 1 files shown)


iannorriswork (approved):

tdooner:
I did not `terraform apply` this change because there is no diff to affected resources (as the change has already been applied in production manually).

PR #920, opened 2025-08-22
FFS-3287: Remove "Prompt" row from applicant info table on /summary

daphnegold:
## Ticket

Resolves [FFS-3287](https://jiraent.cms.gov/browse/FFS-3287).

## Changes
Move param concerns to controller, separate from display concerns

## Context for reviewers
The summary page loop included a non symbol date of birth entry meant only for params validation. The translation looked up an invalid key and fell back to the string "Prompt," causing it to show up as a row label.

I think we should follow the convention of separating params and visual going forward, as a rule of thumb.

## Acceptance testing

### Before
Screenshot 2025-08-22 at 12 45 17 PM

### After
Screenshot 2025-08-22 at 2 41 04 PM

## Preview environment
- Service endpoint: http://p-920-app-dev-733372590.us-east-1.elb.amazonaws.com
- Deployed commit: c7f845b38792fccf467b3d5dbea03e735fbe010b

### DIFF
--- a/app/spec/models/cbv_applicant_spec.rb
+++ b/app/spec/models/cbv_applicant_spec.rb
@@ -1,151 +1,165 @@
 require 'rails_helper'
 
 RSpec.describe CbvApplicant, type: :model do
-    describe "all valid types of agencies" do
-      ClientAgencyConfig.client_agencies.client_agency_ids.each do |client_agency_id|
-        it "has a list of VALID_ATTRIBUTES for #{client_agency_id}" do
-          expect(CbvApplicant.valid_attributes_for_agency(client_agency_id)).to be_present
-        end
+  describe "all valid types of agencies" do
+    ClientAgencyConfig.client_agencies.client_agency_ids.each do |client_agency_id|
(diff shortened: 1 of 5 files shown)


app/spec/models/cbv_applicant_spec.rb
```diff
@@ -1,151 +1,165 @@
...
-        invalid_applicant.set_applicant_attributes
-        expect(invalid_applicant.validate_base_and_applicant_attributes?).to be false
+    it "sets applicant and required attributes based on agency config" do
+      cbv_applicant.set_applicant_attributes
+
+      expect(cbv_applicant.applicant_attributes).to include(:first_name, :middle_name, :date_of_birth)
+      expect(cbv_applicant.applicant_attributes).to all(be_a(Symbol))
+
+      expect(cbv_applicant.required_applicant_attributes).to include(:first_name, :date_of_birth)
+      expect(cbv_applicant.required_applicant_attributes).not_to include(:middle_name)
```
tdooner, line 84: Thanks for the test.

tdooner (approved):

PR #924, opened 2025-08-25
No ticket: Run i18n-health on PR pushes on branch, not main

millerti:
## Changes
When we run our i18n-tasks health GitHub Action on push to a PR, it used to run on main instead of the code in the PR. Now, it will run on the code in the PR.

## Acceptance testing


tdooner (changes requested):
I think the correct fix here is to run the Action `on` `pull_request_target` (instead of `pull_request`). See this doc: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

Notice how `pull_request` events result in a run with "GITHUB_REF" set to "refs/pull/PULL_REQUEST_NUMBER/merge", which I believe is the merge commit of the branch with main (hence the problem you saw).

Compare that to `pull_request_target` which runs with "GITHUB_REF" set to "PR base branch", which I think reflects only the work within the PR (as you intend).

millerti:
> I think the correct fix here is to run the Action `on` `pull_request_target` (instead of `pull_request`). See this doc: [docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)
> 
> Notice how `pull_request` events result in a run with "GITHUB_REF" set to "refs/pull/PULL_REQUEST_NUMBER/merge", which I believe is the merge commit of the branch with main (hence the problem you saw).
> 
> Compare that to `pull_request_target` which runs with "GITHUB_REF" set to "PR base branch", which I think reflects only the work within the PR (as you intend).

It seems like pull_request_target is significantly different than on pull_request, like how it now has write permissions. I'd prefer not to expose ourselves to that vulnerability. Is there something about this structure that you don't like? The documentation says to do what I've done:
"Note that GITHUB_SHA for this event is the last merge commit of the pull request merge branch. If you want to get the commit ID for the last commit to the head branch of the pull request, use github.event.pull_request.head.sha instead."

.github/workflows/i18n-health.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
```
tdooner, line 20: Ok. I guess I am fine as long as you got this from somewhere. The only problem I see though is that this workflow also is configured to run on `push` to the `main` and `develop` branches. Will this variable be defined when it runs on a push to `main`?

tdooner (approved):

.github/workflows/i18n-health.yml
millerti: You're right that it'll be null in that case, and my understanding is that the default behavior is to then use the commit that was just pushed.

PR #930, opened 2025-08-28
FFS-3219 & FFS-3329: /success shareable link updates (simplified with origin)

daphnegold:
## Ticket

Resolves [FFS-3219](https://jiraent.cms.gov/browse/FFS-3219).
Resolves [FFS-3329](https://jiraent.cms.gov/browse/FFS-3329).

## Changes
- Introduces adding an origin parameters in our url generating helper methods
- /success will append `origin=shared` to shareable links
- Shareable links are simplified and will prioritize using just domain rather than path if available
- Update tests

## Acceptance testing


app/app/models/cbv_flow.rb
```diff
@@ -38,15 +38,22 @@ def has_account_with_required_data?
...
       locale: I18n.locale
-    }.compact)
+    }
+    url_params[:origin] = origin if origin.present?
+
+    if client_agency.agency_domain.present?
+      Rails.application.routes.url_helpers.root_url(url_params.compact)
+    else
+      url_params[:client_agency_id] = client_agency_id
+      Rails.application.routes.url_helpers.cbv_flow_new_url(url_params.compact)
```
daphnegold, line 56: We don't have any agencies without domains, so we will never hit this. I left it as a safety fallback or backwards compatibility, but I'm also fine getting rid of it and always relying on domain detection.

tdooner (approved):
Thanks for improving the tests!

PR #932, opened 2025-08-29
Fix AutoTitleTestHelper to use HTML parser instead of regex

tdooner:
## Ticket

N/A - This will prevent failures on my branch where I'm working on [FFS-2543](https://jiraent.cms.gov/browse/FFS-2543).

## Changes

Obligatory reference to the best StackOverflow answer ever:
https://stackoverflow.com/a/1732454

This came up because I'm about to add an `<h1>` tag that has a class on
it, which won't match the H1_REGEX here.

## Acceptance testing

## Infrastructure Changes

  - [ ] Plan reviewed

**Risk / Downtime:**


daphnegold (approved):
The pull request looks exactly as it is supposed to look - there are no problems with its content.

PR #933, opened 2025-08-29
FFS-2543: Use JS to show help alert on /employer_search without refresh

tdooner:
## Ticket

Resolves [FFS-2543](https://jiraent.cms.gov/browse/FFS-2543).

## Changes

**Use JS to show "help" alert on /employer_search without refresh**
> * Remove the poorly-factored `check_help_param` method from
>   ApplicationController.
> * Remove support for the `alert_type` flash param since this was the
>   only place that wasn't an "error" type.
> * Render the "help" alert on the /employer_search page by default in a
>   hidden state and configure Stimulus to reveal it after an aggregator
>   modal is closed.
> * Tweak styling on the `<h1>` as they are not expecting to have anything
>   above them, and the padding is excessive when the alert is hidden.

## Acceptance testing

## Infrastructure Changes

  - [ ] Plan reviewed

**Risk / Downtime:**


tdooner:
Eryn acceptance tested this on Friday. ✅

PR #935, opened 2025-08-29
FFS-3209: Reconfigure database backups

daphnegold:
## Ticket

Resolves [FFS-3209](https://jiraent.cms.gov/browse/FFS-3209).

## Changes
- AWS Backup from 21 to 14 days, disabled for dev
- AWS Aurora RDS retention period from 1 -> 7 for prod, 0 for dev

## Acceptance testing

## Infrastructure Changes

  - [x] Plan reviewed
  - [x] Applied in dev before merge

### DIFF
--- a/infra/modules/database/backups.tf
+++ b/infra/modules/database/backups.tf
@@ -4,11 +4,13 @@
 # Backup plan that defines when and how to backup and which backup vault to store backups in
 # See https://docs.aws.amazon.com/aws-backup/latest/devguide/about-backup-plans.html
 resource "aws_backup_plan" "backup_plan" {
+  count = var.enable_aws_backup ? 1 : 0
+
   name = "${var.name}-db-backup-plan"
 
   rule {
     rule_name         = "${var.name}-db-backup-rule"
-    target_vault_name = aws_backup_vault.backup_vault.name
+    target_vault_name = aws_backup_vault.backup_vault[0].name
     schedule          = "cron(0 7 * * ? *)" # Run daily at 7am UTC (2am EST)
 
     lifecycle {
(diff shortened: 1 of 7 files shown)


infra/modules/database/variables.tf
```diff
@@ -81,11 +81,19 @@ variable "vpc_id" {
 variable "serverless_min_capacity" {
   description = "The minimum capacity for the Aurora Serverless cluster. Value should be in Aurora capacity units (ACU)"
   type        = number
-  default     = 0.5
```
tdooner, line 84: Why remove these? We only set a value for this in prod, I think, so these would still be useful in our demo environment.

infra/modules/database/backups.tf
```diff
@@ -4,22 +4,26 @@
...
 
   rule {
     rule_name         = "${var.name}-db-backup-rule"
-    target_vault_name = aws_backup_vault.backup_vault.name
+    target_vault_name = aws_backup_vault.backup_vault[0].name
     schedule          = "cron(0 7 * * ? *)" # Run daily at 7am UTC (2am EST)
 
     lifecycle {
-      delete_after = 21
+      delete_after = 14
```
tdooner, line 17: should this be...

```suggestion
      delete_after = var.backup_retention_period
```

How does this interact with the RDS resource's `backup_retention_period` attribute? 🤔

infra/modules/database/variables.tf
```diff
@@ -81,11 +81,19 @@ variable "vpc_id" {
...
 }
 
 variable "serverless_max_capacity" {
   description = "The maximum capacity for the Aurora Serverless cluster. Value should be in Aurora capacity units (ACU)"
   type        = number
-  default     = 1.0
+}
+
+variable "backup_retention_period" {
+  description = "The number of days to retain backups for. Set to 0 to disable automated backups."
```
tdooner, line 92: Is this description accurate (that setting it to zero will disable backups)? It seems like maybe you added the `enable_aws_backup` option? If zero is actually a valid value that can disable the backups, let's just use that. If not, let's remove this from the comment.

tdooner (approved):

infra/modules/database/variables.tf
daphnegold: The defaults are in the env-config

PR #936, opened 2025-08-30
FFS-3197: Reenable turbo on /employer_search by fixing style issue in CSP

tdooner:
## Ticket

Resolves [FFS-3197](https://jiraent.cms.gov/browse/FFS-3197).

## Changes

**FFS-3197: Reenable turbo on /employer_search by fixing style issue in CSP**
> We've had to special-case the /employer_search page from the rest of our
> page navigations due to the Content Security Policy. Since Argyle and
> Pinwheel both add `<style>` tags to the page, we need some way to ensure
> those are allowed.

> Previously, we just made sure the page was served without a CSP header.
> This was brittle (we had to remember to disable turbo navigation on any
> link that led to /employer_search so that the page would actually be
> re-requested without the header) and less secure than just properly
> allowing these `<style>` tags through the CSP.

> The most straightforward way is to just pin the hash of the `<style>`
> tag contents. This commit, primarily, does that. (This will break
> whenever we update the Pinwheel/Argyle SDKs, but I've provided
> instructions on how to get the new hashes at that time.) I also had to
> add Argyle's domain to the script-src and connect-src.

> Since that allows for re-enabling of Turbo, this commit also does that,
> resulting in a much faster and smooth experience going between these
> pages.

## Context for reviewers

I tested this locally:
* Arriving at the /employer_search page via Turbo -> both modals work ✅ 
* Refreshing the /employer_search page (no Turbo) -> both modals work ✅ 
* Arriving at the /employer_search page from /add_jobs -> both modals work ✅ 

Since this is pretty hard to write an automated test to capture, I propose we lean on our smoke testing process to catch any potential issues with this.

## Acceptance testing

## Infrastructure Changes

  - [ ] Plan reviewed

**Risk / Downtime:**

### DIFF
--- a/app/app/controllers/cbv/employer_searches_controller.rb
+++ b/app/app/controllers/cbv/employer_searches_controller.rb
@@ -1,6 +1,4 @@
 class Cbv::EmployerSearchesController < Cbv::BaseController
-  # Disable CSP since Pinwheel relies on inline styles
-  content_security_policy false, only: :show
   before_action :check_webhooks_initialization_in_development
   after_action :track_accessed_search_event, only: :show
   after_action :track_applicant_searched_event, only: :show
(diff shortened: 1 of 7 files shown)


tdooner:
I further tested this in a PR Review App, and it seems to work. Between this, and our normal smoke testing process, I am hopeful the risk has been mitigated as much as reasonable.

PR #959, opened 2025-09-18
FFS-3044: Retry failed jobs

bencalegari:
## Ticket

Resolves [FFS-3044](https://jiraent.cms.gov/browse/FFS-3044).

## Changes
Retry failed jobs 2 extra times (3 attempts total). Runs retry attempts [polynomially](https://apidock.com/rails/ActiveJob/Exceptions/ClassMethods/retry_on).

- Fix job_class and failed_at values when recording job failure

## Context for reviewers
- Couldn't figure out a way to write an rspec test showing that it was actually re-enqueueing but confirmed manually through the hitting the `/health/test_queueing endpoint` and checking `/jobs` and counting the attempt in the Finished jobs tab.

## Acceptance testing
- [] Acceptance testing after merge

## Infrastructure Changes
N/A

**Risk / Downtime:**
Job idempotency, doesn't appear to be an issue when looking through `ApplicationJob` implementers.

### DIFF
--- a/app/app/jobs/application_job.rb
+++ b/app/app/jobs/application_job.rb
@@ -1,4 +1,6 @@
 class ApplicationJob < ActiveJob::Base
+  retry_on Exception, wait: :polynomially_longer, attempts: 3
+
   def event_logger
     @event_logger ||= GenericEventTracker.new
   end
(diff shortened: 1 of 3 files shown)


app/app/jobs/application_job.rb
```diff
@@ -1,19 +1,17 @@
...
-    job_class = job || "UnknownJob"
-
     NewRelic::Agent.record_custom_event("SolidQueueJobFailed", {
-      job_class: job_class,
-      queue_name: queue_name,
-      failed_at: Time.now,
+      job_class: (self.class.name || "UnknownJob"),
+      queue_name: (self.queue_name || "default"),
+      failed_at: Time.now.to_s,
       error_class: error.class.name,
```
daphnegold, line 15: Moving the variables is just stylistic preference, yes? But I don't think the `to_s` is necessary here, as the JSON serializer will turn the `Time` object into a string anyway (what was happening before).

app/app/jobs/application_job.rb
```diff
@@ -1,19 +1,17 @@
 class ApplicationJob < ActiveJob::Base
+  retry_on Exception, wait: :polynomially_longer, attempts: 3
```
daphnegold, line 2: It's cool we can just use the top level `Exception` here and that works

app/app/jobs/application_job.rb
bencalegari: Yup the variable moving was all style but the `.to_s` turned out to be necessary. I noticed in the New Relic explorer that `Time.now` does not shows up as an abstract object rather than the time itself. Surprisingly `record_custom_event` doesn't stringify any values. 
Screenshot 2025-09-18 at 12 22 12 PM

PR #969, opened 2025-09-26
FFS-3394: send base64 encoded report in JSON

millerti:
## Acceptance testing

## Infrastructure Changes

  - [ ] Plan reviewed

**Risk / Downtime:**


bencalegari (changes requested):
I accepted it by sending it to the `api-test`, copying the post data and using rails console to save it to a pdf. One difference between the Base64 and downloaded versions was the Base64 version did not show the header image. I see in the `show.pdf.erb` template it checks for the presence of `current_agency` before rendering it, and that may be missing from the context since its not passed in. 

Downloaded client version:
Image

Base64 encoded caseworker version:
Image

PR #990, opened 2025-09-29
FFS 3356: Set account_count field based only on fully synchronized payroll accounts

j-shilling:
## Ticket

Resolves [FFS-3356](https://jiraent.cms.gov/browse/FFS-3356).

## Changes

Creates a new `#fully_synced_payroll_accounts` method on the `CbvFlow` model which filters `#payroll_accounts` based on their `#has_fully_synced?` property.

## Context for reviewers

The case_worker_transmitter_job task triggers a ApplicantSharedIncomeSummary event, which includes an `account_count` property. This property had been set by the count of all payroll accounts associated with the current `CbvFlow`;however, that left an edge case where a user might start to add an account, but fail to finish (by not being able to authenticate completely or similar). These partial accounts were counted.

## Acceptance testing

  * This change is verified by two unit tests: 1) asserts that the new CbvFlow method correctly filters payroll accounts, 2) asserts that the account_count value is set from the new method

## Infrastructure Changes

  - [ ] Plan reviewed

**Risk / Downtime:**

### DIFF
--- a/app/spec/jobs/case_worker_transmitter_job_spec.rb
+++ b/app/spec/jobs/case_worker_transmitter_job_spec.rb
@@ -64,8 +64,9 @@ RSpec.describe CaseWorkerTransmitterJob, type: :job do
 
       expect(fake_event_logger).to have_received(:track)
         .with("ApplicantSharedIncomeSummary", anything, include(
-          cbv_flow_id: cbv_flow.id,
-          flow_started_seconds_ago: 10.minutes.to_i,
+                cbv_flow_id: cbv_flow.id,
+                flow_started_seconds_ago: 10.minutes.to_i,
+                account_count: cbv_flow.fully_synced_payroll_accounts.count
         ))
     end
   end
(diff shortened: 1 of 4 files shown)


app/spec/jobs/case_worker_transmitter_job_spec.rb
```diff
@@ -64,8 +64,9 @@
 
       expect(fake_event_logger).to have_received(:track)
         .with("ApplicantSharedIncomeSummary", anything, include(
-          cbv_flow_id: cbv_flow.id,
-          flow_started_seconds_ago: 10.minutes.to_i,
+                cbv_flow_id: cbv_flow.id,
+                flow_started_seconds_ago: 10.minutes.to_i,
+                account_count: cbv_flow.fully_synced_payroll_accounts.count
```
bencalegari, line 69: minuscule style nit: we should adjust the indentation back to the previous 2 spaces, it'll just get switched back next time someone else auto-aligns the file so that'll avoid unnecessary diffs in the future.

PR #993, opened 2025-09-30
FFS-3419: Allow all Tokenized Links to be reused, including in LA

tdooner:
## Ticket

Resolves [FFS-3419](https://jiraent.cms.gov/browse/FFS-3419).

## Changes

**Always allow invitation reuse**
> We had a setting, allow_invitation_reuse, which would enable a tokenized
> link to be reused. We recently forgot to set this value to true for LA,
> since we've updated our product thinking such that this behavior is the
> default.

> As such, let's simplify the logic here so that these links can always be
> reused.

## Context for reviewers

I am not quite sure we want to do this from a product perspective. I will verify that as part of acceptance testing.

## Acceptance testing

### DIFF
--- a/app/app/controllers/cbv/base_controller.rb
+++ b/app/app/controllers/cbv/base_controller.rb
@@ -19,15 +19,6 @@ class Cbv::BaseController < ApplicationController
         return redirect_to(cbv_flow_expired_invitation_path(client_agency_id: invitation.client_agency_id))
       end
 
-      # using invitation.client_agency_id directly instead of current_agency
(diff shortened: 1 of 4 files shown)


tdooner:
I [discussed in Slack](https://cmsgov.slack.com/archives/C08BBUALWQN/p1759337572525029) with Ming. Although we haven't completely agreed on the fate of the configuration option, she does agree that reusable should be the default.

I'm going to merge this. If we end up deciding that this should still be configurable, I'll bring it back.

PR #994, opened 2025-10-01
Fix flaky test in submits controller spec

tdooner:
## Ticket

N/A

## Context for reviewers

This test was flaky because the account ID being unusually short
resulted in the request stub (`argyle_stub_request_account_response`)
not successfully stubbing the `/account/<id>` API request. This caused VCR
to raise an exception, which caused the report fetching to fail when the
shorter account_id was fetched.

This worked most of the time, because the longer account_id was usually
fetched first. But since we freeze time, sometimes,
nondeterministically, the shorter account_id was fetched first, causing
the report fetching to crash and leave the data fetched empty.

Anyway, the fix is simple: use UUIDs of the proper format so that
regardless of which account is fetched first, it returns the same data.

## Acceptance testing


millerti (approved):
2

PR #999, opened 2025-10-01
FFS-N/A: Fix transmitter signature generation and make API receiver Rails agnostic

daphnegold:
## Ticket
N/A

## Changes
- Make receiver Rails agnostic
- Documentation in README.md for us, new comment at top of receiver file
- Remove `bin/api-test`, let's just run it via `ruby` command in a standard way
- Fix bug where signature generation was failing because it depends on `body`

## Context for reviewers
Test with instructions in README.md

## Acceptance testing

### DIFF
--- a/app/spec/services/transmitters/json_transmitter_spec.rb
+++ b/app/spec/services/transmitters/json_transmitter_spec.rb
@@ -59,8 +59,12 @@ RSpec.describe Transmitters::JsonTransmitter do
   end
 
   context 'signature generation' do
-    it 'generates and sends signature headers' do
-      expect(JsonApiSignature).to receive(:generate).and_return("mock-signature")
+    it 'generates signature with the request body' do
+      expect(JsonApiSignature).to receive(:generate).with(
+        a_string_including(cbv_flow.confirmation_code),
+        anything,
+        anything
+      ).and_return("mock-signature")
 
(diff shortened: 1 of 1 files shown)


app/app/services/transmitters/json_transmitter.rb
```diff
@@ -45,6 +35,17 @@ def deliver
     end
 
     req.body = payload.to_json
+
```
daphnegold, line 38: This must be below `req.body = payload.to_json`

bencalegari (approved):

app/app/services/transmitters/json_transmitter.rb
```diff
@@ -45,6 +35,17 @@ def deliver
...
+    timestamp = Time.now.to_i.to_s
+    req["X-IVAAS-Timestamp"] = timestamp
+    req["X-IVAAS-Signature"] = JsonApiSignature.generate(req.body, timestamp, api_key_for_agency)
+
+    custom_headers = @current_agency.transmission_method_configuration["custom_headers"]
+    if custom_headers
+      custom_headers.each do |header_name, header_value|
+        req[header_name] = header_value
+      end
+    end
```
tdooner, line 48: Hmm, kind of concerning that a unit test didn't catch this.

tdooner (approved):

PR #1000, opened 2025-10-02
Log response body when Income Report API receives an error

tdooner:
## Ticket

N/A

## Context for reviewers

For LA, when a validation error occurs, they return a response body that
looks like:

  {"Exception":"API Gateway encountered an error. Error Message:  API Specification validation failed. Request headers do not match with API headers. Header AccountCode OR Accountcode is missing or invalid in the request. Request Details: Service - CBVIncomeReport_RVS, Operation - /api/v1/income-report, Invocation Time:1:21:24 PM, Date:Oct 2, 2025,  Client IP - XX.X.XX.XXX, User - XXXXXXX and Application:XXXXXXXX"}

Although we could theoretically parse the exception out, let's start by
logging it.

I also consolidated the 500 behavior with the other errors, since I
think we'll want to treat those similarly.

## Acceptance testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,16 +103,12 @@ To run database migrations on the test environment that is used by rpec tests, r
    rails 'users:create_api_token[agency_name]'
    ```
 
-2. **Run the test receiver:**
+2. **Run the standalone test receiver:**
    ```bash
-   # For sandbox (default)
-   bin/api-test
-
-   # For a different agency
-   JSON_API_AGENCY=agency_name bin/api-test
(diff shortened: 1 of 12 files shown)


app/spec/fixtures/vcr_cassettes/json_transmitter_418.yml
```diff
@@ -28,5 +28,5 @@ http_interactions:
       - keep-alive
     body:
       encoding: UTF-8
-      string: "I'm a teapot"
+      string: "Here is my handle, here is my spout."
```
bencalegari, line 31: Nice refactor

bencalegari (approved):

PR #1004, opened 2025-10-02
FFS-3371: Normalize `compensation_unit` in Income response object

daphnegold:
## Ticket

Resolves [FFS-3371](https://jiraent.cms.gov/browse/FFS-3371).

## Changes
- Normalize frequency in `income.rb`, specifically for Pinwheel
- Remove duplicates in `en.yml` & `es.yml`

## Context for reviewers
I looked into separate normalization for `compensation_unit` and `pay_frequency` and I just didn't care for it and the duplication. But I am happy to add that in quickly if we'd like.

Argyle values are the default in all cases. We *could* normalize those values as well, defensively.

## Acceptance testing

### DIFF
--- a/app/config/locales/es.yml
+++ b/app/config/locales/es.yml
@@ -50,7 +50,7 @@ es:
       per_mile: Por milla
       quarterly: Trimestralmente
       semimonthly: Dos veces al mes
-      semiweekly: Semi-semanal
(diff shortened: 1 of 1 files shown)


millerti (approved):
Looks right to me as soon as you add the missing translation

PR #1371, opened 2026-02-11
Revert "No ticket: reverted deductions in json (#1367)"

j-shilling:
This PR is here to create a temp environment with deductions that LA can use for their UAT testing when ready

## Preview environment
- Service endpoint: http://p-1371-app-demo-894761251.us-east-1.elb.amazonaws.com
- Deployed commit: af7c1b368420cb2168307bc9f43352503e20b0eb

### DIFF
--- a/.github/workflows/ci-app-pr-environment-checks.yml
+++ b/.github/workflows/ci-app-pr-environment-checks.yml
@@ -10,6 +10,15 @@ on:
         description: "Commit hash to deploy"
         required: true
         type: string
+      environment:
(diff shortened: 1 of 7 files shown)


millerti (approved):
Only one small suggestion (optional). And I guess my only minor concern is that this is a few distinct things all wrapped together... but I know we're under a deadline.

bencalegari:
@j-shilling Closing this, pretty sure it's totally out of date at this point.

PR #1440, opened 2026-03-03
Run migrations on PR environments

bencalegari:
Update `bin/update-pr-environment` to run migrations on PR environments. Prior to this, the PR env workflow was just building the image and deploying it in Terraform without it. 

## Changes
- `run-command` now takes an optional workspace flag so we can execute migrations and anything else we'd like on different environments.

## Acceptance testing

## Preview environment
♻️ Environment destroyed ♻️


tdooner (changes requested):
We should make it create a separate copy of the database for the review app, too, or else it's going to break Dev when we delete a column (or something like that) in a review app.

app/config/database.yml
```diff
@@ -140,6 +141,7 @@ production:
   password: <%= ENV.fetch("DB_PASS") { nil } %>
   port: <%= ENV.fetch("DB_PORT") { 5432 } %>
   host: <%= ENV.fetch("DB_HOST") { "localhost" } %>
+  schema_search_path: <%= ENV.fetch("DB_SCHEMA", "app") %>
```
tdooner, line 144: The default schema is "public", right? So will we need this to be set to "public" in deployed environments in order for the tables to continue being found?

app/lib/tasks/schema.rake
```diff
@@ -0,0 +1,23 @@
+namespace :db do
+  desc "Create the PostgreSQL schema specified by DB_SCHEMA"
+  task create_schema: :environment do
+    schema = ENV["DB_SCHEMA"]
+    if schema.present? && schema != "public"
```
tdooner: Outputting an error message or raising an exception might be a better developer experience here than silently doing nothing.

tdooner (approved):
I think this looks good, but I'm a bit concerned about the deploy where we have to initially change the schema from our default "public" schemas into the "app" schema namespace. Let's make sure we test that, or maybe make it just fall back to the "public" schema instead of "app"?

Also, in order for Argyle to work, we will need to register separate webhooks for each PR app, which will probably also necessitate fixing the TLS misconfiguration for these review apps.

app/config/database.yml
bencalegari: We already set it [to `app` here.](https://github.com/DSACMS/iv-cbv-payroll/blob/eec026f37028a704fa38f3621ba6dc7131dfa6b8/infra/app/app-config/env-config/database.tf#L7)

I confirmed it in the terraform console too just to be safe:
```
terraform -chdir=infra/app/app-config console
> local.environment_configs.dev.database_config.schema_name
"app"
> local.environment_configs.demo.database_config.schema_name
"app"
> local.environment_configs.prod.database_config.schema_name
"app"
```

bencalegari:
> Also, in order for Argyle to work, we will need to register separate webhooks for each PR app, which will probably also necessitate fixing the TLS misconfiguration for these review apps.

Totally, I'm going to take a crack at that in another PR once this is merged.

PR #1564, opened 2026-04-03
FFS-3980: Limit NH DHHS transmission concurrency to 1

bencalegari:
## Changes
Send 1 NH application at a time

## Context for reviewers
- The `duration` setting is for If a job crashes due to worker process failure (like during a deploy) SolidQueue forcibly expires a job after 2 minutes/

## Acceptance testing
- [x] Acceptance testing locally

## Preview environment
♻️ Environment destroyed ♻️


millerti (approved):
Honestly it looks great, just wondering how we can test it effectively before it hits prod.

PR #1580, opened 2026-04-09
Automatically create PRs to Autofix anchore/trivy failures

bencalegari:
## Changes
Trivy and Anchore scans were failing because fixable OS-level CVEs appeared in the container image between Dependabot's weekly base image updates so we have failing builds way more often than we need to. Until now. This new action will create a PR with the patched Dockerfile, all we have to do is review that PR, close and reopen it to trigger CI confirming its been patched successfully, then merge it. 

## Context for reviewers
- Added .github/workflows/auto-fix-vulnerabilities.yml — new workflow that detects fixable OS-level CVEs in the Docker image via Trivy and automatically opens a PR patching the Dockerfile's apt-get install block
- Added .github/scripts/patch-dockerfile.py — bidirectional Dockerfile patcher; adds packages Trivy flags as fixable, removes them when a base image update resolves them
- Updated .github/dependabot.yml — added github-actions ecosystem to keep action versions pinned (flags trivy-action@master and similar floating refs)
- The check-compliance-with-tfsec job from ci-infra.yml [failed on this PR](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/24170738252/job/70541754428) but since `tfsec` is [deprecated and merged into trivy](https://github.com/aquasecurity/tfsec?tab=readme-ov-file#-tfsec-to-trivy-migration) there's no reason to have both so I just removed it.

When triggered (on CI Vulnerability Scans failure, weekly schedule, or manually via workflow_dispatch), it:
- Builds the Docker image and runs Trivy with JSON output
- Identifies OS packages with available fixes
- Patches the Dockerfile BASE stage using the existing CVE-fix pattern (same pattern as the imagemagick block)
- Opens a PR with the fixes and tagging the person whose run triggered the failure to review it, linking to that run.

See an example here of the [PR it creates.](https://github.com/DSACMS/iv-cbv-payroll/pull/1583)

## Acceptance testing
In addition to this PR, check out [the autofix PR created by these workflows](https://github.com/DSACMS/iv-cbv-payroll/pull/1583) and if it looks correct then we're good! 

**Risk / Downtime:**
Potentially too much noise with a lot of CVE's but that's tweakable. The changes are easy to evaluate on their own.


tdooner:
Can we make it so it doesn't pin the versions explicitly? Otherwise - I can't wait for this. And we should run it every morning or something.

.github/workflows/auto-fix-vulnerabilities.yml
```diff
@@ -0,0 +1,169 @@
...
+
+on:
+  workflow_run:
+    workflows: ["CI Vulnerability Scans"]
+    types: [completed]
+  workflow_dispatch:
+  schedule:
+    - cron: '0 9 * * *' # Daily 9am UTC
+  # Temporary: allows testing from a feature branch before merging to main.
+  # Remove this trigger once the workflow has been validated on main.
```
tdooner: Should this `pull_request` target be removed now?

.github/workflows/ci-infra.yml
```diff
@@ -79,17 +79,3 @@ jobs:
...
-    permissions:
-      contents: read
-      pull-requests: write
-
-    steps:
-      - uses: actions/checkout@v4
-      - name: Run tfsec check
-        uses: aquasecurity/tfsec-pr-commenter-action@v1.3.1
-        with:
-          github_token: ${{ github.token }}
```
tdooner, line 95: Good riddance.

.github/workflows/auto-fix-vulnerabilities.yml
```diff
@@ -0,0 +1,169 @@
...
+      - .github/workflows/auto-fix-vulnerabilities.yml
+      - .github/scripts/patch-dockerfile.py
+
+# Prevent parallel runs from creating conflicting Dockerfile edits or duplicate PRs.
+concurrency:
+  group: auto-fix-vulnerabilities
+  cancel-in-progress: false
+
+permissions:
+  contents: write
```
tdooner, line 17: Does this need `contents` permission? I would totally believe it does, for some reason, but I'd think it should be ok with just PR access?

tdooner (approved):
Can't wait to try it (says the Tom with a currently-failed PR branch because of trivy scan). This is kind of complex/a lot, but it seems like it's worth a try.

.github/workflows/auto-fix-vulnerabilities.yml
bencalegari: [Yes we do.](https://github.com/peter-evans/create-pull-request#token) Not ideal but it's what allows the action to create the commit itself. No matter what it can't make it into the repo without our manual approval and the changes it makes are minimal so I feel comfortable enough with that.

.github/workflows/auto-fix-vulnerabilities.yml
bencalegari: Thanks, missed this when I removed it the PR stuff a second time.

PR #1606, opened 2026-04-14
FFS-4079: Fix expiring links early

bencalegari:
## Changes
- Fix invitation retention so invitation records are redacted 7 days after `expires_at`, rather than being redacted early when a completed CBV flow hits its transmitted-flow retention window.
- Preserve `auth_token` when redacting `CbvFlowInvitation` so previously-issued links still resolve and show the expired invitation page instead of falling back to the generic flow.

## Acceptance testing
- [x] Acceptance testing in PR Environment (see comment)

## Preview environment
♻️ Environment destroyed ♻️

### DIFF
--- a/app/app/services/data_retention_service.rb
+++ b/app/app/services/data_retention_service.rb
@@ -1,7 +1,7 @@
 # This class is responsible for redacting data on all models in accordance with
 # our data retention policy.
 class DataRetentionService
-  # Redact unstarted and incomplete invitations 7 days after they expire
+  # Redact invitations 7 days after they expire
   REDACT_UNUSED_INVITATIONS_AFTER = 7.days
 
   # Redact transmitted CbvFlows 7 days after they are sent to caseworker
@@ -19,13 +19,13 @@ class DataRetentionService
 
   def redact_invitations
     CbvFlowInvitation
-      .unstarted
       .unredacted
+      .includes(:cbv_applicant, :cbv_flows)
       .find_each do |cbv_flow_invitation|
         next unless Time.current.after?(cbv_flow_invitation.expires_at + REDACT_UNUSED_INVITATIONS_AFTER)
 
         cbv_flow_invitation.redact!
-        cbv_flow_invitation.cbv_applicant&.redact!
+        cbv_flow_invitation.cbv_applicant&.redact! if cbv_flow_invitation.cbv_flows.none?
       end
   end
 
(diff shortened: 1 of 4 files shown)


bencalegari:
Screenshot 2026-04-14 at 9 25 46 AM

app/app/services/data_retention_service.rb
```diff
@@ -19,13 +19,13 @@ def redact_all!
...
     CbvFlowInvitation
-      .unstarted
       .unredacted
+      .includes(:cbv_applicant, :cbv_flows)
       .find_each do |cbv_flow_invitation|
         next unless Time.current.after?(cbv_flow_invitation.expires_at + REDACT_UNUSED_INVITATIONS_AFTER)
 
         cbv_flow_invitation.redact!
-        cbv_flow_invitation.cbv_applicant&.redact!
+        cbv_flow_invitation.cbv_applicant&.redact! if cbv_flow_invitation.cbv_flows.none?
```
tdooner, line 28: Not sure I understand why we don't want to redact the `cbv_applicant` in this situation. I understand the removal of `unstarted` since this method will now become the only way that CbvFlowInvitation's are redacted... but shouldn't it cascade to the cbv_applicant still?

tdooner (approved):

app/app/services/data_retention_service.rb
bencalegari: We could redact it here but it'll get redacted in `redact_incomplete_cbv_flows`. This way we don't redact it twice although I suppose there's no harm in it.

bencalegari:
I patched `bin/run-command` to accept a workspace and ran some ruby in there to accept it. I'll push that patch as a separate change but here's what I did: 

- Created a user, applicant, invitation, then a flow from that invitation.
- Set the flow to have been transmitted 8 days ago. 
- Manually ran `DataRetentionService.new.redact_complete_cbv_flows` and checked that the invitation had not been redacted and opened the link successfully.
- Set the invite's `expired_at` to 8 days ago. 
- Ran `DataRetentionService.new.redact_invitations`
- Reloaded the invitation and confirmed:
  - `redacted_at` present
  - `auth_token` still populated and not `REDACTED`.
  - `invitation_email`: `REDACTED@example.com`
- Finally confirmed the same tokenized link then redirected to the expired invitation page instead of falling back to the generic flow.

PR #1613, opened 2026-04-15
Allow run-command on PR envs

bencalegari:
## Changes
**Allow run-command on PR envs**
I wanted to [accept this story](https://github.com/DSACMS/iv-cbv-payroll/pull/1606#issuecomment-4254451020) by running commands on a PR environment but wasn't able to. Now you can. 🌟 

- Add `--workspace` flag to `bin/run-command` for selecting Terraform workspace, enables targeting PR environments like `p-1606`)
- Fix help text: correct flag format from underscores to hyphens, fix typo ("teh" → "the")

## Context for reviewers
PR environments use Terraform workspaces (e.g. `p-1606`) separate from the default workspace. Without this flag, `run-command` could only target the default workspace for each environment. Now we can run one-off commands (rails console, migrations, etc.) against PR environments.

## Acceptance testing
  * Try running this against the PR environment: 
    * `bin/run-command --workspace p-1613 app dev '["bin/rails","runner", "puts ENV[\"IMAGE_TAG\"]"]'`
    * You should see the current sha `2c2037b37b87a85742be1d6e5675bfd4a9d8796b` put out in the logs, confirming it works.

## Preview environment
♻️ Environment destroyed ♻️

### DIFF
--- a/bin/run-command
+++ b/bin/run-command
@@ -10,6 +10,8 @@
 #     e.g. '[{ "name" : "DB_USER", "value" : "migrator" }]'
 #   --task_role_arn - the IAM role ARN that the task should assume. Overrides the
 #     task role specified in the task definition.
+#   --workspace - the Terraform workspace to use when resolving service outputs.
+#     This is useful for PR environments, which use workspaces like "p-1606".
 #
 # Positional parameters:
 #   app_name (required) - the name of subdirectory of /infra that holds the
@@ -27,6 +29,7 @@ set -euo pipefail
 # Parse optional parameters
 environment_variables=""
 task_role_arn=""
+workspace=""
 while :; do
   case "$1" in
     --environment-variables)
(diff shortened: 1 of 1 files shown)


bin/run-command
```diff
@@ -64,10 +73,19 @@ echo "  environment=${environment}"
...
 echo "  environment_variables=${environment_variables:-}"
 echo "  task_role_arn=${task_role_arn:-}"
+echo "  workspace=${workspace:-}"
 echo
 
 # Initialize terraform and select workspace if specified
+# If a workspace is specified, reset to default first so that a stale
+# .terraform/environment from a previous run doesn't cause terraform-init to fail.
+if [ -n "${workspace}" ]; then
+  echo "default" > "infra/${app_name}/service/.terraform/environment"
```
daphnegold, line 83: Nit: it'd be nice to add a guard to check `.terraform/environment` exists before writing to it since its creation depends on whether tf has been initialized locally. But if it fails for that reason, not really a big deal and unlikely.

daphnegold (approved):
Gr8!

bin/run-command
bencalegari: Easy change to make, thanks!

PR #1619, opened 2026-04-19
Fix 0 community service hours in CE PDF report

ilyakamens:
## Changes
The community service section of the CE PDF report (`/activities/submit.pdf`) was displaying 0 hours for volunteering activities. The template read `activity.hours` instead of summing the actual monthly hour records from `volunteering_activity_months`. `activity.hours` is never written to by the current volunteering flow AFAICT; hours are tracked per-month in `volunteering_activity_months`.

This is consistent with how the work programs section already computes hours via `job_training_activity_months.sum(:hours)`.

## Context for reviewers

The `volunteering_activities.hours` column appears to be a legacy field (?) that isn't populated by the current monthly hours entry flow. The real data lives in `volunteering_activity_months`. The total hours summary row at the top of the PDF was already correct (it uses the progress calculator), but the per-activity row in the community service table was wrong.

I added a test that creates monthly hour records without setting the top-level `hours` attribute, then asserts the PDF renders the correct sum. This test fails without my changes, and passes with them.

## Acceptance testing

### DIFF
--- a/app/app/views/activities/submit/show.pdf.erb
+++ b/app/app/views/activities/submit/show.pdf.erb
@@ -91,7 +91,7 @@
       <% date_value = activity.date ? l(activity.date) : not_provided %>
       <%= table.with_row do |row| %>
         <% row.with_data_cell.with_content(activity.organization_name.presence || not_provided) %>
-        <% row.with_data_cell.with_content(activity.hours || 0) %>
+        <% row.with_data_cell.with_content(activity.volunteering_activity_months.sum(:hours)) %>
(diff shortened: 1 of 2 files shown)


ilyakamens:
cc @lippytak

serverless-mom:
Great bugfix @ilyakamens ! the team will get this reviewed this week.

tdooner (approved):
I verified the fix works locally. Thank you for the contribution @ilyakamens . If you want more bugs to fix, let us know (cc @serverless-mom) ! We've got a bunch in our backlog.

PR #1649, opened 2026-04-28
FFS-4241: Silence 403, 408, 502 errors from LA

tdooner:
## Changes

**FFS-4241: Silence 403, 408, 502 errors from LA**
> This adds a new configuration option,
> `silence_errors_from_response_codes`, which allows the JSON and HTTP PDF
> transmitters to raise a different class of exception upon receiving
> those response codes. This exception, the `ApplicationJob::SilencedError`
> then is not reported to NewRelic.

## Acceptance testing

### DIFF
--- a/app/app/services/transmitters/json_transmitter.rb
+++ b/app/app/services/transmitters/json_transmitter.rb
@@ -26,8 +26,7 @@ class Transmitters::JsonTransmitter
       @cbv_flow.touch(:json_transmitted_at)
       "ok"
     else
-      Rails.logger.error "Unexpected response from agency: code=#{res.code} message=#{res.message} body=#{res.body}"
-      raise JsonTransmitterError.new("Unexpected response from agency: code=#{res.code} message=#{res.message} body=#{res.body}")
+      raise_response_error!(res, JsonTransmitterError)
     end
   end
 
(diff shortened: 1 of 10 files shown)


app/app/services/transmitters/json_transmitter.rb
```diff
@@ -44,6 +43,20 @@ def payload
...
+    error_class = silence_errors_from_response_codes.include?(response.code.to_s) ?
+      ApplicationJob::SilencedError :
+      default_error_class
+    raise error_class, message
+  end
+
+  def silence_errors_from_response_codes
+    Array(@current_agency.transmission_method_configuration["silently_retry_error_codes"]).map(&:to_s)
+  end
+
```
bencalegari, line 59: Any reason not to move this to `BaseTransmitter` for both the json and http_pdf_transmitter?

bencalegari (approved):

app/app/services/transmitters/json_transmitter.rb
tdooner: Codex initially put it there (I suppose as a piece of evidence that you are in good company with this opinion). But I overrode it, because I thought it made sense to put in the relevant transmitters that do HTTP-related stuff. In other words, the responsibility of handling HTTP error codes doesn't make sense in a base class which does other stuff than HTTP.

That said, I suppose I could DRY it up into a module that gets mixed in. But I thought a bit of duplication here wasn't the end of the world.

wdyt? Should I move it?

PR #1663, opened 2026-04-28
Anchore fix PR no longer auto-removes packages

bencalegari:
## Changes
- Switch `patch-dockerfile.rb` from bidirectional (add+remove) to add-only reconciliation of the Dockerfile auto-fix block
- Log stale already-patched packages instead of removing them
- Update PR body copy in `auto-fix-vulnerabilities.yml` to explain add-only behavior

## Context for reviewers
As you can see in #1647 and #1618, Trivy with `--ignore-unfixed` cannot distinguish "fix shipped in new base image" from "fix already applied by auto-fix block." Auto-removing risked reintroducing CVEs on next build. Cleanup of the auto-fix block has to be a manual process whenever we bump the Ruby base image.

## Acceptance testing

### DIFF
--- a/.github/scripts/patch-dockerfile.rb
+++ b/.github/scripts/patch-dockerfile.rb
@@ -1,16 +1,16 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-# Bidirectional Dockerfile OS package patcher.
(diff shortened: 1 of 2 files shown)


millerti (approved):
I think this makes sense to me. Let's try it!

PR #1671, opened 2026-04-30
FFS-4282: Spike for creating pre-filled volunteer activities 

bencalegari:
## [FFS-4282](https://jiraent.cms.gov/browse/FFS-4282)
- Add `activities` array param to `POST /api/invitations`, that when provided responds with an `activity_tokenized_url`. (`volunteering` is the only acceptable parameter right now)
- Gated behind feature flag and currently only enabled in Sandbox.
- Following the link bring you to the hub prompting you to fill in more information on the hub:
Screenshot 2026-04-30 at 10 19 45 AM

## Context for reviewers
- I had the url get returned in addition to the normal tokenized url since we wanted to add the functionality to the existing invitations API, separating it out could be a little cleaner. 

## Steps to demo (local only)
1. Run migrations
2. Generate API token if you don't already have one handy:
```
bin/rails runner 'u = User.find_or_create_by!(email: "service@sandbox.test", client_agency_id: "sandbox") { |x| x.is_service_account = true }; puts u.api_access_tokens.create!.access_token'
```
3. Mint invitation with pre-populated Volunteering activity
```
TOKEN=<paste-token-from-step-2>
# Happy path — months in window
curl -X POST http://localhost:3000/api/v1/invitations \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "email_address": "demo@example.com",
    "language": "en",
    "agency_partner_metadata": {
      "first_name": "Jane",
      "last_name": "Doe",
      "case_number": "123",
      "date_of_birth": "01/01/1990"
    },
    "activities": [{
      "type": "volunteering",
      "organization_name": "Red Cross",
      "street_address": "123 Main St",
      "city": "Boston",
      "state": "MA",
      "zip_code": "02101",
      "coordinator_name": "Pat Smith",
      "coordinator_email": "pat@redcross.org",
      "coordinator_phone_number": "555-0100",
      "months": [
        {"month": "2026-04-01", "hours": 12}
      ]
    }]
  }'
```
4. Use the `activity_tokenized_url`, get to the hub and you should see a page matching the one seen in the screenshot.


daphnegold (approved):
This is a great starting point! If we’re going to merge this, it’d be good to add some marker/field like `pre_populated: true` so we can differentiate API generated activities from regular draft activities. As it is right now, any unfinished, user-gen draft shows up as an API added activity in the hub. 

Maybe we could filter on `data_source: validated` for now, if that's easier.

bencalegari:
Good call, I'll just scope it to validated for now.

bencalegari:
Some additions based [on Ruvti's feedback:](https://cmsgov.slack.com/archives/C08BBUALWQN/p1777650425476469?thread_ts=1777649939.716949&cid=C08BBUALWQN)
- Accept optional months: [{month, hours}] array per activity in POST /api/v1/invitations, persisted on the invitation.
- Validate each month falls within the reporting window the eventual ActivityFlow would have (anchored to invitation's created_at); return 422 with activities[N].months[M].month field errors if not.

Screenshot 2026-05-01 at 11 40 50 AM

app/spec/controllers/api/invitations_controller_spec.rb
```diff
@@ -155,5 +155,106 @@
...
+      context "when prefilled_activities_enabled is false for the agency" do
+        before do
+          stub_client_agency_config_value("sandbox", :prefilled_activities_enabled, false)
+        end
+
+        it "ignores activities, mints only the cbv invitation, and omits activity_tokenized_url" do
+          expect {
+            post :create, params: params_with_activities
+          }.to change(CbvFlowInvitation, :count).by(1)
+            .and not_change(ActivityFlowInvitation, :count)
```
daphnegold: This is only defined in the [`demo_launcher_controller_spec.rb`](https://github.com/DSACMS/iv-cbv-payroll/blob/cb07a02043f2c2bb74a9501a673bdc03d253c3d2/app/spec/controllers/demo_launcher_controller_spec.rb#L3), so if you try to run this spec file in isolation, the `not_change` helper will cause an error. 

```
 NoMethodError:
       undefined method 'not_change'
```

daphnegold (approved):
Approved again, but suggesting we do something about the spec helper method :D

app/spec/controllers/api/invitations_controller_spec.rb
bencalegari: Nice, I just got rid of it rather than define a new matcher.

PR #1673, opened 2026-04-30
FFS-4009: Basic demo launcher

millerti:
## Changes
1. Split the /demo page into two versions: /launcher, which is a new, simplified version focused on creating excellent demos, and /launcher/advanced, which is the same as the pre-existing /demo page, focused on allowing thorough testing and debugging.
2. Created a more adversarial test suite for the /launcher page so that, if we share with states, we catch misuse of parameters.

## Acceptance testing

## Preview environment
- Link to demo: http://p-1673-app-dev-452793321.us-east-1.elb.amazonaws.com/demo
- Deployed commit: 6729428baf135e4c5e9d6025aba081dcb4b67c64


app/config/routes.rb
```diff
@@ -148,7 +148,10 @@
 
   if Rails.application.config.is_internal_environment
     mount Lookbook::Engine, at: "/lookbook"
-    get "/demo", to: "demo_launcher#show"
-    post "/demo", to: "demo_launcher#create"
+    get  "/launcher",          to: "demo_launcher#launcher"
+    post "/launcher",          to: "demo_launcher#simple_create"
+    get  "/launcher/advanced", to: "demo_launcher#advanced"
+    post "/launcher/advanced", to: "demo_launcher#create"
+    get  "/demo",              to: redirect("/launcher")
```
bencalegari, line 155: Any reason not to just keep it `/demo` and `/demo/advanced` rather than having the redirect?

bencalegari (approved):
Neat. Only suggestion I have is that I like the new style for the link copy/create buttons and it'd be cool to have it on the advanced launcher, even just for style consistency.

Image

app/config/routes.rb
millerti: I'm trying to move away from using "demo" terminology at all. Just for our own sanity.

bencalegari: Totally fair its way too overloaded

millerti:
> Neat. Only suggestion I have is that I like the new style for the link copy/create buttons and it'd be cool to have it on the advanced launcher, even just for style consistency.
> 
> Image

I agree with this suggestion but will pass to keep this PR focused on the basic mode. Let's do this in the future, though!

millerti:
Ming approved this for Acceptance Testing in Slack.

bencalegari (approved):

PR #1697, opened 2026-05-11
Fix warnings in Nokogiri / Templates

bencalegari:
## Changes
- Pin `css_parser` to `~> 1.17` to silence premailer positional-arg deprecation warnings until premailer ships a fix
- Rename component templates from `*.erb` to `*.html.erb` to remove Rails template-handler warnings

## Context for reviewers
We we're getting some noisy warnings when running tests that I wanted to fix:
```
WARNING: Template format for /Users/bencalegari/workspace/iv-cbv-payroll/app/app/components/contact_info_review_table_component.erb is missing, defaulting to :html.
```
as well as:
```
/Users/bencalegari/.rbenv/versions/3.4.9/lib/ruby/gems/3.4.0/gems/premailer-1.23.0/lib/premailer/adapter/nokogiri.rb:65: warning: [DEPRECATION] positional arguments are deprecated use keyword instead.
```

This fixes it by pinning the gem for now and adding the correct suffix to the the templates. Run `/bin/rspec` to test it out. 

## Acceptance testing


millerti (commented):
I'm still seeing the css_parser deprecation warning when I run rspec locally (and in the GHA). I wonder if that issue is really fixed. Could it be because 1.22 still has the problem? Looks like that version is still in the gemfile.

bencalegari:
@millerti  My bad, ~> 1.17 allowed bundler to pick 1.22.0 but since warn was added in css_parser 1.19.0 I set it to ~> 1.17.0 which pins it 1.17.1, to the last release before that one. Specs run clean now.

millerti (approved):

PR #1698, opened 2026-05-11
FFS-4008: Spike: Investigate speeding up our E2E test suite

bencalegari:
## [FFS-4008](https://jiraent.cms.gov/browse/FFS-4008)

I tried to speed up the test sweet but honestly, it's not a LOT faster. 

## Changes
- **Parallelize RSpec on PR CI**: run `parallel_test spec/` with 4 processes in the rspec workflow (new gem `parallel_tests ~> 5.5`). Test `database.yml` is suffixed with `TEST_ENV_NUMBER` so each worker has its own DB; CI seeds 4 schemas via `parallel:create[4]` + `parallel:load_schema[4]`.
- **Split e2e into its own PR workflow**: new `.github/workflows/e2e.yml` runs `spec/e2e` with full axe coverage on every PR parallel to the rspec workflow rather than serialized inside it. `spec/e2e/` is excluded from the rspec job.
- **Axe mode toggle**: `verify_page` now honors `AXE_MODE`: `sample` (default for local; one axe check per unique path per process), `full` (CI), `off` (fast local iteration).
- **Convert three slow `:js` e2e specs to controller specs** that assert the same page titles via `render_views`: `applicant_information`, `expired_invitation`, `synchronization_failures`.
- **Add `test-prof` gem** (dev/test only); document `FACTORY_PROF=flamegraph` and `EVENT_PROF=sql.active_record` usage in `AGENTS.md`.
- **Gitignore `.rspec-local`** If you want to override it locally (e.g. `--fail-fast=3`); documented in `AGENTS.md`.
- **`FeedbacksController` spec**: drop hard-coded `id: 123`; explicitly force an id collision between `CbvFlow` and `ActivityFlow` so the assertion that the controller resolves by `session[:flow_type]` (not by id alone) survives parallel runs.

## Context for reviewers
Benchmark against the most recent app-touching merge (#1710):

| Metric | Baseline #1710 | This PR (`-n 2`) | Δ |
|---|---|---|---|
| Rspec test execution | 1m 54s | **1m 8s** | **−46s (−40%)** |
| E2E test execution (full axe) | 2m 10s | 2m 4s | −6s |
| Rspec job total | 3m 13s | 2m 37s | −36s |
| E2E job total | 3m 39s | 3m 23s | −16s |
| **Wall clock (parallel max)** | **3m 39s** | **3m 23s** | **−16s (≈7%)** |

Notes:
- We also tried `-n 4`: same test-execution time (~1m 7s) but worse wall clock (4m 17s) because 4 processes on a 2-vCPU runner over-saturated CPU and doubled DB-setup time.
- First run on a changed Gemfile.lock pays a one-time `bundle install` cache miss (~80s extra in `setup-project`). The benchmark above is on a warm cache. `ruby/setup-ruby` only does exact-match cache restore, so dependabot bumps on main will continue to invalidate per-PR caches.
- `AXE_MODE=full` on PR e2e keeps a11y coverage at baseline parity. Locally, devs can set `AXE_MODE=sample` for one-check-per-path coverage or `AXE_MODE=off` to skip axe entirely while iterating.

Two main wins:
1. **Parallel rspec inside the job**: 2 workers, each with its own test DB.
2. **E2E + axe split into a sibling workflow** instead of running serially after rspec — both fire on PR open/update and run concurrently, so total PR wall-time is `max(rspec, e2e)` instead of `rspec + e2e`.

## Acceptance testing
  * CI / test-tooling only; no user-facing change.


.github/workflows/e2e-tests.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           DATABASE_URL: ${{ steps.setup.outputs.database_url }}
           RAILS_MASTER_KEY: 0f5db7adba61669d4cc45dfa6b2464d7
           E2E_RUN_TESTS: 1
+          AXE_MODE: full
```
tdooner: Nit: I was trying to keep E2E-related env variables starting with `E2E_`

.github/workflows/rspec.yml
```diff
@@ -30,10 +33,27 @@ jobs:
...
+          DB_HOST: localhost
+          DB_USERNAME: cidbuser
+          DB_PASSWORD: postgres
           RAILS_MASTER_KEY: 0f5db7adba61669d4cc45dfa6b2464d7
-        run: bundle exec rspec
+        run: |
+          bundle exec parallel_test spec/ \
+            --type rspec \
+            -n 2 \
+            --exclude-pattern '^spec/e2e/'
```
tdooner, line 59: Yeah I guess I can see why this didn't have a significant impact - I'd expect we are CPU-bound on the CI containers, and parallelizing will probably not help too much if that's the case? But still, interesting idea to try. I wish Github had a better way to do this other than by using [a complicated Github Marketplace Action](https://github.com/marketplace/actions/split-tests) for it.

app/config/database.yml
```diff
@@ -95,7 +95,7 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  database: iv_cbv_payroll_test
+  database: iv_cbv_payroll_test<%= ENV["TEST_ENV_NUMBER"] %>
```
tdooner, line 98: I also modified this, in https://github.com/DSACMS/iv-cbv-payroll/pull/1719, to allow my agents to make different databases per branch. But I can rebase that atop this.

tdooner (approved):

PR #1709, opened 2026-05-13
Fix PR environment DB access

bencalegari:
## Changes
- **DB access on PR envs**: add `db:grant_app_access` rake task and wire it into `db-migrate` to grant runtime app-user privileges on per-PR schemas (the `role_manager` Lambda only covers env-level schemas).
- **Background jobs on PR envs**: run Solid Queue inside Puma when `SOLID_QUEUE_IN_PUMA=true` (set automatically on temporary envs in Terraform). PR envs don't get a dedicated `solid_queue` ECS service, so jobs queued by web requests (e.g. `NscSynchronizationJob` from the Education flow) had nothing draining them.
- **Webhook delivery to PR envs**:
  - New `WebhookRegistrar` extracted from the existing ngrok-development setup; new `pr_environment_webhooks.rb` initializer registers Pinwheel/Argyle subscriptions under the name `pr-<pr_number>` on PR-env web boot.
  - New `pr_webhooks:destroy` rake task and wired into `bin/destroy-pr-environment` to clean up subscriptions on PR close.
- **HTTPS reachability for PR envs**: introduce a `pr_subdomain` module variable (`p-<pr_number>.navapbc.cloud`) covered by the existing `*.navapbc.cloud` wildcard cert. Route53 alias to the PR ALB; `DOMAIN_NAME` env var and `service_endpoint` output now resolve to the HTTPS subdomain. Without this, Pinwheel/Argyle webhook delivery fails TLS validation against the ELB DNS hostname.
- **Deploy gate**: `bin/update-pr-environment` polls `${service_endpoint}/health` until 200 (5 min deadline) before declaring the env ready, so DNS propagation and HTTPS readiness are confirmed before the PR description is updated.

## Context for reviewers
Merging #1440 split each PR environment into its own database schema (`p_<pr_number>`). Three subsystems were silently relying on PR envs sharing the `app` schema with dev/demo:

| Subsystem | How it worked before #1440 | How it broke |
|---|---|---|
| DB permissions | `role_manager` Lambda granted on env-level schemas; PR envs used the shared `app` schema and inherited grants | New per-PR schemas had no grants; app user couldn't query |
| Background jobs | PR enqueued into `app.solid_queue_jobs`; dev/demo's worker drained it | PR now writes `p_N.solid_queue_jobs`; no worker polls that schema |
| Webhook events | Pinwheel sent webhooks to dev's URL; demo's webhook controller updated `app.cbv_flows` (where PR's rows lived) | PR's `cbv_flows` row lives in `p_N`; demo's handler queries `app.*`, finds nothing |

Each subsystem is fixed inside the PR env now instead of relying on demo. The HTTPS subdomain change is required for Pinwheel/Argyle to accept the webhook delivery target.

## Acceptance testing
  * Verified end-to-end against this PR's own environment: open https://p-1709.navapbc.cloud/demo, launch with the **Rick Banas** tokenized link, add an **Education** activity (NSC sync flow), then a **Pinwheel** connection in the income flow. Both should advance past the "We're gathering details…" loading page within seconds instead of hanging.

## Infrastructure Changes
- [x] Plan reviewed

Plan reviewed, no apply needed — changes are no-ops for dev (gated on is_temporary)

### DIFF
--- a/bin/destroy-pr-environment
+++ b/bin/destroy-pr-environment
@@ -27,6 +27,16 @@ if ! terraform -chdir="infra/${app_name}/service" workspace select "${workspace}
   exit 0
 fi
 
+echo "::group::Remove webhook subscriptions"
+./bin/run-command \
+  --workspace "${workspace}" \
+  --environment-variables "[{\"name\":\"PR_NUMBER\",\"value\":\"${pr_number}\"}]" \
+  "${app_name}" "${environment}" '["bin/rails", "pr_webhooks:destroy"]'
+echo "::endgroup::"
+
+# Re-select workspace after run-command reinitializes terraform
+terraform -chdir="infra/${app_name}/service" workspace select "${workspace}"
+
 echo "::group::Drop PR database schema"
 terraform -chdir="infra/${app_name}/app-config" init > /dev/null
 terraform -chdir="infra/${app_name}/app-config" apply -auto-approve > /dev/null
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -178,8 +178,9 @@ RUN apt-get update -qq && \
         && \
     rm -rf /var/lib/apt/lists /var/cache/apt/archives
 
-# Install custom db migrate script
+# Install custom database scripts
 COPY bin/db-migrate /usr/bin/
+COPY bin/db-drop-schema /usr/bin/
 
 # Copy built artifacts: gems, application, binary dependencies
 COPY --from=release-build /usr/local/bundle /usr/local/bundle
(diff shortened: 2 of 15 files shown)


daphnegold (changes requested):
Looks very promising as a fix for the issues we are seeing with migrations! Since the drop schema script is run as a container command, I do think we need to update the Dockerfile to copy it into the image.

daphnegold (approved):
Great work

bin/destroy-pr-environment
```diff
@@ -27,6 +27,16 @@ if ! terraform -chdir="infra/${app_name}/service" workspace select "${workspace}
   exit 0
 fi
 
+echo "::group::Remove webhook subscriptions"
```
daphnegold, line 30: We may have to do more cleanup of orphaned PR environments in the future if something out of our control goes wrong and this blocks destroy, but that's fine and I think this is the right way to do it

PR #1710, opened 2026-05-13
FFS-4319: Pre-populate employment data

millerti:
## Changes
* Enabled pre-populated data for the Employment activity type.
* Added an easy way to test pre-populated data on the /launcher/advanced page.

## Acceptance testing

## Preview environment
- Link to launcher: https://p-1710.navapbc.cloud/launcher/advanced
- Deployed commit: 8ac36893954b8cc8353120cf0d010da187a4c2fc

### DIFF
--- a/app/app/controllers/demo_launcher_controller.rb
+++ b/app/app/controllers/demo_launcher_controller.rb
@@ -128,6 +128,45 @@ class DemoLauncherController < ApplicationController
     end
   end
 
+  def build_pre_populated_activities
+    activities = []
+    month_strings = reporting_window_months_for_activities(launcher_params[:client_agency_id])
+
+    if launcher_params[:volunteering_enabled] == "1"
+      hours = launcher_params[:volunteering_hours_per_month].to_i
+      activities << {
+        "type" => "volunteering",
+        "organization_name" => launcher_params[:volunteering_organization_name].presence || "Red Cross",
+        "months" => month_strings.map { |m| { "month" => m, "hours" => hours } }
+      }
+    end
+
+    if launcher_params[:employment_enabled] == "1"
+      hours = launcher_params[:employment_hours_per_month].to_i
+      gross_income = launcher_params[:employment_gross_income_per_month].to_i
+      activities << {
+        "type" => "employment",
+        "employer_name" => launcher_params[:employment_employer_name].presence || "Acme Corp",
+        "months" => month_strings.map { |m| { "month" => m, "hours" => hours, "gross_income" => gross_income } }
+      }
+    end
+
+    activities
+  end
+
+  def reporting_window_months_for_activities(client_agency_id)
+    # Uses the agency's default application window regardless of the launcher's reporting_window
+    # setting, because ActivityFlowInvitation validates months against the same default range.
+    range = ActivityFlow.expected_reporting_window_range(client_agency_id)
+    months = []
+    current = range.begin.beginning_of_month
+    while current <= range.end
+      months << current.strftime("%Y-%m-%d")
+      current = current.next_month
+    end
+    months
+  end
+
   def launch_overrides(flow_type)
     overrides = if flow_type == "cbv"
                   launcher_params.slice(:demo_timeout).select { |_, v| v.present? }
@@ -154,7 +193,14 @@ class DemoLauncherController < ApplicationController
       :renewal_required_months,
       :reporting_window_start,
       :demo_timeout,
-      :launch_type
+      :launch_type,
+      :volunteering_enabled,
+      :volunteering_organization_name,
+      :volunteering_hours_per_month,
+      :employment_enabled,
+      :employment_employer_name,
+      :employment_hours_per_month,
+      :employment_gross_income_per_month
     )
   end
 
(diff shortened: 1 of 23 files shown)


app/app/controllers/demo_launcher_controller.rb
```diff
@@ -207,9 +253,11 @@
   end
 
   def build_tokenized_url(client_agency_id, overrides)
+    pre_populated = build_pre_populated_activities
     invitation = ActivityFlowInvitation.create!(
       client_agency_id: client_agency_id,
-      reference_id: "demo-#{SecureRandom.hex(4)}"
+      reference_id: "demo-#{SecureRandom.hex(4)}",
+      pre_populated_activities: pre_populated
```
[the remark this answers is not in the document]
millerti: what

app/app/controllers/demo_launcher_controller.rb
```diff
@@ -128,6 +128,45 @@ def normalize_date_param(date_str)
...
+      gross_income = launcher_params[:employment_gross_income_per_month].to_i
+      activities << {
+        "type" => "employment",
+        "employer_name" => launcher_params[:employment_employer_name].presence || "Acme Corp",
+        "months" => month_strings.map { |m| { "month" => m, "hours" => hours, "gross_income" => gross_income } }
+      }
+    end
+
+    activities
+  end
```
tdooner, line 155: Neat!

app/app/models/activity_flow.rb
```diff
@@ -52,6 +53,20 @@ def self.hydrate_pre_populated_activities!(flow, invitation)
           month_attrs = month_entry.respond_to?(:stringify_keys) ? month_entry.stringify_keys : month_entry.to_h.stringify_keys
           activity.volunteering_activity_months.create(month_attrs.slice(*VolunteeringActivityMonth::FIELDS))
         end
+      when "employment"
+        next if flow.employment_activities.exists?
+
+        activity = flow.employment_activities.create(
+          attrs.slice(*EmploymentActivity::FIELDS)
+            .merge("draft" => true, "data_source" => "validated")
```
tdooner: Hmm, I'm thinking we might want a different `data_source` here at some point to indicate that the data was state-provided, rather than from a validated Emmy data source. But that's not really related to your commit.

app/app/models/activity_flow.rb
```diff
@@ -52,6 +53,20 @@ def self.hydrate_pre_populated_activities!(flow, invitation)
...
+        )
+        next unless activity.persisted?
+
+        Array(attrs["months"]).each do |month_entry|
+          next unless month_entry.is_a?(Hash) || month_entry.respond_to?(:to_h)
+          month_attrs = month_entry.respond_to?(:stringify_keys) ? month_entry.stringify_keys : month_entry.to_h.stringify_keys
```
tdooner: I think this might be overly defensive? It seems like `months` will always either be a hash or `nil`? But that's fine I suppose. (And chalk it up as a question that would be a lot easier to answer if we had a proper type system....)

app/app/models/activity_flow_invitation.rb
```diff
@@ -39,9 +39,10 @@ def pre_populated_activities_shape
         next
       end
 
-      organization_name = entry["organization_name"] || entry[:organization_name]
-      if organization_name.blank?
-        errors.add("pre_populated_activities[#{idx}].organization_name", "can't be blank")
+      name_field = { "volunteering" => "organization_name", "employment" => "employer_name" }[type.to_s]
```
tdooner: Would be nice to extract this as a constant named something like `PRE_POULATED_REQUIRED_FIELDS` or `PRE_POPULATED_NAME_FIELDS`

app/spec/controllers/demo_launcher_controller_spec.rb
```diff
@@ -550,6 +561,79 @@
       end
     end
 
+    context "with pre-populated activities" do
+      it "creates an invitation with a volunteering activity when volunteering is enabled" do
+        Timecop.freeze(Date.new(2026, 5, 13)) do
```
tdooner, line 566: I think it would be worth trying to remove these Timecop.freeze's - I'm not sure they are necessary.

app/spec/models/employment_activity_spec.rb
```diff
@@ -1,6 +1,23 @@
...
+        street_address_line_2
+        city
+        state
+        zip_code
+        contact_name
+        contact_email
+        contact_phone_number
+      ])
+    end
+  end
```
tdooner: This test is pointless - let's remove it. It's just going to fail when this constant is modified, and doesn't indicate anything of consequence when it fails.

tdooner (approved):

app/app/models/activity_flow.rb
millerti: Great idea.

app/spec/controllers/demo_launcher_controller_spec.rb
millerti: I do think this particular one is necessary because of the months, but I removed two others.

millerti:
Rutvi, Paul, and Allyse approved in Slack.

PR #1723, opened 2026-05-18
Dockerfile autofix PRs include Grype/Anchore fixes

bencalegari:
## Changes
- Extend Dockerfile autofix script to include the union of Grype (Anchore) and Trivy results.
- Skip reviewer lottery when PR already has a requested reviewer, so autofix PR reviewers survive the close/reopen CI dance.
- Change formatting of table to use html so we don't see a blank header like in #1722. 
- Added tests! They run in our precommit hook. when patch-dockerfile.rb changes 

## Context for reviewers
- Trivy and Grype use different vulnerability DBs and disagree often; union avoids missing fixes either scanner catches alone. You can see this in #1722 where all but the `anchore-scan` passed successfully.

## Acceptance testing

### DIFF
--- a/.github/scripts/patch-dockerfile.rb
+++ b/.github/scripts/patch-dockerfile.rb
@@ -20,24 +20,25 @@
 # Only manages blocks marked with the AUTOFIX_MARKER comment.
 # Never touches the manually-maintained "Upgrade packages in base image" block.
 
-require "json"
-require "set"
 
-AUTOFIX_MARKER    = "# Auto-fix: OS package vulnerabilities detected by Trivy and Grype"
-INSERTION_ANCHOR  = "# Rails app lives here"
-OS_PACKAGE_TYPES  = Set["deb", "apk", "rpm"].freeze
+
+require 'json'
+
+AUTOFIX_MARKER    = '# Auto-fix: OS package vulnerabilities detected by Trivy and Grype'
+INSERTION_ANCHOR  = '# Rails app lives here'
+OS_PACKAGE_TYPES  = Set['deb', 'apk', 'rpm'].freeze
 
 def parse_trivy(trivy_path)
   data = JSON.parse(File.read(trivy_path))
   needed = Set.new
 
-  (data["Results"] || []).each do |result|
-    next unless result["Class"] == "os-pkgs"
+  (data['Results'] || []).each do |result|
+    next unless result['Class'] == 'os-pkgs'
 
-    (result["Vulnerabilities"] || []).each do |vuln|
-      next if vuln["FixedVersion"].to_s.empty?
+    (result['Vulnerabilities'] || []).each do |vuln|
+      next if vuln['FixedVersion'].to_s.empty?
 
-      needed.add(vuln["PkgName"])
+      needed.add(vuln['PkgName'])
     end
   end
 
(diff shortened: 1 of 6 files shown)


.github/scripts/patch-dockerfile.rb
```diff
@@ -3,25 +3,27 @@
...
+# on the next build. Clear the block manually when bumping the base Ruby image.
 #
 # Only manages blocks marked with the AUTOFIX_MARKER comment.
 # Never touches the manually-maintained "Upgrade packages in base image" block.
 
 require "json"
 require "set"
 
-AUTOFIX_MARKER    = "# Auto-fix: OS package vulnerabilities detected by Trivy"
+AUTOFIX_MARKER    = "# Auto-fix: OS package vulnerabilities detected by Trivy and Grype"
```
daphnegold: I see you updated the auto-fix marker, but I don't see a corresponding update to the Dockerfile

.github/workflows/auto-fix-vulnerabilities.yml
```diff
@@ -98,17 +108,17 @@ jobs:
...
+                echo "<tr><td><b>Source PR</b></td><td>#${TRIGGER_PR_NUMBER}</td></tr>"
               fi
+              echo "</table>"
             else
               echo "**Triggered by:** manual \`${EVENT_NAME}\`"
             fi
             echo ""
-            echo "Trivy identified OS packages with available CVE fixes. The auto-managed"
+            echo "Trivy and Grype identified OS packages with available CVE fixes. The auto-managed"
             echo "\`RUN apt-get install\` block in the Dockerfile BASE stage has been updated:"
```
daphnegold, line 122: Is `apt-get` appropriate for all fixes found by Trivy & Grype, specifically?

daphnegold (commented):
Some questions before 🛳️, thank you!

.github/scripts/patch-dockerfile.rb
bencalegari: The Dockerfile will get updated with whatever fix is needed after it gets merged triggers an update. I didn't have it run on this PR just to keep the fix separated.

.github/workflows/auto-fix-vulnerabilities.yml
bencalegari: Not necessarily, only for the purposes of this autofix PR maker. Ruby/JS/GithubAction vulnerabilities should continue to be handled by dependabot. I'm glad you asked since Grype is not scoped to OS vulnerabilities by default, making that fix now.

.github/scripts/patch-dockerfile.rb
daphnegold: Okay, but doesn't the marker need to be changed in the Dockerfile so it updates the existing block, or is it expected that a new block is created?

Referring to [here](https://github.com/DSACMS/iv-cbv-payroll/blob/67aef9dc9929c8372123a13e6ac39b9e6f92f4e7/app/Dockerfile#L43), just to be clear

bencalegari: Ah I misunderstood you are totally right. Tough that this isn't easy to test until after merge, I'm going to try to write some tests to make sure that isn't missed in the future.

daphnegold (approved):
wooohoo

PR #1726, opened 2026-05-19
FFS-4106: Update table and column header content to match designs

daphnegold:
## Changes
- Updated activity table headings to match the "approved" designs
- Added `activity_flow_labels` so shared components can use activity wording without changing Emmy Income labels
- "Edit" column header is now `usa-sr-only` to match designs but preserve screen reader functionality
- Changed education section heading from "Monthly details" to "Term details"
- All activity month rows now show month + year, as in January 2026 (per discussion with @paulgehrig)
- Suppressed old payroll footnotes when activity flow labels are shown (tbd new content updates)
- Fixed an existing work program phone `data-label` issue

## Acceptance testing

## Preview environment
- Link to launcher: https://p-1726.navapbc.cloud/launcher/advanced
- Deployed commit: 6874d61f9209b78e6ceafb6ff9c928d779b792fd

### DIFF
--- a/app/app/components/report/w2_monthly_summary_table_component.rb
+++ b/app/app/components/report/w2_monthly_summary_table_component.rb
@@ -4,11 +4,20 @@ class Report::W2MonthlySummaryTableComponent < ViewComponent::Base
 
   attr_reader :employer_name
 
-  def initialize(report, payroll_account, is_caseworker: false, show_footnote: true, show_header: true, is_pdf: false)
+  def initialize(
+    report,
+    payroll_account,
+    is_caseworker: false,
+    show_footnote: true,
+    show_header: true,
+    is_pdf: false,
+    activity_flow_labels: false
+  )
     @report = report
     @show_footnote = show_footnote
     @show_header = show_header
     @is_pdf = is_pdf
+    @activity_flow_labels = activity_flow_labels
 
     # Note: payroll_account may either be the ID or the payroll_account object
     @account_id = payroll_account.class == String ? payroll_account : payroll_account.aggregator_account_id
@@ -43,8 +52,16 @@ class Report::W2MonthlySummaryTableComponent < ViewComponent::Base
     4
   end
 
+  def hours_header
+    if @activity_flow_labels
+      I18n.t("components.report.monthly_summary_table.activity.community_engagement_hours")
+    else
+      I18n.t("components.report.monthly_summary_table.w2.title_total_hours_worked")
+    end
+  end
+
   def show_footnote?
-    @show_footnote
+    @show_footnote && !@activity_flow_labels
   end
 
   def format_accrued_gross_earnings(month_summary)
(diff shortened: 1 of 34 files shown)


app/app/components/report/w2_monthly_summary_table_component.rb
```diff
@@ -58,10 +75,16 @@ def format_paychecks_count(month_summary)
...
 
   def format_no_payments_found
     I18n.t("cbv.payment_details.show.none_found", report_data_range: @report_data_range)
   end
+
+  def format_activity_hours(hours)
+    return hours unless @activity_flow_labels
+
+    hours.to_f == hours.to_i ? hours.to_i : hours
+  end
```
millerti, line 89: Is it worth trying to share this with the gig version somehow?

app/spec/components/report/gig_monthly_summary_table_component_spec.rb
```diff
@@ -174,6 +174,20 @@
         expect(subject.css("tbody tr:nth-child(3) td:nth-child(4)").to_html).to include "4.7"
       end
 
+      it "renders activity flow labels when requested" do
```
millerti, line 177: Is it worth testing the opposite (doesn't render when not requested)?

millerti:
Just thinking out loud here: I wonder if it would make our lives easier to make some kind of new test page that shows how all the tables render. I'm imagining something like going to /table-test after playing around in Emmy and being able to see all versions of all the tables side by side instead of having to jump between the individual summary pages and the Review and Submit page. And maybe we could toggle things too, like the various booleans (or maybe that's overkill).

What do you think? Good idea or bad idea? Not saying you need to do this as part of this ticket, just wondering.

millerti (approved):

daphnegold:
> Just thinking out loud here: I wonder if it would make our lives easier to make some kind of new test page that shows how all the tables render. I'm imagining something like going to /table-test after playing around in Emmy and being able to see all versions of all the tables side by side instead of having to jump between the individual summary pages and the Review and Submit page. And maybe we could toggle things too, like the various booleans (or maybe that's overkill).
> 
> What do you think? Good idea or bad idea? Not saying you need to do this as part of this ticket, just wondering.

This is a great idea, and I can create a ticket to track it. I'd suggest we use lookbook previews over a `/table-test` endpoint.

app/app/components/report/w2_monthly_summary_table_component.rb
daphnegold: Hmm, I'm not sure it's worth it quite yet. I'm also checking in with @paulgehrig to see if we even like this.

UPDATE: Paul says "perfecto" :D

app/spec/components/report/gig_monthly_summary_table_component_spec.rb
daphnegold: The opposite path is already covered by the other default specs where it isn't requested. They assert that the default headers and footnote copy are shown.

PR #1748, opened 2026-05-26
[FFS-4393] Add support for publishing Docker images to Artifactory

iannorriswork:
This change introduces the ability to publish built Docker images to an external Artifactory repository in addition to the existing AWS ECR workflow. This is required because the application is now deploying to two different AWS setups which utilize different image repositories.

Note that if you can convince Jake out of artifactory into an AWS ECR setup this might be even simpler, but for now assuming that is important.

Changes:
- Created `bin/publish-release-artifactory`: A new script to handle tagging and pushing images specifically for Artifactory.
- Updated `Makefile`: Added a `release-publish-artifactory` target to standardize the Artifactory publishing process.
- Modified `.github/workflows/build-and-publish-to-artifactory.yml`:
  - Added logic to retrieve Artifactory credentials from AWS Secrets Manager.
  - Implemented `docker login` for the Artifactory endpoint.
  - Integrated the new `release-publish-artifactory` Makefile target.
  - Included a fallback mechanism to use standard ECR publishing if `EMMY_IMAGE_REPO` is not defined.

## Acceptance testing

- [ X ] No acceptance testing needed
  * this is solely affecting github, and also warning i haven't figured out how to test this github action cleanly so this may require a followup PR if the general approach looks fine.

### DIFF
--- a/.github/workflows/build-and-publish-to-artifactory.yml
+++ b/.github/workflows/build-and-publish-to-artifactory.yml
@@ -27,6 +27,8 @@ jobs:
   get-commit-hash:
     name: Get commit hash
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     outputs:
       commit_hash: ${{ steps.get-commit-hash.outputs.commit_hash }}
     steps:
--- a/.github/workflows/build-and-publish-to-artifactory.yml
+++ b/.github/workflows/build-and-publish-to-artifactory.yml
@@ -1,0 +1,86 @@
+name: Build and publish
+run-name: Build and publish to CMS ${{ inputs.app_name }}:${{ inputs.ref }}
+
+on:
+  workflow_call:
+    inputs:
+      app_name:
+        description: "name of application folder under infra directory"
+        required: true
+        type: string
+      ref:
+        description: The branch, tag or SHA to checkout. When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, use branch or tag that triggered the workflow run.
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      app_name:
+        description: "name of application folder under infra directory"
+        required: true
+        type: string
+      ref:
+        description: The branch, tag or SHA to checkout. When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, use branch or tag that triggered the workflow run.
+        required: true
+        type: string
+
+jobs:
+  get-commit-hash:
+    name: Get commit hash
+    runs-on: ubuntu-latest
+    outputs:
+      commit_hash: ${{ steps.get-commit-hash.outputs.commit_hash }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.ref }}
+      - name: Get commit hash
+        id: get-commit-hash
+        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-publish-to-artifactory.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Set up Terraform
+        uses: ./.github/actions/setup-terraform
+
+      - name: Configure AWS credentials
+        uses: ./.github/actions/configure-aws-credentials
+        with:
+          app_name: ${{ inputs.app_name }}
+          environment: shared
+
+      - name: Login to Artifactory
```
j-shilling, line 71: IDK if it's worth setting up some kind of x-repo action sharing, but this is what we have over in the api:

https://github.com/CMSgov/emmy-api/blob/ca37f0bf9c80e503b310075c79d3564d63081050/.github/workflows/publish-image.yml#L68-L114

The only real difference is that it grabs the credentials as github outputs then passes them to a 3rd party docker login action

.github/workflows/build-and-publish-to-artifactory.yml
```diff
@@ -0,0 +1,88 @@
...
+    runs-on: ubuntu-latest
+    needs: get-commit-hash
+    concurrency: ${{ github.workflow }}-${{ needs.get-commit-hash.outputs.commit_hash }}
+
+    permissions:
+      contents: read
+      id-token: write
+
+    env:
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
```
j-shilling, line 55: We probably don't want the dockerhub_token in here just in case we accidently login to dockerhub

bin/publish-release-artifactory
```diff
@@ -0,0 +1,32 @@
...
+  exit 1
+fi
+
+artifactory_image="${image_repo}/${image_name}:${image_tag}"
+
+echo "Tagging image: ${image_name}:${image_tag} -> ${artifactory_image}"
+docker tag "${image_name}:${image_tag}" "${artifactory_image}"
+
+echo "Pushing image to Artifactory: ${artifactory_image}"
+docker push "${artifactory_image}"
```
j-shilling, line 30: buildx drivers let us build and push in the same go. I don't know the context behind why we don't do that historically, but I think we are using buildx anyway

j-shilling (approved):
lgtm

bin/publish-release-artifactory
```diff
@@ -0,0 +1,32 @@
...
+echo "  image_name=${image_name}"
+echo "  image_tag=${image_tag}"
+echo "  image_repo=${image_repo}"
+
+if [[ -z "${image_repo}" ]]; then
+  echo "Error: image_repo is required for Artifactory publish"
+  exit 1
+fi
+
+artifactory_image="${image_repo}/${image_name}:${image_tag}"
```
daphnegold, line 24: Is this where the new AWS infra expects to pull from? Just checking

daphnegold (approved):
Piggybacking on Jake's comments with one question. 🚢

bin/publish-release-artifactory
iannorriswork: as it currently stands, yup! 

for context the deploy script takes in a artifactory image and uses that for the new deploy

iannorriswork:
merging with the understanding i'm fixing the issues in a followup PR.

PR #1770, opened 2026-06-01
Fix time related spec failures

bencalegari:
Our entry into the summer months exposed some weird edge cases with our tests: May is the first summer month and the since the reporting window moved from March - April to May - June we incorrectly applied the spring term carryover. Now the reporting-window math is deterministic and will avoid this in the future.

## Changes
- Added `include ActiveSupport::Testing::TimeHelpers` to `activities_helper_spec`, `education_activity_spec`, and `activity_flow_progress_calculator_spec`.
- Wrapped the `#education_cards`, `#progress_hours_for_month`, and education progress example groups in `travel_to(Date.new(2025, 11, 15))` so they run against a fixed date.

## Acceptance testing

### DIFF
--- a/app/spec/helpers/activities_helper_spec.rb
+++ b/app/spec/helpers/activities_helper_spec.rb
@@ -1,6 +1,8 @@
 require "rails_helper"
 
 RSpec.describe ActivitiesHelper do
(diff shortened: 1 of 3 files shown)


iannorriswork (approved):
nice

PR #1772, opened 2026-06-02
FFS-4318: Extend skeleton implementation to Work Program activities via API

daphnegold:
## Changes
* Last skeleton activity, Work Program!
* Added pre-populated Work Program draft cards to the Activity Hub with the Complete CTA.
* Moved shared `pre_populated_drafts` and `pre_populated_draft?` behavior into `Activity` model in a little refactor
* Updated advanced launcher with pre-populated Work Program option

Note: It kinda bugs me that we still mix "Work Program" and "Job Training" naming. I'd like to clean that up sometime. I'll make a ticket. 

## Acceptance testing

Screenshot 2026-06-02 at 3 34 30 PM
Screenshot 2026-06-02 at 3 34 40 PM

## Preview environment
- Link to launcher: https://p-1772.navapbc.cloud/launcher/advanced
- Deployed commit: be5afb140a12eb5e3172124e6e58b2ca05929044


app/app/models/activity_flow.rb
```diff
@@ -74,6 +74,18 @@ def self.hydrate_pre_populated_activities!(flow, invitation)
...
+
+        activity = flow.job_training_activities.create(
+          attrs.slice(*JobTrainingActivity::FIELDS)
+            .merge("draft" => true, "pre_populated" => true)
+        )
+        next unless activity.persisted?
+
+        Array(attrs["months"]).each do |month_entry|
+          activity.job_training_activity_months.create(month_entry.stringify_keys.slice(*JobTrainingActivityMonth::FIELDS))
+        end
```
bencalegari, line 88: What do you think about deduplicating the logic here?

app/db/schema.rb
```diff
@@ -257,6 +257,7 @@
     t.boolean "draft", default: false, null: false
     t.string "organization_address"
     t.string "organization_name"
+    t.boolean "pre_populated", default: false, null: false
```
bencalegari, line 260: Just thinking out loud here.. Activities can be a draft, pre populated, and have either a validated/self attested data source. Do we need all those?

For example: isn't every pre-populated activity self attested? Could we just have an `agency_provided` data source? This would be a large refactor probably, definitely outside the scope but curious what you think.

bencalegari (approved):

app/db/schema.rb
daphnegold: I do think we need all these, they are tracking different things and different concerns.

`pre-populated` tracks origin
`draft` tracks completion state
`data_source` is attestation and verification

And a `pre-populated` activity can still go through the normal self-attested path. Education is a special case in particular that uses `fully_self_attested`. Conceptually, `pre-populated` does not supersede `data_source` which is core to how the app determines how to treat an activity. 

Update: I wanted to add that I could see a rename of `data_source` for clarity.

app/app/models/activity_flow.rb
daphnegold: Considered it, but not in this PR perhaps. I think it's okay to keep hydration branches separate and explicit for clarity, for now. Especially since education is a special case.

app/db/schema.rb
bencalegari: Yeah maybe something like `attestation_level`?

PR #1773, opened 2026-06-03
FFS-4394: Tweak ECS deploy service_name

iannorriswork:
## Changes
Forgot to check in a fix to prior PR before merging because the new github action meant had to test only locally vs via the action. 🤦 

This is a modification in support of the github action calling it like:
```
      - name: Deploy release
        shell: bash
        run: |
          make APP_NAME=${{ inputs.app_name }} \
            ENVIRONMENT=${{ inputs.environment }} \
            EMMY_IMAGE_REPO=${{ env.EMMY_IMAGE_REPO }} \
            EMMY_IMAGE_TAG=${{ inputs.image_tag }} \
            release-update-ecs
```

## Acceptance testing

- [ X ] No acceptance testing needed

### DIFF
--- a/bin/deploy-ecs-manual
+++ b/bin/deploy-ecs-manual
@@ -5,6 +5,9 @@ set -euo pipefail
 # without relying on Terraform to manage the task definition resource.
 #
 # Usage: bin/deploy-ecs-manual <app_name> <environment> <image_repo> <image_tag>
+#
+# cluster_name is presumed as emmy-{environment}
+# service_name is presumed as emmy-{environment}-app
 
 app_name="$1"
 environment="$2"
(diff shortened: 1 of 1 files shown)


bin/deploy-ecs-manual
```diff
@@ -59,7 +57,7 @@ new_task_def=$(echo "${current_task_def}" | jq --arg IMAGE "${image_repo}:${imag
   del(.compatibilities) |
   del(.registeredAt) |
   del(.registeredBy) |
-  (.containerDefinitions[] | select(.name == "'"${service_name}"'")) .image = $IMAGE
+  (.containerDefinitions[] | select(.name == "'"emmy-app"'")) .image = $IMAGE
```
tdooner, line 60: So the container definition name is always "emmy-app"? Just checking this is correct.

tdooner (approved):

bin/deploy-ecs-manual
iannorriswork: resource "aws_ecr_repository" "app" {
  count = var.ecr_repository_app_url == null ? 1 : 0
  name  = "${var.prefix}-app"

where var.prefix = emmy

PR #1794, opened 2026-06-08
FFS-4425: Enable specific agencies to run in an iframe

bencalegari:
## [FFS-4425](https://jiraent.cms.gov/browse/FFS-4425)
Screenshot 2026-06-08 at 4 22 01 PM

## Changes
- Add per-agency `allowed_iframe_ancestors` config (validated as a list) that sets a CSP `frame-ancestors` allowlist so opted-in agencies can be embedded in cross-origin iframes; `dsacms.github.io` allowed for sandbox.
- Remove the Rails default `X-Frame-Options: SAMEORIGIN` header for embedding agencies so CSP `frame-ancestors` is the single source of truth.
- Issue session and device-id cookies as `SameSite=None; Secure` for embedding agencies so the session survives inside a cross-origin iframe; other agencies keep the stricter `SameSite=Lax` default.
- Enable Action Cable cross-origin requests in development only when an agency has iframe embedding configured.
- ngrok dev: point the sandbox agency's `agency_domain` at the active tunnel host so it resolves by domain when embedded over ngrok.
- Add `docs/iframe-test-harness.html` for manually testing iframe embedding from a real origin. 
- Add request specs for CSP/X-Frame-Options/cookie behavior and config specs for `allowed_iframe_ancestors`.

## AI Usage

## Context for reviewers

- The embedding relaxations (CSP allowlist, dropped `X-Frame-Options`, `SameSite=None; Secure` cookies, dev Action Cable forgery protection) are all gated on `current_agency.allowed_iframe_ancestors` being present, so non-embedding agencies are unaffected.
- `X-Frame-Options` and CSP `frame-ancestors` interact: a browser blocks framing if *either* disallows it, so `X-Frame-Options` must be removed for the allowlist to take effect.
- `allowed_iframe_ancestors` must be a YAML list — a string raises an `ArgumentError` at config load.
- Local testing: uncomment the `localhost:8000` / `*.ngrok-free.app` entries in `client-agency-config.yml`, then serve the harness via `cd docs && python3 -m http.server 8000`. Cookies need `Secure`, so a cross-origin iframe requires HTTPS (ngrok).

## Acceptance testing

### Local: 
1. Run and open the harness html from the project root like so: 
```
cd ../docs && python3 -m http.server 8000
```
2. Open http://localhost:8000/iframe-test-harness.html
3. Use the ngrok link generated in `bin/dev` as the url to load in the iframe. We can't use the site running in localhost because it's not served over HTTPS. 

### On Dev After Merge
1. [Setup github pages](https://github.com/DSACMS/iv-cbv-payroll/settings/pages) on root to host iframe-test-harness
2. Open the harness and confirm it reports origin https://dsacms.github.io.
3. Generate a flow on the demo launcher then paste its start link into the harness
4. Verify end-to-end inside the frame, sync a job and an education, generate and download a report.

### DIFF
--- a/app/spec/requests/iframe_embedding_spec.rb
+++ b/app/spec/requests/iframe_embedding_spec.rb
@@ -17,6 +17,12 @@ RSpec.describe "Iframe embedding", type: :request do
         expect(response.headers["Content-Security-Policy"])
           .to include("frame-ancestors 'self' https://portal.example.com")
       end
+
+      it "removes the X-Frame-Options header that would otherwise block framing" do
+        get path
+
+        expect(response.headers).not_to include("X-Frame-Options")
+      end
     end
 
     context "when the agency does not permit iframe embedding" do
@@ -30,6 +36,39 @@ RSpec.describe "Iframe embedding", type: :request do
         csp = response.headers["Content-Security-Policy"]
         expect(csp).to include("frame-ancestors 'self'")
         expect(csp).not_to include("portal.example.com")
+      end
+
+      it "keeps the X-Frame-Options header" do
+        get path
+
+        expect(response.headers["X-Frame-Options"]).to eq("SAMEORIGIN")
+      end
+    end
+  end
+
+  describe "session cookie SameSite" do
+    context "when the agency permits iframe embedding" do
+      before do
+        stub_client_agency_config_value("sandbox", "allowed_iframe_ancestors", [ "https://portal.example.com" ])
+      end
+
+      it "issues the session cookie as SameSite=None; Secure so it survives a cross-site iframe" do
+        get path
+
+        expect(request.session_options[:same_site]).to eq(:none)
+        expect(request.session_options[:secure]).to be(true)
+      end
+    end
+
+    context "when the agency does not permit iframe embedding" do
+      before do
+        stub_client_agency_config_value("sandbox", "allowed_iframe_ancestors", [])
+      end
+
+      it "does not relax the session cookie SameSite" do
+        get path
+
+        expect(request.session_options[:same_site]).not_to eq(:none)
       end
     end
   end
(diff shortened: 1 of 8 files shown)


docs/iframe-test-harness.html
```diff
@@ -0,0 +1,113 @@
...
+<body>
+<header>
+  <h1>Emmy iframe embedding test harness</h1>
+  <form id="frame-form">
+    <input
+      type="url"
+      id="target-url"
+      placeholder="https://sandbox.example.org/help?client_agency_id=sandbox"
+      required
+    />
```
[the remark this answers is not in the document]
bencalegari: Not necessary for a test harness as its not part of the app.

daphnegold (commented):
Can you check that this PR will enable the CSP and cookie changes `/start/:token`, the noted path we will actually be using for the iframe?

app/spec/requests/iframe_embedding_spec.rb
```diff
@@ -0,0 +1,81 @@
+require "rails_helper"
+
+RSpec.describe "Iframe embedding", type: :request do
+  # Use a lightweight, agency-scoped page that resolves `current_agency` from
+  # the `client_agency_id` query param and renders without creating a flow.
+  let(:path) { "/help?client_agency_id=sandbox" }
```
daphnegold, line 6: Should we add coverage for the actual tokenized link path to confirm that works?

bencalegari: Good suggestion. I was treating the spec as hitting an arbitrary endpoint but this is better

daphnegold (approved):
I think we might still have a cookie issue in the token path. New coverage tests headers, but not cookies, so not confident. I’m okay testing this in demo and following up if it shows up there, but it could be worthwhile to double check this and adding more coverage.

PR #1805, opened 2026-06-10
Run the reviewer lottery only when no human reviewer is already requested

bencalegari:
## Changes
Adding copilot as a reviewer meant that the review lottery action was skipped as it already detected a reviewer. Now it will run only when no human reviewer is already requested. A reviewer set at PR creation (e.g. by the auto-fix-vulnerabilities workflow) is type "User" and still preserved.

## Acceptance testing
It can't be accepted until its committed, after that we can check it against another PR. 

### DIFF
--- a/.github/workflows/pr-reviewer-lottery.yml
+++ b/.github/workflows/pr-reviewer-lottery.yml
@@ -1,23 +1,19 @@
 name: "Add random reviewer via Reviewer Lottery"
 on:
   pull_request_target:
-    types: [opened, ready_for_review, reopened]
+    types: [ opened, ready_for_review, reopened ]
 
 jobs:
   reviewer-lottery:
-    # Skip when the PR already has a requested reviewer — preserves any reviewer
-    # set at PR creation (e.g. by the auto-fix-vulnerabilities workflow) across
-    # the close/reopen dance that the autofix PR body instructs maintainers to
-    # perform to force CI.
     if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-reviewer-lottery.yml
```diff
@@ -1,23 +1,19 @@
...
 jobs:
   reviewer-lottery:
-    # Skip when the PR already has a requested reviewer — preserves any reviewer
-    # set at PR creation (e.g. by the auto-fix-vulnerabilities workflow) across
-    # the close/reopen dance that the autofix PR body instructs maintainers to
-    # perform to force CI.
     if: |
       github.actor != 'dependabot[bot]' &&
-      toJSON(github.event.pull_request.requested_reviewers) == '[]'
+      !contains(github.event.pull_request.requested_reviewers.*.type, 'User')
```
millerti, line 10: Might be worth preserving the above comment and further explaining that we need to make an exception for the Copilot reviewer.

millerti (approved):
How annoying. Thanks for fixing. Adding one suggestion.

.github/workflows/pr-reviewer-lottery.yml
bencalegari: I prefer to leave it in the commit message since a comment can get out of date with the code its referring to but the commit will always tie the intent to that point in history.

PR #1808, opened 2026-06-10
[FFS-4474] fix issue with cookie store

iannorriswork:
https://api.rubyonrails.org/classes/ActionDispatch/Session/CookieStore.html its expire_after not expires_after

## Changes

DPW noticed that sometimes peoples sessions weren't expiring if it held on to the cookie/back-button'd. We synced about it and found this issue.
see https://api.rubyonrails.org/classes/ActionDispatch/Session/CookieStore.html --> it's expire_after not expires_after.

Couldn't figure out a test for this that wasn't goofy.

## Acceptance testing

## AI Usage

### DIFF
--- a/app/config/environments/test.rb
+++ b/app/config/environments/test.rb
@@ -62,8 +62,6 @@ Rails.application.configure do
   # Use TestAdapter in test suite
   config.active_job.queue_adapter = :test
 
-  config.cbv_session_expires_after = 100.years
-
   # Raises error for missing translations.
   # config.i18n.raise_on_missing_translations = true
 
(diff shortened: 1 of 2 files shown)


neilmb (approved):

daphnegold (commented):
Why do we need `Rails.env.test?` here? It seems like the actual fix is just the typo.

iannorriswork:
@daphnegold guess its not obvious from the commit history haha but i think there's something goofy about the way our e2e tests work that causes things to explode if the sessions actually are allowed to time out. I tried setting them to 100 years expiration and that didn't work, so I wonder if there's something about the caching that causes the cookies to be set back in a way that throws off the e2e tests. 

The way Tom had "fixed" this for the e2e tests was through that typo that ended up removing any session deletion at all. So this seemed like an alternative that gives us a win and lets us decide the priority of fixing whatever weird cookie session timestamp later.

daphnegold (approved):

PR #1848, opened 2026-06-25
FFS-4499: Use SHA image tags for UAT deploys

daphnegold:
## Changes

- Added a new "Deploy to CMS" wrapper workflow to ease deploys with a single workflow dispatch until transition
- CMS build and publish workflow reruns don’t fail when the SHA image tag already exists
- There's an image check so we fail early if someone tries to deploy a tag that does not exist in ECR

I tested the individual pieces from this PR because the new wrapper workflow can’t be dispatched until it exists on main. I'll follow up with testing it after merge.

### Test runs
1. [First build & publish](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/28182998650)
2. [Second build & publish (testing already-published SHA)](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/28183325024)
3. [Deploy to ECS](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/28183363582)
4. [Deploy to CMS (post-merge)](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/28187224798)

**Note:** Verified the wrapper works after merging to main!

## Acceptance testing

## AI Usage

### DIFF
--- a/.github/workflows/deploy-cms.yml
+++ b/.github/workflows/deploy-cms.yml
@@ -1,13 +1,9 @@
 name: Deploy to CMS
-run-name: Deploy ${{ inputs.ref }} to CMS ${{ inputs.app_name }} (${{ inputs.environment }})
+run-name: Deploy ${{ inputs.ref }} to CMS app (${{ inputs.environment }})
 
 on:
   workflow_call:
     inputs:
-      app_name:
-        description: "name of application folder under infra directory"
-        required: true
-        type: string
       environment:
         description: "the name of the application environment (e.g. uat)"
         required: true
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy-cms.yml
```diff
@@ -0,0 +1,81 @@
...
+        description: "the name of the application environment (e.g. uat)"
+        required: true
+        type: string
+      ref:
+        description: "the branch, tag, or SHA to deploy"
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      app_name:
```
bencalegari: I guess it doesn't hurt to make it configurable, but is there any reason we'd dispatch anything other than `app` here without changing it everywhere?

bencalegari (approved):

.github/workflows/deploy-cms.yml
daphnegold: Good call, removed it from the wrapper and hardcoded app. No need to bubble this up to workflow dispatch from the lower-level code.

PR #1854, opened 2026-06-29
FFS-4526: Deploy to Solid Queue ECS container too

tdooner:
## Context for reviewers

This updates the deploy script to also bump the image_tag of the
solid-queue worker container(s) when deploying an app container.

## Acceptance testing

## AI Usage

### DIFF
--- a/bin/deploy-ecs-manual
+++ b/bin/deploy-ecs-manual
@@ -1,13 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-# This script updates an ECS service's task definition with a new image
+# This script updates ECS service task definitions with a new image
 # without relying on Terraform to manage the task definition resource.
 #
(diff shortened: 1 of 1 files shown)


daphnegold (commented):
Blocking: could you rebase and make sure we keep the check that the image exists before updating ECS?

Non-blocking: Maybe worth considering the AWS GitHub Actions rather than manually editing the task definition? I don't think that needs to block this PR, but it might be easier to reason about long term and more stable.

tdooner:
@daphnegold thanks for the review!

> Blocking: could you rebase and make sure we keep the check that the image exists before updating ECS?

Rebased! The image check remains.

> Non-blocking: Maybe worth considering the AWS GitHub Actions rather than manually editing the task definition? I don't think that needs to block this PR, but it might be easier to reason about long term and more stable.

Agreed, that is part of https://jiraent.cms.gov/browse/FFS-4483.

daphnegold (approved):
Looks great!

PR #1867, opened 2026-06-30
[FFS-4528] create custom testing configuration base for hawaii

iannorriswork:
## Changes
Creates base configuration for Hawaii to be able to play with in the demo environment. 

## Acceptance testing

- [ X ] Acceptance testing in PR Environment

## AI Usage
- [ ] USED_AI: I attest that I have read, understood, and take ownership of all AI-generated code in this PR.

## Infrastructure Changes

  - [ ] Plan reviewed


infra/app/app-config/env-config/environment-variables.tf
neilmb, line 1: **question (non-blocking)**: Do you also need to add the new `HAWAII_INCOME_REPORT_URL` to this file around https://github.com/DSACMS/iv-cbv-payroll/blob/ee2790021b5b127541f4b43e030e3b22e55c0f44/infra/app/app-config/env-config/environment-variables.tf#L93

iannorriswork: good catch, handled

neilmb (approved):

app/config/client-agency-config.yml
```diff
@@ -213,3 +213,45 @@
...
+    first_name:
+      required: true
+    middle_name:
+      required: false
+    last_name:
+      required: true
+    case_number:
+      required: true
+    date_of_birth:
+      required: true
```
bencalegari, line 248: This list needs to match the applicant's `VALID_ATTRIBUTES` otherwise it you won't be able to create an invitation link through the API. We shouldn't have to define this both places though, I may attempt a refactor to remedy it but for now just copy these to `hawaii.rb`.

bencalegari (changes requested):

PR #1886, opened 2026-07-07
FFS-4316: Only list prefilled activities in a hub

bencalegari:
## [FFS-4316](https://jiraent.cms.gov/browse/FFS-4316)
Now only the invitation's activity types render on the hub, anything not provided won't be shown. 

## Context for reviewers
- Made a little refactor: the old `case`/constant approach had the same block of activity / month creation four separate times, it's now one loop keyed off `ACTIVITY_TYPES`. If (in the unlikely event) that we add an activity type, there are required fields and methods to implement that the `activity_spec.rb` enforces.
- You can no longer pre-populate activities that an agency doesn't support. Now it just boils down to what activities are enabled for that agency in its config.

## Acceptance testing

## AI Usage

## Preview environment
- Link to launcher: https://p-1886.navapbc.cloud/launcher/advanced
- Deployed commit: 7754079c175e79f347cd16d70c405bb8d2e97b7a

### DIFF
--- a/app/app/helpers/activities_helper.rb
+++ b/app/app/helpers/activities_helper.rb
@@ -1,4 +1,12 @@
 module ActivitiesHelper
+  def show_activity?(type)
+    if @flow&.pre_populated_session?
+      @flow.pre_populated_activity_types.include?(type.to_sym)
+    else
+      activity_type_enabled?(type)
+    end
+  end
+
   def activity_hub_state(any_activities_added:, monthly_results:, required_month_count: monthly_results.length)
     return :empty unless any_activities_added
 
--- a/app/app/models/education_activity.rb
+++ b/app/app/models/education_activity.rb
@@ -20,7 +20,7 @@ class EducationActivity < Activity
   has_many :education_activity_months, dependent: :destroy
   has_activity_months :education_activity_months
 
-  def self.display_name
+  def self.activity_type
     :education
   end
 
--- a/app/app/models/education_activity.rb
+++ b/app/app/models/education_activity.rb
@@ -14,10 +14,19 @@ class EducationActivity < Activity
     contact_email
     contact_phone_number
   ].freeze
+  PRE_POPULATED_REQUIRED_FIELDS = %w[school_name].freeze
 
   has_many :nsc_enrollment_terms, dependent: :destroy
   has_many :education_activity_months, dependent: :destroy
   has_activity_months :education_activity_months
+
+  def self.display_name
+    :education
+  end
+
+  def self.pre_populated_defaults
+    { "data_source" => "fully_self_attested" }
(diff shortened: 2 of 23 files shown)


app/app/models/education_activity.rb
```diff
@@ -14,11 +14,20 @@ class EducationActivity < Activity
...
   ].freeze
+  PRE_POPULATED_REQUIRED_FIELDS = %w[school_name].freeze
 
   has_many :nsc_enrollment_terms, dependent: :destroy
   has_many :education_activity_months, dependent: :destroy
   has_activity_months :education_activity_months
 
+  def self.display_name
+    :education
+  end
```
millerti: Is "display_name" accurate if we're returning a symbol?

app/spec/models/activity_spec.rb
```diff
@@ -45,4 +45,35 @@
       expect(build(:volunteering_activity, activity_flow: flow, draft: true, pre_populated: false)).not_to be_pre_populated_draft
     end
   end
+
+  describe ".display_name" do
+    it "raises NotImplementedError on the abstract base" do
+      expect { described_class.display_name }.to raise_error(NotImplementedError)
+    end
+  end
```
millerti: Do we do this elsewhere? Seems of limited use.

app/app/helpers/activities_helper.rb
```diff
@@ -1,4 +1,12 @@
 module ActivitiesHelper
+  def show_activity?(type)
+    if @flow&.pre_populated_session?
+      @flow.pre_populated_activity_types.include?(type.to_sym)
+    else
+      activity_type_enabled?(type)
+    end
```
millerti, line 7: You say in the notes that I can only pre-populate activities that are configured. But I disabled some activities locally and was still able to pre-populate them via the launcher.

I think what you assert in the notes makes sense. If a state disables an activity but sends us info about it... maybe we should hide it. Or throw an error and talk to them about it.

millerti (approved):
Looks pretty good! A couple small questions.

rutvigupta-design:
Looks great to me!

app/spec/models/activity_spec.rb
bencalegari: I want to defend against creating/modifying an activity without defining this method, otherwise it will break and it won't be easy to tell.

app/app/helpers/activities_helper.rb
bencalegari: Good catch. The launcher makes invitations in ways that can drift from the API that is a pain and unrealistic to boot.. I'll move the validation into the model and make it clear in the launcher that you can't select activities that haven't been enabled for that agency.

PR #1894, opened 2026-07-10
FFS-4484: Enable CMS Cloud migration action (and simplify it)

bencalegari:
## [FFS-4484](https://jiraent.cms.gov/browse/FFS-4484)

**Run migrations in CMS environments**

- Added a `migrate` job to `deploy-cms.yml`, sequenced between `build-and-publish` and `deploy` (build → migrate → deploy).
- Rewrote `database-migrations-cms.yml` to use AWS actions instead of the custom shell script.
- Updated `.github/workflows/README.md` to document the new workflow and reordering.

## Context for reviewers
- Migrations will now run automatically on every CMS deploy
- Network configuration is derived at runtime from the running `emmy-<env>-app` service rather than hardcoded, so the migrate task lands in the same subnets/security groups as the app.

## Acceptance testing
Ran migration task [successfully on UAT here.](https://github.com/DSACMS/iv-cbv-payroll/actions/runs/29119872390/job/86453351134)

## AI Usage


neilmb (approved):
This approach looks great. Thanks.

tdooner (approved):
Nice. It feels like a simplification to move all this stuff out of those scripts into the deploy workflows. I suppose we will want to document a way to run db migrations manually (in case they fail in GH actions or something) but that will probably be easier after my work is done (to allow arbitrary execution of commands).

PR #1929, opened 2026-07-24
FFS-4545: Delete delivered documents during redaction 

bencalegari:
## [FFS-4545](https://jiraent.cms.gov/browse/FFS-4545)
- Delete uploaded documents past the retention window using `DataRetentionService#delete_delivered_documents`
- Create`data_deletion:delete_delivered_documents` rake task for independent deletion/testing. 
- Log deleted blob keys, activity flow id, and deletion timestamp for each processed flow.
- Add `processed` S3 service to `config/storage.yml` backed by `PROCESSED_BUCKET_NAME` which is how the var is exposed in CMS environments.

## Context for reviewers
- Document deletion reuses the existing `REDACT_ACTIVITY_FLOW_SUMMARIES_AFTER` window (7 days), so documents disappear on the same schedule as the summaries they support. It's keyed off `completed_at` since we don't transmit it yet but that should be updated when implemented. 
- `attachment.purge` deletes the blob and the stored file synchronously (not `purge_later`), so a rake run is slower but leaves nothing queued.
- `update_column` skips validations/callbacks and does not touch `updated_at` since this is a bookkeeping stamp, not a user-facing change. I don't hold strongly to this opinion so if it's not what you'd expect let me know and I'll have it update. 
- This will not work in Nava environments as it stands today, it assumes `PROCESSED_BUCKET_NAME` is set in each environment.

## Acceptance testing

## AI Usage

## Infrastructure Changes
[See this PR in emmy-infra.](https://github.com/CMS-Enterprise/emmy-infra/pull/20)


app/app/services/data_retention_service.rb
```diff
@@ -121,6 +130,29 @@ def self.redact_case_numbers_by_agency(agency_id)
 
   private
 
+  def delete_documents_for(activity_flow)
```
millerti, line 139: If this errors, what catches it? Want to make sure we log and move onto the next doc instead of erroring out of the whole redaction process

app/app/services/data_retention_service.rb
```diff
@@ -96,6 +97,14 @@ def redact_activity_flow_summaries
       end
   end
 
+  def delete_delivered_documents
+    ActivityFlow
+      .completed
```
millerti: Is this the right flag to look at to decide if we can delete? I notice we don't have transmitted_at on ActivityFlow like we had on CbvFlow. Will we need to go back and update this later when we have that? Or am I missing some reason that this works fine?

millerti (commented):
I have a couple questions, generally looks pretty clean

app/app/services/data_retention_service.rb
bencalegari: Good point, adding that fix.

app/app/services/data_retention_service.rb
bencalegari: Good catch, I wanted to have them key off the same thing but for now we can just use completed_at.

millerti: I think we're going to have to create a transmitted_at field here because right now I think this will just redact reports after the user clicks submit, but not necessarily after our SolidQueue job finishes delivering. For example, if we had the same schema issue we're currently dealing with for CE reports, this would redact docs way before we were able to manually send those reports.

daphnegold: I think this concern might not be fully addressed still. I see in a later commit you added the `transmitted_at` field, but where does it get set on the `ActivityFlow`? How will this query find rows to delete?

daphnegold (commented):
One question!

daphnegold (approved):

PR #1955, opened 2026-08-04
FFS-4542: Upload supporting docs directly to S3 client-side

bencalegari:
## [FFS-4542](https://jiraent.cms.gov/browse/FFS-4542)
Here are the steps for how the upload works:
1. Browser validates type and size and hashes with SHA-256.
2. POST `/activities/presigned_uploads` to the server to create a Blob row (just metadata) and signs an upload policy locally, granting permission to put bytes at its key. Responds with a url, fields, and signed_id per file. 
3. Browser POSTs the bytes to url with those fields.
  - CMS: Goes straight to the S3 bucket. S3 validates the policy conditions, exact key, exact content type, 0–25 MB all against the signature.
  - Nava: POST `/activities/local_uploads`, Rails receives the bytes and does the validating (flow session, key format, size) and writes the file to the container's ephemeral disk.
4. Signed ID is set in a hidden field.
5. On Continue, Rails resolves all the submitted signed IDs and sets the activity's attachments.

Here's a diagram showing the different layers of interaction:
```
browser                                      rails          s3
───────                                      ─────          ──
select file(s)
  ├─ validate type + size (≤25 MB)
  ├─ SHA-256 via crypto.subtle, one file at a time
  └─ POST /activities/presigned_uploads ──────►│
       { filename, content_type,               │  create Blob row per file
         byte_size, checksum }                 │  (identified + analyzed preset)
                                               │  sign POST policy
     ◄─────────────────────────────────────────┘
     [{ url, fields, signed_id }, …]
  └─ FormData: policy fields first, file LAST
     POST ─────────────────────────────────────────────────►│ enforces:
                                                            │  • exact key
                                                            │  • exact content-type
     ◄──────────────────────────────────────────── 204 ─────┘  • 0..25 MB
  └─ append <input type=hidden
       name="activity[document_uploads][]" value=signed_id>

[Continue]
  └─ POST /activities/…/document_uploads ─────►│ attach blobs by signed_id
       signed_ids only, no bytes               │ (rows already exist)
```

### Nava / Local environments
If there's no quarantine bucket, `PresignedUploadService#destination_for` returns
`/activities/local_uploads` instead of a bucket URL and adds `authenticity_token`
to the form fields.
```
browser                                      rails
───────                                      ───── 
     POST /activities/local_uploads ───────────►│ Activities::LocalUploadsController
                                                │  • flow session required
                                                │  • CSRF via form field
                                                │  • key ~ /\A[a-z0-9]{28}\z/
                                                │  • ≤ 25 MB
     ◄─────────────────────────────── 204 ──────┘  write → DiskService (ephemeral)
```

---
- Add `PresignedUploadService`: validates requested files and returns a presigned POST policy + signed ID per file.
- Add `document_upload` Stimulus controller: client-side validation, browser-computed SHA (hashed a file at a time to not blow up memory usage), direct POST to the bucket, then submits signed IDs as hidden fields with an add/remove file list.
- `DocumentUploadable` now attaches to the unscanned service instead of the old bucket.
- Add `LocalUploadsController`: When `UNSCANNED_BUCKET_NAME` is blank so local/test/Nava env uploads exercise the same path.
- Add `UnattachedUploadCleanupJob`, scheduled hourly, to destroy unattached unscanned blobs older than 24 hours. Blobs are created when they are first put into the file selector and if they are removed after the fact their objects remain. 
- `document_uploads#destroy` now deletes the attachment and blob record without purging the remote object.

## Context for reviewers
- Because we create blobs before the form is submitted, abandoned uploads accumulate, hence the hourly cleanup job with a 24 hour cutoff (long enough that a user still on the page isn't affected).
- `SERVICE_NAME` is resolved at class-load time from `ENV["UNSCANNED_BUCKET_NAME"]`, without the env var everything falls back to local disk. In Nava environments it will keep working but the docs will just sit on the container until they are turned over. Otherwise nothing ever deletes them.  

## Acceptance testing
We have to test this in CMS-land once it's deployed, but we can confirm the local upload path move on a PR env. 

## AI Usage

## Infrastructure Changes
See https://github.com/CMS-Enterprise/emmy-infra/pull/24

## **Risk / Downtime:**
First significant divergence between Nava / CMS infra. 

## Preview environment
- Link to launcher: https://p-1955.navapbc.cloud/launcher/advanced
- Deployed commit: 180a55535e97ca1ae0458c0d87019b050c768d35


app/config/routes.rb
```diff
@@ -144,6 +145,10 @@
     end
   end
 
+  if ENV["UNSCANNED_BUCKET_NAME"].blank?
+    post "/uploads/fake_s3", to: "fake_s3_uploads#create", as: :fake_s3_uploads
+  end
```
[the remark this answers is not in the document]
bencalegari: This is intentional: Nava envs won't have that set.

app/app/services/presigned_upload_service.rb
```diff
@@ -0,0 +1,100 @@
...
+  private
+
+  attr_reader :service, :authenticity_token
+
+  def normalize(file)
+    {
+      filename: file[:filename].to_s,
+      content_type: file[:content_type].to_s,
+      byte_size: file[:byte_size].to_i,
+      checksum: file[:checksum].to_s
```
daphnegold, line 41: The `fileChecksum` method generates a `SHA 256` value, but, seems like Active Storage expects `MD5` by default.

app/app/javascript/controllers/document_upload_controller.js
```diff
@@ -0,0 +1,209 @@
...
+        return
+      }
+    }
+
+    this.#showError(null)
+
+    try {
+      const uploads = await this.#requestPresignedUploads(files)
+
+      await Promise.all(uploads.map((upload, index) => this.#upload(files[index], upload)))
```
daphnegold, line 45: Do we want to let the user continue while the upload is still pending?

app/app/services/presigned_upload_service.rb
```diff
@@ -0,0 +1,100 @@
+require "aws-sdk-s3"
+
+class PresignedUploadService
+  SERVICE_NAME = ENV["UNSCANNED_BUCKET_NAME"].present? ? :unscanned : :unscanned_local
+  MAX_UPLOAD_BYTES = 25.megabytes
```
daphnegold, line 5: The ticket specified 10 mb

app/app/services/presigned_upload_service.rb
```diff
@@ -0,0 +1,100 @@
...
+    if service.respond_to?(:bucket)
+      config = service.client.client.config
+
+      post = Aws::S3::PresignedPost.new(
+        config.credentials,
+        config.region,
+        service.bucket.name,
+        key: key,
+        content_type: content_type,
+        content_length_range: 0..MAX_UPLOAD_BYTES,
```
daphnegold: Should upload require the byte size match what was used to create the blob rather than a range?

daphnegold (commented):
Had a couple questions, but looking great overall!

app/app/services/presigned_upload_service.rb
bencalegari: I changed it to 25 because I noticed [the Lambda limit was 25MB](https://github.com/CMS-Enterprise/emmy-infra/blob/444ad3989c174c0c251b9c7fe6364d5ee7f63735/emmy/modules/upload_processing/lambda/src/handler.rb#L8).. I think that's probably too high and but I decided to match it for now.. think I should change both? I checked [the ticket](https://jiraent.cms.gov/browse/FFS-4144) for when you added the limit and didn't see anything specified.

app/app/javascript/controllers/document_upload_controller.js
bencalegari: No we do not, good idea. I'll disable submit.

app/app/services/presigned_upload_service.rb
daphnegold: Ohh, I chose that. I'm fine with leaving it at 25! agree it's high. We can adjust later.

app/app/services/presigned_upload_service.rb
bencalegari: That's right, I decided not to add md5 to cut down on the JS dependency I would've had to add but made a change to explicitly check against SHA256 equality.

app/app/services/presigned_upload_service.rb
bencalegari: Thanks, fixed

daphnegold (approved):
Nice! Rspec is failing, though

PR #1957, opened 2026-08-07
FFS-4634: Implement shared event filing for Emmy CE

millerti:
## Changes
Added the Trackable concern to allow Emmy CE flows to easily file Mixpanel events.
Emmy Income flows should be entirely unchanged.
Created the HubViewed event as a demonstration of how to use the concern.

## Context for reviewers
Noticed that the continuing dichotomy of flow vs cbv_flow causes some complexity; we might want to do a cleanup ticket soon.

## Acceptance testing

## AI Usage

## Preview environment
- Link to launcher: https://p-1957.navapbc.cloud/launcher/advanced
- Deployed commit: 89bb7f7b07cdaa8acf1bc489b6f4cf8813d8e31a

### DIFF
--- a/app/app/controllers/concerns/trackable.rb
+++ b/app/app/controllers/concerns/trackable.rb
@@ -1,0 +1,43 @@
+# Shared Mixpanel event-tracking helper for both the CBV (Income) and
+# Activities (CE) flows. See the taxonomy design doc for the full naming
+# convention and standard-parameter list:
+# docs/superpowers/specs/2026-08-06-ce-mixpanel-taxonomy-confluence-page.txt
+#
+# Usage from a controller:
+#
+#   track_event(TrackEvent::HubViewed, some_extra_param: "value")
+#
+# `client_agency_id`, `cbv_flow_id`, `cbv_applicant_id`, and `invitation_id`
+# are attached automatically from `current_agency`/`@flow`. `time`,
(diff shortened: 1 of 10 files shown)


app/app/controllers/concerns/trackable.rb
```diff
@@ -0,0 +1,25 @@
+module Trackable
+  extend ActiveSupport::Concern
```
tdooner, line 2: No need for this since the commit doesn't actually use any of the `Concern` methods (like `included` or `class_methods` blocks) but I guess it doesn't hurt to have in case we want to add those later.

app/spec/controllers/activities/activities_controller_spec.rb
```diff
@@ -4,6 +4,18 @@
...
+  describe "tracking" do
+    let(:current_flow) { create(:activity_flow) }
+    let(:tracked_flow) { current_flow }
+    let(:perform_tracked_action) do
+      session[:flow_id] = current_flow.id
+      session[:flow_type] = :activity
+      get :index
+    end
+
+    it_behaves_like "tracks an event", TrackEvent::HubViewed
```
tdooner: nit: Would prefer that this is included in the `describe "#index"` group to benefit from the same setup, but not a huge deal.

tdooner (approved):

PR #1976, opened 2026-08-12
FFS-4555: Send activity documents over HTTP

tdooner:
## Changes

Add an HTTP activity-document transmitter that posts cleared uploads
with normalized filenames and signed headers. Make the sandbox receiver
and configuration support local end-to-end document transmission.

Extract common document management logic into ActivityDocumentsService.

## Acceptance testing

## AI Usage

- AI tools used: Codex

### DIFF
--- a/app/app/services/activity_documents_service.rb
+++ b/app/app/services/activity_documents_service.rb
@@ -1,0 +1,49 @@
+require "rack/mime"
+
+class ActivityDocumentsService
+  ACTIVITY_ASSOCIATIONS = %i[
+    volunteering_activities
+    job_training_activities
+    education_activities
+    employment_activities
+  ].freeze
+
+  Document = Struct.new(:attachment, :file_name)
+
+  def initialize(activity_flow)
+    @activity_flow = activity_flow
+  end
+
+  def all
+    page_numbers = Hash.new(0)
+
+    ACTIVITY_ASSOCIATIONS.flat_map do |association|
+      @activity_flow.public_send(association).published.order(:id).flat_map do |activity|
+        activity.document_uploads_attachments.includes(:blob).order(:id).map do |attachment|
+          activity_type = activity.class.activity_type
+          document_name, extension = document_name_and_extension(attachment)
+          page_number_key = [ activity_type, document_name ]
+          page_number = page_numbers[page_number_key] += 1
+
+          Document.new(
+            attachment,
+            [ @activity_flow.confirmation_code, activity_type, document_name, page_number ].join("_") + extension
+          )
+        end
+      end
+    end
+  end
+
+  private
+
+  def document_name_and_extension(attachment)
+    file_name = attachment.filename.to_s
+    original_extension = File.extname(file_name)
+    extension = original_extension.downcase
+    extension = Rack::Mime::MIME_TYPES.key(attachment.blob.content_type) || ".bin" if extension.blank?
+
+    document_name = File.basename(file_name, original_extension).parameterize(separator: "_")
+    document_name = "document" if document_name.blank?
+    [ document_name, extension ]
+  end
+end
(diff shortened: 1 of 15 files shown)


app/app/services/activity_documents_service.rb
```diff
@@ -0,0 +1,49 @@
+require "rack/mime"
+
+class ActivityDocumentsService
```
daphnegold, line 3: I really like pulling this out into its own service!

app/app/services/transmitters/http_document_transmitter.rb
```diff
@@ -0,0 +1,111 @@
+require "tmpdir"
+
+class Transmitters::HttpDocumentTransmitter
+  TRANSMISSION_METHOD = "http"
+
+  def initialize(activity_flow, current_agency)
+    @activity_flow = activity_flow
+    @current_agency = current_agency
+    @processed_s3_service = S3Service.new("bucket" => ENV.fetch("PROCESSED_BUCKET_NAME"))
```
daphnegold: You mentioned supporting "local" end-to-end transmission, but I don't think that will work with this setup since local uploads never make it to the processed bucket. I'm fine with that, but just commenting that the local bit might not work.

app/app/services/transmitters/http_document_transmitter.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def deliver
+    documents = ActivityDocumentsService.new(@activity_flow).all
+
+    Dir.mktmpdir("activity-flow-http-transmission-") do |directory|
+      documents.each do |document|
+        file_path = File.join(directory, document.file_name)
+        @processed_s3_service.download_file(document.attachment.blob.key, file_path)
+        deliver_document(document, File.binread(file_path))
```
daphnegold, line 19: If one doc succeeds and the next fails, will we resend everything again on retry and create duplicates?

daphnegold (approved):
Looking amazing! A couple questionssss.

app/app/services/transmitters/http_document_transmitter.rb
daphnegold: FYI I tried this and got `KeyError: key not found: "PROCESSED_BUCKET_NAME"`. Setup step missing?

tdooner: I'll follow up with you in Slack about how you've been testing this stuff, because I didn't add this environment variable (just copying it from the S3 branch) and also the UNPROCESSED_BUCKET_NAME doesn't seem to be configured anywhere.

app/app/services/transmitters/http_document_transmitter.rb
tdooner: Yes, it will retry everything from scratch, potentially creating duplicates. I was considering adding a per-document `transmitted_at`. I seem to remember chatting about that - do you remember whether we came down in favor of doing that or not?

tdooner:
@daphnegold Can you take another look? I was able to test this locally by abstracting out a ProcessedDownloadService (named after the PresignedUploadService) which uses either local storage or S3, similar to how the uploading works.

app/app/services/transmitters/http_document_transmitter.rb
daphnegold: I recall conversation about `transmitted_at` but not what we decided. I do think we should probably avoid duplicate HTTP sends? But fine with that being outside the scope of this PR. :D

daphnegold (approved):
It works! I tried end-to-end in browser and with the receiver. Nice one!

tdooner:
Since there is no frontend aspect of this work, and Daphne and I were able to test this using the reference receiver, I'm going to merge this and we can acceptance test it later (when it comes time for an agency to use it).

PR #1980, opened 2026-08-14
Instrument Mixpanel events for the Employment Self-Attested flow in Emmy CE

millerti:
Fixes FFS-4635

## Changes
- Added Employment Self-Attested `TrackEvent` constants and wired `track_event` (via the FFS-4634 `Trackable` concern) into `Activities::EmploymentController`, `Activities::Employment::MonthsController`, and the employment scope of `Activities::DocumentUploadsController` so each page view, form submission, and form validation-failure fires a Mixpanel event.
- Extended `ActivityFlowHeaderComponent` with an optional `track_event_prefix:` param and refactored its Stimulus controller to fire `ExitClicked`, `ExitModalConfirmed`, `ExitModalCancelled`, and `BackClicked` events (prefixed with `Employment`) via the existing `trackUserAction` client helper. Modal buttons and the back link are wired via Stimulus targets so the listeners survive USWDS reparenting the modal to `<body>`.
- Set `track_event_prefix: "Employment"` on the four employment views and on `activities/document_uploads/new.html.erb` when it's rendered under the employment scope; other activity flows remain uninstrumented per the ticket's scope.
- Added controller specs using the existing `"tracks an event"` shared example for each new server-side event (viewed / submitted / validation-failed variants).

## Testing instructions
1. Use the testing environment link below
2. Go through the employment self-attestation flow and verify that the events fire as you'd expect. Use the [event taxonomy](https://confluenceent.cms.gov/pages/viewpage.action?pageId=1447694203&spaceKey=SFIV&title=Emmy+CE+Events) as reference for what to expect. (Any ideas for how we could improve the taxonomy page?)
3. Pay attention to the properties on each even (open an event in Mixpanel -- they'll drop on the Dev Env page -- you may want to click the Your Properties filter). Does each event have the info you want/need?
4. Verify that navigating other flows (community service, education, work programs) does NOT fire any `Employment*` events. Similarly verify that Emmy Income events seem to be unaffected.

## Acceptance testing

## AI Usage

## Preview environment
- Link to launcher: https://p-1980.navapbc.cloud/launcher/advanced
- Deployed commit: a3f3ab279a956158657faa5aec2e8ed23234781a


millerti:
# Code Review — FFS-4635: Instrument Mixpanel events for the Employment Self-Attested flow

Scope: commit `adc72179`, 15 files, +293 / −81. Adds Mixpanel event tracking for the Employment activity flow (info, month, review, document upload, and header interactions), plus DOM/JS changes to the shared `ActivityFlowHeaderComponent` to support client-side tracking of Exit / Back / Modal interactions.

## Summary

The overall approach is sound: server-side events fire via `after_action` filters using the existing `Trackable` concern; client-side events go through the same API path as CBV (`trackUserAction` → `/api/events/user_action`). New events are added to `TrackEvent` constants and validated by `Api::UserEventsController`. Test coverage uses the existing `"tracks an event"` shared example. There are a handful of concrete correctness gaps around event firing conditions, one JS reliability issue with tracking on navigation, and some schema inconsistencies worth cleaning up before we depend on this data.

---

## Correctness

### 1. `DocumentUploadsController#track_employment_upload_submitted_event` fires on validation failures — no successful-response guard

`app/app/controllers/activities/document_uploads_controller.rb:197-205`

```ruby
def track_employment_upload_submitted_event
  return unless params[:employment_id].present?

  track_event(
    TrackEvent::EmploymentDocumentUploadSubmitted,
    employment_activity_id: @activity.id,
    document_count: @activity.document_uploads.count
  )
end
```

The `viewed` variant above it (`document_uploads_controller.rb:191-195`) correctly gates on `response.successful?`, but the submitted variant does not. The `create` action can end in three states:

1. `params.exclude?(:activity)` → redirects to `after_activity_path` (3xx). Still fires `EmploymentDocumentUploadSubmitted` with `document_count: 0` — reporting a "submission" when the user hit Submit with no files attached.
2. `@activity.update(document_upload_params)` succeeds → 3xx. Fires.
3. Update fails → `render :new` (`4xx`). Fires — reporting a submission that never persisted.

Compare to `EmploymentController#create` (`employment_controller.rb:23-30`), which cleanly splits into `EmploymentInfoSubmitted` (success) vs. `EmploymentInfoValidationFailed` (failure). The document-upload path should either add a `response.successful?` guard, or (better) split into two events to match the naming convention already established for info/month/review.

Additionally, the existing test (`document_uploads_controller_spec.rb:247-253`) sends `post :create, params: { employment_id: ... }` with no `:activity` key — exercising branch (1) above. It passes because the event still fires, but that's masking the ambiguity described here.

### 2. `EmploymentInfoValidationFailed` schema differs between create and update

`app/app/controllers/activities/employment_controller.rb:27` vs `:48`

```ruby
# create failure (line 27) — omits employment_activity_id
track_event(TrackEvent::EmploymentInfoValidationFailed)

# update failure (line 48) — includes it
track_event(TrackEvent::EmploymentInfoValidationFailed, employment_activity_id: @employment_activity.id)
```

Downstream Mixpanel schemas will be inconsistent for the same event name. In the `create` branch `@employment_activity` exists but is unpersisted, so `id` is `nil`. Recommend passing `employment_activity_id: @employment_activity&.id` in both branches to normalize the property set. As-is, funnel analyses that group by `employment_activity_id` will silently drop the create-failure events.

### 3. `MonthsController#track_month_submission_event` fires even on the out-of-range redirect

`app/app/controllers/activities/employment/months_controller.rb:22-29` + `concerns/monthly_hours_input.rb:57-60`

If the user submits `PATCH #update` with an out-of-range `:id`, `set_hours_input_vars` (a before_action) redirects to month 0. The action body never runs — `@error` remains falsy — but `after_action` still runs on redirects. The result: an `EmploymentMonthSubmitted` event with the invalid `@month_index` and no state change. Suggest guarding on `response.successful? || @error` (i.e. fire only when the action actually rendered or reached the error branch of `MonthlyHoursInput#update`).

### 4. Client-side tracking events can be dropped on navigation

`app/app/javascript/controllers/activity_flow_header_controller.js:27-35, 50-59`

`trackUserAction` posts asynchronously via `fetchInternal`. Three of the four bound handlers fire on elements that navigate immediately:

- `modalConfirmButton` — a `link_to exit_url` (`activity_flow_header_component.html.erb:47-50`); default navigation follows the click.
- `backButton` — `link_to back_url` (`activity_flow_header_component.html.erb:77-83`); navigates.
- `handleExit`'s no-modal branch — `window.location.href = this.exitUrlValue` (line 57); navigates.

In all three cases the in-flight `fetch` is likely cancelled by the unload, so `EmploymentExitModalConfirmed`, `EmploymentBackClicked`, and (when `confirmOnExit`/`isDirty` are both false) `EmploymentExitClicked` will be lost intermittently. Preferred fix: use `navigator.sendBeacon(...)` or `fetch(..., { keepalive: true })` in `trackUserAction`, or block navigation until the request resolves. The modal-cancel / modal-close handlers are safe because they don't navigate.

### 5. `EmploymentInfoViewed` on `#edit` fires even on non-2xx responses

`app/app/controllers/activities/employment_controller.rb:14, 109-111`

```ruby
after_action :track_info_viewed_event, only: %i[new edit]
...
def track_info_viewed_event
  track_event(TrackEvent::EmploymentInfoViewed, employment_activity_id: @employment_activity&.id)
end
```

In practice `set_employment_activity` raises `ActiveRecord::RecordNotFound` before the action runs, so `after_action` is skipped — this is probably safe today. But since the neighbouring `track_review_viewed_event` (line 113-117) explicitly guards on `response.successful?`, the inconsistency is worth resolving (add the same guard here) so future non-2xx paths on `#new`/`#edit` don't silently taint the funnel.

### 6. DOM restructuring of the header component

`app/app/components/activity_flow_header_component/activity_flow_header_component.html.erb`

The outer `data-controller="activity-flow-header"` div lost the `.activity-flow-header` class; `.activity-flow-header` now lives on an inner wrapper that also contains the modal and hidden trigger button. `.back-nav` moved from being a sibling of the old outer div to a child of the new outer div (but still a sibling of `.activity-flow-header`).

Reviewed `activity_flow_header_component.scss` — nothing depends on the old sibling relationship, and the modal is `position: fixed` so the sticky container change has no visible effect. Worth eyeballing in the app on both mobile and tablet widths, especially the modal-open interaction with the sticky title (component styles at `activity_flow_header_component.scss:4-8`).

Also: the four new targets (`modalConfirmButton`, `modalCancelButton`, `modalCloseButton`, `backButton`) rely on the target elements being present in the controller's DOM subtree at `connect()` time — before USWDS moves the modal to `<body>`. This ordering holds today because Stimulus's MutationObserver `connect()` runs on mount, prior to USWDS's `turbo:load` init, and event listeners survive DOM reparenting. The inline comment (line 24-26) captures this, but it's a fragile invariant. If a future refactor delays Stimulus connection (e.g., lazy controllers) the modal-button targets will silently fail to bind. Consider using MutationObserver-based rebinding or a lightweight assertion / warn.

---

## Test coverage

### What's tested well
- Server events (`EmploymentInfoViewed`, `EmploymentInfoSubmitted`, `EmploymentInfoValidationFailed`, `EmploymentMonthViewed`, `EmploymentMonthSubmitted`, `EmploymentMonthValidationFailed`, `EmploymentReviewViewed`, `EmploymentReviewSubmitted`, `EmploymentDocumentUploadViewed`, `EmploymentDocumentUploadSubmitted`) all have at least one shared-example assertion.
- The month validation-failed branch uses realistic invalid params (`gross_income: 0, hours: 0`) rather than mocking — good.

### Gaps

1. **No JS tests.** `activity_flow_header_controller.js` gained ~40 lines of tracking logic, event bindings, and lifecycle cleanup with zero test coverage. At minimum, a vitest test verifying that `trackUserAction` is called with the correct prefixed event name and that listeners are removed on `disconnect` would catch the most likely regressions (double-binding, missing prefix short-circuit, etc.).

2. **No test for the "no `:activity` param" branch** in `DocumentUploadsController#create` (see issue #1). The existing test at `document_uploads_controller_spec.rb:247-253` accidentally exercises this branch without asserting anything about it. Add a distinct test asserting that clicking Submit with zero files still tracks (or doesn't, depending on intent).

3. **No test that `EmploymentInfoValidationFailed` in `#create` omits `employment_activity_id`** (see issue #2). Add or, better, fix the inconsistency and assert both branches include the id.

4. **Failure-branch tests use `allow_any_instance_of(...).to receive(:save).and_return(false)`** (`employment_controller_spec.rb:57-61, 133-137`). This bypasses ActiveModel validation entirely, so the tests would pass even if the `create`/`update` methods stopped reading `.save` return values correctly. Prefer sending genuinely invalid params (e.g. missing `employer_name`) so the failure path is validated end-to-end.

5. **No test for the out-of-range month index edge case** described in issue #3.

6. **`EmploymentBackClicked` / `EmploymentExitClicked` / `EmploymentExitModal*`** are added to `TrackEvent` but have no server-side test — their only enforcement is the constant check in `Api::UserEventsController#user_action:21` (which is exercised by `api/user_events_controller_spec.rb`). This is fine as long as we're confident the JS emits the exact strings; consider a spec that walks the constant list and asserts the JS strings match.

---

## Security

1. **Event-name validation is good.** `Api::UserEventsController:21` (`TrackEvent.constants.map(&:to_s).include?(event_name)`) prevents arbitrary event names from being injected via the client. New constants are exact-match, so no risk of collision.

2. **Free-form `attributes` payload.** `user_action_params` (`api/user_events_controller.rb:42-44`) permits `attributes: {}` — meaning any keys the client sends are forwarded to Mixpanel and merged with server-supplied identity attributes. Not a regression from this PR, but this PR broadens the surface: any page rendering `ActivityFlowHeaderComponent` with a `track_event_prefix` now has a JS endpoint that will forward arbitrary attributes. If Mixpanel schemas ever gate on attribute keys (or a downstream system trusts them), this is exploitable. Consider allowlisting attribute keys per event, or at least documenting the trust boundary. In this PR only `modal_shown` (boolean) is sent from JS, so no immediate concern.

3. **No PII in the new server-side attributes** — only `employment_activity_id`, `month_index`, `document_count`. Good.

4. **`flow.cbv_applicant.client_agency_id`** is dereferenced in the shared example (`tracks_event_examples.rb:20`) — depends on the `cbv_applicant` association being present on `activity_flow`. Existing pattern; not affected by this PR.

---

## Minor / style

- `activity_flow_header_controller.js:65-67` — `_hasTarget(name)` reimplements Stimulus's built-in `hasXxxTarget` property. Since the target names are fixed at authoring time, prefer the direct `this.hasModalConfirmButtonTarget && this.modalConfirmButtonTarget` idiom; drops five lines and removes reflection.
- `activity_flow_header_controller.js:44` — `(this._trackingBindings || [])` guard in `disconnect()` is defensive against a partial-connect state that Stimulus doesn't produce. Safe to drop.
- `activity_flow_header_component.html.erb:19-20` — `data-activity-flow-header-track-event-prefix-value="<%= track_event_prefix %>"` renders empty string when nil. JS handles this (empty string is falsy in `_trackEvent`), but it's cleaner to omit the attribute entirely when nil, e.g. `data: { "activity-flow-header-track-event-prefix-value": track_event_prefix }.compact`.
- `track_event.rb:83-93` — new constants are alphabetized; nice.
- `document_uploads_controller_spec.rb:110-116` and `:247-253` — both blocks use `let(:tracked_flow) { activity_flow }` where `activity_flow` is defined outermost. Fine, but the redundant let is noise; the shared example could just reference `activity_flow` directly.

---

## Suggested follow-ups (priority order)

1. Fix `DocumentUploadsController#track_employment_upload_submitted_event` to distinguish success from validation-failure (#1).
2. Normalize `EmploymentInfoValidationFailed` attributes across create/update (#2).
3. Move the JS `trackUserAction` call to `keepalive: true` / `sendBeacon` for navigation-tracking events (#4).
4. Add the out-of-range guard to `track_month_submission_event` (#3).
5. Add vitest coverage for `activity_flow_header_controller.js` (test gap #1).
6. Replace `allow_any_instance_of(...).receive(:save/update).and_return(false)` with real-invalid-params tests (test gap #4).

bencalegari (changes requested):
[Here's the logs in Mixpanel](https://mixpanel.com/project/3511740/view/4013544/app/profile#distinct_id=device-142e90cc-100f-49c8-af47-99d2b65038be) for the test. Code looks good but there were two events I was unable to see.
1. `EmploymentExitModalCancelled` - Neither the X nor go back button seemed to trigger it
2. `EmploymentDocumentUploadValidationFailed`

ghost:
lgtm! 🚢 ittt

bencalegari (approved):
Missing modal events [showed up](https://mixpanel.com/project/3511740/view/4013544/app/profile#distinct_id=device-142e90cc-100f-49c8-af47-99d2b65038be) upon retesting. Good to go!

PR #1982, opened 2026-08-17
FFS-4668: Add CMS Cloud review apps

tdooner:
## Changes

Provision per-PR ECS services and database schemas while sharing the CMS
dev ALB and supporting agency-specific hostnames.

Run Solid Queue inside the review app Puma process and clean up
resources automatically when PRs close.

## Acceptance testing

## AI Usage

- AI tools used: Github Copilot

♻️ Environment destroyed ♻️

### DIFF
--- a/.github/workflows/scan-orphaned-cms-pr-environments.yml
+++ b/.github/workflows/scan-orphaned-cms-pr-environments.yml
@@ -1,0 +1,33 @@
+name: Scan orphaned CMS Cloud review environments
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "30 7 * * *"
+
+permissions:
+  contents: read
+  id-token: write
+  pull-requests: write
+
+jobs:
+  scan:
+    name: Remove review apps for closed pull requests
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+
+      - name: Configure CMS Cloud nonprod credentials
+        uses: ./.github/actions/configure-cms-aws-credentials
+        with:
+          environment: dev
+          nonprod-role-to-assume: ${{ vars.AWS_NONPROD_ROLE_TO_ASSUME }}
+          default-role-to-assume: ${{ vars.AWS_ROLE_TO_ASSUME || format('arn:aws:iam::{0}:role/{1}', vars.AWS_ACCOUNT_ID, vars.AWS_ROLE_NAME) }}
+
+      - name: Remove orphaned review apps
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: ./bin/cms-pr-environment scan
--- a/.github/workflows/scan-orphaned-cms-pr-environments.yml
+++ b/.github/workflows/scan-orphaned-cms-pr-environments.yml
@@ -7,6 +7,7 @@ on:
 
 permissions:
   contents: read
+  deployments: write
   id-token: write
   pull-requests: write
 
--- a/bin/cms-pr-environment
+++ b/bin/cms-pr-environment
@@ -1,0 +1,683 @@
+#!/usr/bin/env bash
+# Create, update, destroy, or garbage-collect CMS Cloud dev review apps.
+set -euo pipefail
+
+readonly environment="dev"
+readonly cluster_name="emmy-${environment}"
+readonly base_service_name="emmy-${environment}-app"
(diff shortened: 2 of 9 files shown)


tdooner:
Testing whether the review app gets cleaned up properly.

tdooner:
Testing deletion again...

tdooner:
It worked! Ok, time for review!

bin/cms-pr-environment
```diff
@@ -0,0 +1,922 @@
...
+  if [[ "$cleanup_failed" == "true" ]]; then
+    log "AWS resources were removed, but webhook or schema cleanup failed"
+    return 1
+  fi
+
+  log "Review app destruction completed for PR ${pr_number}"
+}
+
+# Discovers PR numbers from review ECS resources and destroys those whose PRs are no longer open.
+scan_environments() {
```
millerti, line 848: Do we need a guard here so that we continue if something goes wrong?

bin/cms-pr-environment
```diff
@@ -0,0 +1,922 @@
...
+      .family = $family |
+      .containerDefinitions |= map(
+        if .name == $container then
+          .image = $image |
+          .secrets = ((.secrets // []) | map(select(
+            .name != "LA_LDH_DOMAIN_NAME" and
+            .name != "NH_DHHS_DOMAIN_NAME" and
+            .name != "SANDBOX_DOMAIN_NAME" and
+            .name != "RESEARCH_DOMAIN_NAME" and
+            .name != "ACCENTURE_DOMAIN_NAME"
```
millerti: bleh, is there a way to reuse this list? I see we use it in a few lines again, and maybe elsewhere too

bin/cms-pr-environment
```diff
@@ -0,0 +1,922 @@
...
+
+# Builds the common ECS tag payload used to identify and garbage-collect review resources.
+tags_json() {
+  local pr_number="$1"
+  local image_tag="${2:-unknown}"
+  jq -nc \
+    --arg pr_number "$pr_number" \
+    --arg image_tag "$image_tag" \
+    '[
+      {key: "Project", value: "EMMY"},
```
millerti: All caps Emmy huh? 👀 

This is copy-pasted too... maybe an opportunity for reuse?

.github/workflows/scan-orphaned-cms-pr-environments.yml
```diff
@@ -0,0 +1,34 @@
...
+    - cron: "30 7 * * *"
+
+permissions:
+  contents: read
+  deployments: write
+  id-token: write
+  pull-requests: write
+
+jobs:
+  scan:
```
millerti, line 15: Claude is worried about a concurrency lock here. "...it can race with the per-PR locks used by update/destroy workflows, causing concurrent mutation of the same ECS service/ALB resources". Concurrency bugs are confusing so I thought I'd share, but I'm not sure about how concerning this really is.

millerti: Actually, does this mean that we could trigger two runs for the same PR and they'd run at once?

millerti (commented):
A couple of things I noticed. Not entirely sure how to test manually.

.github/workflows/scan-orphaned-cms-pr-environments.yml
tdooner: The "update" ones have a `concurrency: cms-pr-environment-${{ needs.resolve.outputs.pr_number }}` block so that only one instance runs at a time for the same PR number. I guess Claude is pointing to the fact that _this script_ could run simultaneously with an update for a PR. I think that's pretty unlikely. This scanning/cleanup script also has checks to only delete review apps that are orphaned (i.e. their PR is no longer open)... so I think the chance of a race condition is basically none.

bin/cms-pr-environment
tdooner: Yeah I see what you mean... I suppose I don't love the fact that we would have to add new agencies to this list for them to work. But, I'm also trying to imagine this already-kinda-crazy jq function being dynamically generated. At least this duplicated list is somewhat simple to add new entries to. I'm inclined to keep it for that reason (minimizing complexity).

bin/cms-pr-environment
tdooner: Good point, but we can add one later if this script hits arbitrary timeouts that prevent the deletion of a single Review App or something.

millerti (approved):

PR #1995, opened 2026-08-20
Instrument Mixpanel events for the Education Self-Attested flow in Emmy CE

millerti:
## Changes
- Added `Education*` events for the self-attested and partially-self-attested flows.

## Testing instructions
1. Use the test link below
2. Start an Education self-attested activity ("I don't see my school" / manual entry path).
3. Open Mixpanel's Events view for the Dev Env project and confirm the events fire as you interact with the flow
4. Repeat for the partially self-attested flow
5. Pay attention to the properties on each even (open an event in Mixpanel -- they'll drop on the Dev Env page -- you may want to click the Your Properties filter). Does each event have the info you want/need?

## Acceptance testing

## AI Usage

♻️ Environment destroyed ♻️


millerti:
# Code Review — FFS-4636: Instrument Mixpanel events for the Education Self-Attested flow

Scope: single commit `8825f60c`, ~326 line additions across 16 files. The change mirrors the pattern already established for the Employment Self-Attested flow (PR #1980), adding controller `after_action` hooks + inline `track_event` calls for the Education flow, plus the corresponding `TrackEvent` constants, JS-side `track_event_prefix` wiring on the view components, and controller specs using the existing `"tracks an event"` shared example.

Overall the change is small, follows the established Employment pattern closely, and is well-tested. A few concerns below.

---

## Correctness

### 1. Nested ternary in `document_uploads/new.html.erb:16` is hard to read and easy to break
```erb
track_event_prefix: (params[:employment_id].present? ? "Employment" : params[:education_id].present? ? "Education" : nil)
```
Ruby's right-associative `?:` means this parses correctly today, but as more activity types adopt the header component (community_service, job_training) the nested form will get worse. `DocumentUploadsController` already has a clean `if/elsif` chain on the same param set (`set_activity`, `after_activity_path`, `upload_path`, etc.) — this view is now the only place duplicating that dispatch logic. Consider either (a) exposing a helper/attr on `@activity` (`@activity.track_event_prefix`) or (b) computing it once in the controller and assigning to `@track_event_prefix`. As-is, community_service and job_training silently get `nil` and lose exit-modal / back-click tracking, which is easy to overlook when someone adds Volunteering tracking next.

### 2. `EducationReviewViewed` / `EducationReviewSubmitted` fire for *all* data sources, not only self-attested
`education_controller.rb:18` and `education_controller.rb:115` unconditionally track for every review action, but the ticket scope is the "Self-Attested flow." The review page renders for `validated`, `partially_self_attested`, and `fully_self_attested` (`review.html.erb`, and the "when validated" spec at `education_controller_spec.rb:615-654` confirms this path).

Compare with the intentional guard added at `education_controller.rb:228` for `track_info_viewed_event`, which specifically skips the non-fully-self-attested `edit`. The review tracker has no similar guard. Either:
- **This is intentional** and the review page counts as part of the "self-attested review" for all data sources — worth a comment explaining why.
- **This is a bug** — the validated-flow user hitting `#review` will also fire an event named `EducationReview*`, which is misleading for analytics.

Please confirm with the analytics stakeholder which is desired.

### 3. `track_info_viewed_event` uses `action_name == "edit"` gate but not symmetric for `new`
`education_controller.rb:226-231`:
```ruby
def track_info_viewed_event
  return unless response.successful?
  return if action_name == "edit" && !@education_activity&.fully_self_attested?
  track_event(TrackEvent::EducationInfoViewed, education_activity_id: @education_activity&.id)
end
```
Because `new` always renders the self-attest form, this is functionally correct today. But the condition reads awkwardly — the negation-inside-conjunction is easy to get wrong on a future edit. A slightly clearer form:
```ruby
return if action_name == "edit" && !@education_activity&.fully_self_attested?
```
already exists; consider also renaming to `track_self_attested_info_viewed` so intent is clear at the callback declaration on line 17.

### 4. `education_controller.rb:197` — `@education_activity&.id` after `@education_activity = ... .new`
```ruby
@education_activity = @flow.education_activities.new(...)
...
if @education_activity.save
  ...
else
  track_event(TrackEvent::EducationInfoValidationFailed,
    education_activity_id: @education_activity&.id,   # <— &. is unreachable
    ...)
```
`@education_activity` cannot be nil on line 197 — it was just built on line 189. The safe-nav is harmless but misleading; drop it (or, if kept for parity with `employment_controller.rb:29`, add a short comment there). The spec at `education_controller_spec.rb:54` asserts `education_activity_id: nil` for this branch, which is correct because a freshly built (unsaved) record has `id == nil`.

### 5. `education_controller.rb:112-121` — `save_review` fires the event unconditionally
```ruby
def save_review
  @education_activity.update(review_params)
  @education_activity.publish!
  track_event(TrackEvent::EducationReviewSubmitted, education_activity_id: @education_activity.id)
  ...
```
Nothing checks the return of `update`. Since `review_params` only permits `:additional_comments` (which has no validations here), this is likely fine in practice — but the pattern differs from `update` where a failed save skips the "Submitted" event. If `additional_comments` ever gets validation added, the event will still fire on failure. Low priority, but worth aligning with the `if update(...)` pattern used elsewhere.

### 6. `months_controller.rb:24-35` and `term_credit_hours_controller.rb:55-65` guard on `@activity_month.present?` / `@current_term.present?` for good reason — verify it covers all early-return paths
Both controllers can bail out of their before-action (`set_hours_input_vars` returning early on out-of-range `id`, or `set_term_credit_hours_vars` on empty terms). The guards correctly skip tracking in those cases and the specs at `months_controller_spec.rb:23,63` verify it for `id: 99`. Good coverage. However, there is no similar test for the `@terms.empty?` early return in `term_credit_hours_controller.rb:98-101` — worth adding to lock in the behavior.

### 7. `EducationBackClicked` / `EducationExitClicked` / `EducationExitModalCancelled` / `EducationExitModalConfirmed` constants exist but have no server-side callers
`track_event.rb:82-88`. Those four are fired from the Stimulus controller (`activity_flow_header_controller.js`) via `trackUserAction`, which posts to `api/user_events`. That controller uses the string name from the browser directly; the Ruby constants are not referenced in Ruby source (I grepped). The constants are still useful as a canonical registry / for grep discoverability, so this isn't a bug — just noting that the Ruby-side reader doesn't get compile-time coupling. Consider a comment in `track_event.rb` clarifying which constants are JS-fired vs Ruby-fired, since the file is now non-trivial in length.

---

## Test Coverage

### Strong points
- Uses the shared example `"tracks an event"` consistently, so `cbv_flow_id`, `cbv_applicant_id`, `invitation_id`, and `client_agency_id` are asserted for every event without duplication.
- Both success and validation-failure paths are covered for `EducationInfo*`, `EducationMonth*`, and `EducationTermCreditHours*`.
- Negative cases are covered: `education_controller_spec.rb:278-291` verifies `EducationInfoViewed` is not fired for non-self-attested edit, `document_uploads_controller_spec.rb:266-284` verifies no `EducationDocumentUploadSubmitted` on update failure, and out-of-range month/term index don't fire events (`months_controller_spec.rb:23-27, 63-67`).

### Gaps
- **`save_review` failure path**: no test verifies the behavior if the activity fails to publish or update. Given the current code always fires the event, this may not be a real gap, but see comment #5.
- **`review` for non-self-attested data sources**: `education_controller_spec.rb:615-654` renders review for a validated activity but does not assert whether `EducationReviewViewed` is fired. If the intent is to fire only for self-attested, this test would catch a regression. If the intent is to fire for all, an explicit `it_behaves_like "tracks an event"` in the validated context would document that.
- **Term-credit-hours early-return** (`@terms.empty?`, see comment #6) — no assertion that tracking is skipped.
- **Duplicate params** — `document_uploads_controller.rb` will fire *both* `Employment*` and `Education*` events if a request somehow carries both `employment_id` and `education_id` (they are independent `params.present?` checks, not `elsif`). This is not reachable via the current routes but is a latent footgun; not a blocker but a spec asserting mutual exclusivity would be nice, or reshape the tracking dispatch to use the same `if/elsif` chain as `set_activity`.
- **`education_controller_spec.rb:47-55`** stubs `save` to return `false` rather than passing invalid params that would naturally fail validation. Works, but it hides whichever real validation would produce the `error_message`. Prefer passing genuinely invalid params (`{ school_name: "" }`) *and* asserting the model does have validation errors — otherwise the test could pass if all validations were removed.
- **`education_controller_spec.rb:962-969`** stubs both `fully_self_attested?` *and* `update` on `EducationActivity` to force the validation-failed branch. `allow_any_instance_of` on both a query and mutation method is brittle. Preferable: set up a real fully-self-attested activity (already done in the `context`) and pass empty required fields (`school_name: ""`) so the real validations fail. That also exercises the real `errors.full_messages` value being tracked.

---

## Edge Cases

### 8. `education_controller.rb:194-199` — `error_message` on validation failure trusts model errors
```ruby
track_event(
  TrackEvent::EducationInfoValidationFailed,
  education_activity_id: @education_activity&.id,
  error_message: @education_activity.errors.full_messages.join(", ")
)
```
For freshly built self-attested activities, permitted params (`fully_self_attested_education_params`) include `school_name`, `street_address`, `city`, `state`, `zip_code`, etc. If a model validation ever changes to interpolate submitted user input into the error message (uncommon but possible with custom validators), that input would flow into the analytics pipeline. Currently EducationActivity's validations produce standard "can't be blank" style messages so this is fine, but see security note below.

### 9. `months_controller.rb:31` and `term_credit_hours_controller.rb:62` — event property naming
`month: I18n.l(@current_month, format: :month_year)` is a locale-formatted string ("March 2026" vs "marzo 2026"). If the user has switched to Spanish, this analytics property differs from English users. For funnel analytics this can be painful. Consider emitting an ISO date string (`@current_month.strftime("%Y-%m")`) as a canonical machine-parseable value alongside the display label. Not a blocker.

### 10. `document_uploads_controller.rb:12-27` — the "no files added" branch
When a user submits with no `activity` param (empty upload), both `track_employment_upload_submitted_event` and `track_education_upload_submitted_event` are called on line 15-16. Only one will fire (mutually exclusive route params). This behavior matches Employment's existing pattern, so it's consistent — just worth noting the growing "call all, guard inside" idiom is a smell that will worsen when volunteering/job_training tracking lands. A single `track_upload_submitted_event` that dispatches on activity type would scale better.

---

## Security / Privacy

### 11. Error messages in event payloads
Comment #8 above applies: `error_message: @activity_month.errors.full_messages.join(", ")` (`months_controller.rb:33`), same for term credit hours (`term_credit_hours_controller.rb:63`) and education info (`education_controller.rb:61`, `:198`). ActiveRecord's default validators do not leak user input into error strings, so today's messages are safe. However, if custom validators are added later (e.g. "'{value}' is not a valid ZIP"), user-entered PII (address, contact info from `fully_self_attested_education_params`) would flow to the analytics pipeline. Consider:
- Emitting `error_fields: @activity.errors.attribute_names` (or similar) as a safer alternative — fields tell you *what* failed without exposing *what the user typed*.
- Or documenting in `Trackable` that `error_message` values are considered safe and must remain PII-free.

### 12. `education_activity_id` in events
IDs are internal integers with no external significance — safe.

### 13. No new authorization surface introduced
All new callbacks are `after_action`, called only when the underlying action succeeds authentication/authorization. No policy changes needed.

---

## Style / Consistency

- **`education_controller.rb:15-18`** puts two `after_action` hooks after `before_action` hooks. Consistent with `employment_controller.rb:11-15`. Good.
- **Callback naming**: `track_month_submission_event` (`months_controller.rb:24`) vs `track_month_submitted_event` would read more naturally alongside `EducationMonthSubmitted`. Minor.
- **The event constant block** at `track_event.rb:82-98` is alphabetized correctly. Good.

---

## Summary

The change is small, mirrors an established pattern, and is well-covered by tests. Nothing here is blocking; the highest-value follow-ups are:

1. **Confirm the intended tracking scope for `EducationReviewViewed`/`EducationReviewSubmitted`** on validated / partially-self-attested paths (#2).
2. **De-duplicate the activity-type dispatch** in `document_uploads/new.html.erb:16` (nested ternary) and `DocumentUploadsController` tracking helpers (#1, #10) before more activity types adopt the pattern.
3. **Consider `error_fields` in place of `error_message`** to remove a latent PII surface (#11).
4. Fill in the small spec gaps (#6, plus verifying whether-or-not review tracking fires for validated).

ghost:
All events fire as intended! approved! ✅

app/app/controllers/activities/document_uploads_controller.rb
```diff
@@ -12,11 +13,13 @@ def create
     if params.exclude?(:activity)
       # User clicked the submit button without adding any files
       track_employment_upload_submitted_event
+      track_education_upload_submitted_event
```
tdooner: It's weird that we're going to have to put all these `track_*_event` methods in here and then all of them but one will `return` immediately. We should probably just make a `track_submitted_event` method which has a `case` statement similar to the `set_back_url` method.

app/app/controllers/activities/document_uploads_controller.rb
```diff
@@ -12,11 +13,13 @@ def create
...
       return redirect_to after_activity_path
     end
 
     if @activity.update(document_upload_params)
       track_employment_upload_submitted_event
+      track_education_upload_submitted_event
```
tdooner: Same comment, about the `track_*_event` methods.

Also, can we DRY these up? If we put these as an `after_action` it should run regardless of whether the method returns a redirect, I think?

tdooner (approved):

PR #1996, opened 2026-08-20
FFS-4643: CE timeout screen header and copy incorrectly reference "Report My Income" instead of Emmy

daphnegold:
## Changes

- Updated Emmy CE timeout page to match designs
- Existing Emmy Income timeout page experience is untouched

## Acceptance testing

## AI Usage

Screenshot 2026-08-20 at 9 08 44 AM

♻️ Environment destroyed ♻️


app/app/controllers/pages_controller.rb
```diff
@@ -25,6 +30,14 @@ def error_500
 
   private
 
+  def activity_flow?
+    activity_flow_timeout? || super
+  end
+
+  def activity_flow_timeout?
+    params[:activity_flow_timeout].present?
```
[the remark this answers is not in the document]
daphnegold: This pattern is already established for public branding via URL params. It changes public copy and doesn't expose anything sensitive. Leaving as-is.

app/config/i18n-tasks.yml
```diff
@@ -141,6 +141,7 @@ ignore_missing:
     # Activity Hub flow pages/models
     - "shared.credit_hours*"
     - "activities.*"
+    - "pages.home.activity_flow_timeout.*"
     - "cbv.*.activity_flow.*" # ActivityFlow copy overrides in controllers/views from CBV flow
```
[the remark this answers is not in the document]
daphnegold: We explicitly do not want to add Spanish now

app/app/controllers/cbv/sessions_controller.rb
```diff
@@ -24,12 +24,19 @@ def current_agency
   end
 
   def timeout_path
+    if activity_flow? && session[:flow_id].present?
+      client_agency_id = ActivityFlow.includes(:cbv_applicant).find(session[:flow_id]).cbv_applicant.client_agency_id
```
tdooner: nit: no need for `.includes` here since we're just fetching a single one. But it also doesn't hurt.

app/app/controllers/cbv/sessions_controller.rb
```diff
@@ -24,12 +24,19 @@ def current_agency
...
+      return root_url(cbv_flow_timeout: true, activity_flow_timeout: true, client_agency_id: client_agency_id)
```
tdooner: So like do we need `cbv_flow_timeout` AND `activity_flow_timeout`? Can we update the logic so whatever is happening on the homepage just uses `activity_flow_timeout` instead of checking both? I find it a bit confusing/smelly to have both these params in the URL simultaneously.

tdooner (approved):

app/app/controllers/cbv/sessions_controller.rb
daphnegold: You're absolutely right, should have refactored that

ghost:
approved by design! 🚢 itttttt

PR #1997, opened 2026-08-20
FFS-4705: Add "Choose how you want to add your work" page to Emmy CE Employment flow

bencalegari:
## [FFS-4705](https://jiraent.cms.gov/browse/FFS-4705)
- Clicking the option to add employment in the hub now brings you to a page asking which type of employment it is. If its the first option it takes you to the employer search, if its the other two it sends you to the self attested employment page. If no option is selected and you hit continue, it sends you back prompting you to select one.
- Track `ApplicantAccessedAddYourWorkPage` on page view and `ApplicantContinuedFromAddYourWorkPage` with one of the following `add_work_method` options based on what was selected: `connect_automatically`, `enter_paid_manually`, or `enter_unpaid_manually`
- Fix radio button creating unintentional `hint` attribute in the `input` which does nothing.

## Context for reviewers
- The new page becomes the entry point for adding employment, so employer search is no longer the first step and no longer renders a back link.
- Both manual options currently route to the same self-attested employment form. The paid vs. unpaid distinction is only captured in the tracking event for now.
- `add_work_method` is validated against an allowlist in the controller rather than a model, since nothing is persisted at this step.
- The `UswdsFormBuilder#radio_button` change is a general fix: `:hint` was being passed through to the `input` tag along with the label options. It only affects radios that pass a hint.

## Acceptance testing

## AI Usage

♻️ Environment destroyed ♻️

### DIFF
--- a/app/app/views/activities/employment/add_your_work/show.html.erb
+++ b/app/app/views/activities/employment/add_your_work/show.html.erb
@@ -1,0 +1,19 @@
+<% content_for :title, t(".header") %>
+
+<% content_for :activity_flow_header do %>
+  <%= render ActivityFlowHeaderComponent.new(
+    title: t("activities.employment.title_singular"),
+    exit_url: activities_flow_root_path
+  ) %>
+<% end %>
+
+<h1><%= t(".header") %></h1>
+<p><%= t(".description") %></p>
+
+<%= form_with(url: activities_flow_income_add_your_work_path, class: "maxw-none", builder: UswdsFormBuilder, id: "add_your_work_form") do |f| %>
+  <%= f.radio_button(:add_work_method, "connect_automatically", { label: tag.span(t(".options.automatic.label"), class: "text-bold"), hint: t(".options.automatic.hint") }) %>
+  <%= f.radio_button(:add_work_method, "enter_paid_manually", { label: tag.span(t(".options.paid.label"), class: "text-bold"), hint: t(".options.paid.hint") }) %>
+  <%= f.radio_button(:add_work_method, "enter_unpaid_manually", { label: tag.span(t(".options.unpaid.label"), class: "text-bold"), hint: t(".options.unpaid.hint") }) %>
+
+  <%= f.submit t("continue") %>
+<% end %>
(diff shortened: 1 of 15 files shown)


app/app/views/activities/employment/add_your_work/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<h1><%= t(".header") %></h1>
+<p><%= t(".description") %></p>
+
+<%= form_with(url: activities_flow_income_add_your_work_path, class: "maxw-none", builder: UswdsFormBuilder, id: "add_your_work_form") do |f| %>
+  <%= f.radio_button(:add_work_method, "connect_automatically", { label: tag.span(t(".options.automatic.label"), class: "text-bold"), hint: t(".options.automatic.hint") }) %>
+  <%= f.radio_button(:add_work_method, "enter_paid_manually", { label: tag.span(t(".options.paid.label"), class: "text-bold"), hint: t(".options.paid.hint") }) %>
+  <%= f.radio_button(:add_work_method, "enter_unpaid_manually", { label: tag.span(t(".options.unpaid.label"), class: "text-bold"), hint: t(".options.unpaid.hint") }) %>
+
+  <%= f.submit t("continue") %>
```
tdooner, line 18: Huh, we have a top-level "continue" key? TIL

app/spec/controllers/activities/income/employer_searches_controller_spec.rb
```diff
@@ -26,6 +26,11 @@
       expect(response.body).to include("exit-confirmation-modal")
     end
 
+    it "does not render a back link" do
+      get :show
+      expect(Capybara.string(response.body)).not_to have_link("Back")
+    end
```
tdooner, line 32: Interesting. We have a ticket to change this, right?

tdooner (approved):

app/spec/controllers/activities/income/employer_searches_controller_spec.rb
millerti: It's AC #5 on this ticket, so let's just do it here

bencalegari: Exit and back would do the same do the same thing on this page, and the intro pages for the other activities don't have a back button either so I'll leave it out.

tdooner: I don't think they would do the same thing.... "Exit" would go back to Activity Hub, but "Back" would go back to your new page, right?

mayene-nava:
Looks good - ✅ approved from design.

PR #2009, opened 2026-08-26
FFS-3788: Investigate and fix undefined method 'identity' for an instance of CbvFlow

krista-skylight:
## Changes
When restoring the saved session we used a generic lookup that could return the wrong model. So `@flow` could be a `CbvFlow `when the Activities controller expected an ActivityFlow. Calling `@flow.identity` on the wrong object raised the error. When loading from session, we now ask for the specific kind of flow the controller expects instead of using the generic finder. Also added a regression spec.

## Context for reviewers
Sorry I don't have EUA access yet and I'm with skylight so I can't get the ticket link or tag acceptance testing in slack!

## Acceptance testing

## AI Usage

## Infrastructure Changes

  - [ ] Plan reviewed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,6 +53,7 @@ repos:
     name: prettier
     entry: bin/run-linter --cd app -- npm run format:precommit
     language: node
+    additional_dependencies: ["prettier"]
     files: '^app/.*\.(js|ts|json)$'
 
   - id: terraform_fmt
--- a/app/app/controllers/flow_controller.rb
+++ b/app/app/controllers/flow_controller.rb
@@ -74,10 +74,19 @@ class FlowController < ApplicationController
       track_invitation_clicked_event(invitation, @flow)
     elsif session[:flow_id]
       begin
-        # Use the controller's `flow_param` so controllers scoped to a
-        # particular flow (e.g. activities) don't accidentally load a
-        # different flow type from the session (e.g. CbvFlow).
-        @flow = flow_class(flow_param).find(session[:flow_id])
+        # For activity-scoped controllers, require the ActivityFlow class be
+        # used so an activities controller cannot accidentally load a
+        # CbvFlow from a session id. For other controllers preserve the
+        # previous behavior of resolving the flow class from the
+        # `session[:flow_type]` when present to avoid changing unrelated
+        # behavior (eg. tests that expect a CBV controller to honor the
+        # session's flow type).
+        if flow_param == :activity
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -53,6 +53,7 @@ repos:
     name: prettier
     entry: bin/run-linter --cd app -- npm run format:precommit
     language: node
+    additional_dependencies: ["prettier"]
```
bencalegari, line 56: Not opposed to the dependency but surprised to see this in the commit since we aren't changing any js here. Any idea why this change came in?

app/app/controllers/flow_controller.rb
```diff
@@ -74,7 +74,19 @@ def set_flow
...
-        @flow = flow_class.find(session[:flow_id])
+        # For activity-scoped controllers, require the ActivityFlow class be
+        # used so an activities controller cannot accidentally load a
+        # CbvFlow from a session id. For other controllers preserve the
+        # previous behavior of resolving the flow class from the
+        # `session[:flow_type]` when present to avoid changing unrelated
+        # behavior (eg. tests that expect a CBV controller to honor the
+        # session's flow type).
+        if flow_param == :activity
+          @flow = flow_class(flow_param).find(session[:flow_id])
```
bencalegari, line 85: Could we get rid of the conditional and just always use `flow_class(flow_param)` here? Or is the issue that `flow_param` isn't set correctly?

bencalegari (approved):

.pre-commit-config.yaml
krista-skylight: Yep, didn't change any js. Pre-commit failed trying to install the Node hook because there was no `package.json` in the repo root. I added this to make the pre-commit node env installable and get around that. how have folks dealt with this in the past?

app/app/controllers/flow_controller.rb
krista-skylight: I added a conditional statement due to this test that failed in CI. 
`Cbv::BaseController#append_log_tags for an activity flow adds log tags
Failure/Error: expect(response).to have_http_status(:ok)
expected the response to have status code :ok (200) but it was :found (302)`
For CBV controllers we keep the old behavior so existing CBS tests that rely on that continue to get 200 OK instead of a redirect.

.pre-commit-config.yaml
bencalegari: Hmm I haven't noticed this, I wonder if its based on where you ran the commit from? I work out of the `/app` directory if that makes any difference. All good we can keep it in rather than debug it endlessly

PR #2010, opened 2026-08-27
FFS-4709 Manual employment entry incorrectly requires both income AND hours

krista-skylight:
## Changes
Changed logic of `valid_hours_submission` to OR instead of AND for income and hours. Modified the language accordingly for the relevant test.

## Acceptance testing

## AI Usage

## Infrastructure Changes

  - [ ] Plan reviewed

♻️ Environment destroyed ♻️


krista-skylight:
Acceptance testing screenshot
image

app/app/controllers/flow_controller.rb
daphnegold, line 83: Nit: I'm not a fan of huge comments like as they are so verbose and hard to parse. Or comments in general unless really necessary, they get out of date. But, again, I think this is accidentally showing up in the diff of this PR.

app/spec/controllers/activities/employment/months_controller_spec.rb
```diff
@@ -170,7 +170,7 @@
         )
       end
 
-      it "requires at least one month with both income and hours on the last page" do
+      it "requires at least one month with either income or hours on the last page" do
```
daphnegold, line 193: Can we add tests for income only and hours only? This test still uses 0 for both, so it doesn't cover the new behavior at all.

app/app/controllers/activities/employment/months_controller.rb
```diff
@@ -116,10 +116,10 @@ def valid_hours_submission?
     hours = @activity_month.hours || 0
 
     if @months.length == 1
-      income > 0 && hours > 0
+      income > 0 || hours > 0
```
daphnegold, line 119: I think we should update the copy to reflect the "or" condition, as it still says "and"

.gitignore
```diff
@@ -35,6 +35,7 @@ Brewfile.lock.json
 # Ignore local dotenv overrides
 .env*.local
 .env.test
+**/.env
```
daphnegold:
```suggestion
```

This file is already tracked, let's drop this from the PR. Doing this will not hide local changes.

.pre-commit-config.yaml
daphnegold, line 56: Why do we need this dependency? I think this might be from an unrelated commit, anyway.

daphnegold (changes requested):
There are changes from another PR in your code diff. Could you update this branch from main, removing the unrelated commits? 

Good progress, some adjustments requested!

daphnegold:
It looks like you reverted PR #2009 after merging main, so this PR now removes that fix and its regression test. Could you restore those changes and update the branch? I think the easiest thing to do is revert the revert and push.

app/app/controllers/activities/employment/months_controller.rb
krista-skylight: Are you referring to the error message? If so, just made a change to fix that!

app/spec/controllers/activities/employment/months_controller_spec.rb
```diff
@@ -128,6 +128,26 @@
...
+
+    it "passes validation when only hours are provided" do
+      patch :update, params: {
+        employment_id: employment_activity.id,
+        id: 0,
+        employment_activity_month: { gross_income: 0, hours: 10 }
+      }
+
+      expect(response).to redirect_to(new_activities_flow_income_employment_document_upload_path(employment_id: employment_activity))
+    end
```
daphnegold, line 149: These cover the singlular month case, but could we add a success for multi-month to be thorough?

daphnegold (approved):
Looking great! Suggested some enhanced testing before merge.

mayene-nava:
Design AT: looks good, can get to employment review screen entering only hours OR income.

PR #2012, opened 2026-08-28
FFS-4592: Fix NSC test scenarios in launcher

bencalegari:
## [FFS-4592](https://jiraent.cms.gov/browse/FFS-4592)
Rick and other NSC results were failing because `nameOnSchoolRecord` had gone missing from the api response. We were getting `NoMethodError`s, but the retry logic was so broad we didn't see the failure show up locally for many many minutes later. I updated the fixtures and guarded against it, defaulting to the identity's first and last name. Personas also changed status in the sandbox: Dominique now returns `CC` (currently enrolled) instead of `CN`, so she took over the full-time scenario and Scott Tobin was added to cover the not-currently-enrolled case.

## Changes
- Re-recorded all NSC fixtures against the live NSC sandbox; replaced the ad-hoc `nsc_save_fixture_for_user` helper with a `FIXTURE_PERSONAS` table plus `nsc_save_all_fixtures` so fixtures can be regenerated in one command.
- Added the Scott Tobin launcher scenario (not currently enrolled / enrollment older than 18 months) and moved Dominique Ricardo to the "enrolled full time" scenario, matching the re-recorded fixture data.
- Changed the simple launcher's default student status from Lynette to Dominique and added a "Enrolled, intensity not reported" option backed by Lynette.
- `Launcher::NscForwardDatingService` anchors forward-dating on the latest term ending on or before the persona's launcher as-of date, rather than the latest term overall, and adds a stable as-of date for Dominique.
- `ApplicationJob` no longer retries programming errors (`NameError`, `TypeError`, `ArgumentError`) or `ActiveRecord::RecordInvalid`, they fail on the first attempt.
- Deleted the unused `lakhshmi_saxena` fixture.
- Snuck in a change Paul requested to show credit hours/hours if present on education card.

## Context for reviewers
- The NSC sandbox stopped returning `nameOnSchoolRecord` and `studentAddress` and now returns `schoolType`, so the re-recorded fixtures dropped those fields. `schoolType` is a string of a number referring to whether its a 4 year, 2 year, or 1 (less than 2 year) institution which isn't relevant for our purposes. 
- Lynette's fixture now has two terms, and the newer one (2026-05-01 – 2027-05-01) begins after her as-of date. Anchoring on the max term end shifted dates *backwards*; anchoring on the latest term at or before the as-of date is what keeps forward-dating correct.
- There was a bug where self attested credit hours weren't being shown on the doc upload page for Lynette and I snuck in the fix: the issue was `enrollment_less_than_half_time?` only returns true when the status is literally `less_than_half_time` where we actually want it to be true when its `less_than_half_time`, `enrolled`, or `unknown`, which is the logic we used everywhere except there. 

## Acceptance testing

## AI Usage

- Link to launcher: https://pr-2012.dev.emmy.cms.gov/launcher/advanced
- Deployed commit: 962cd6a07e0d4662ce4efa11dfce6a34b976f49d


app/app/jobs/application_job.rb
```diff
@@ -1,9 +1,17 @@
...
+    NameError,
+    TypeError,
+    ArgumentError,
+    ActiveRecord::RecordInvalid
+  ].freeze
+
   around_perform :with_error_reporting
 
   retry_on Exception, wait: :polynomially_longer, attempts: 5
+  retry_on(*NON_RETRYABLE_ERRORS, attempts: 1)
```
tdooner, line 14: Perhaps we should just `retry_on StandardError` instead of `Exception`. All of the `NON_RETRYABLE_ERRORS` you list are not StandardErrors, and there are other ones that are not StandardErrors too that we probably also don't want to retry on, like SyntaxError, NotImplementedError, and LoadError

app/app/services/nsc_data_fetcher_service.rb
```diff
@@ -66,19 +66,23 @@ def update_education_activity(education_activity, response_data)
...
         next unless activity_flow.within_reporting_window?(term_begin, term_end)
 
         @education_activity.nsc_enrollment_terms.create!(
           school_name: enrollment_detail["officialSchoolName"],
-          first_name: enrollment_detail["nameOnSchoolRecord"]["firstName"],
-          middle_name: enrollment_detail["nameOnSchoolRecord"]["middleName"],
-          last_name: enrollment_detail["nameOnSchoolRecord"]["lastName"],
+          first_name: name_on_school_record["firstName"] || identity&.first_name,
+          middle_name: name_on_school_record["middleName"],
+          last_name: name_on_school_record["lastName"] || identity&.last_name,
```
tdooner, line 85: Hmm. I'm not sure we should fall back to saving the identity data again in the `nsc_enrollment_terms` - since `nsc_enrollment_terms` is just capturing the NSC result, supplementing it with other data might confuse us as to whether we got the data from NSC or somewhere else.

Can we just leave these fields blank if NSC returns nothing, and then handle it in whatever is reading the NSC enrollment term?

Also, did you say NSC no longer returns this field, or only sometimes? Or do we think it will be returned in production but not in sandbox?

pauljgehrig:
LGTM! 🚢-it

PR #2013, opened 2026-08-31
FFS-4596 No consent pdf showing

ccvetter:
## Changes

Added a guard to prevent a user from trying to load a pdf without having checked the consent checkbox.

## Context for reviewers

The PDF endpoint could render without recorded consent, causing a 500 error. This happens if a user tries to load the PDF url directly without consent. The user should be redirected back to the agreement step if trying to directly load the PDF without giving consent.

## Acceptance testing

## AI Usage
- [ ] USED_AI: I attest that I have read, understood, and take ownership of all AI-generated code in this PR.

### DIFF
--- a/app/app/controllers/cbv/submits_controller.rb
+++ b/app/app/controllers/cbv/submits_controller.rb
@@ -21,7 +21,7 @@ class Cbv::SubmitsController < Cbv::BaseController
     respond_to do |format|
       format.html
       format.pdf do
-        return if redirect_unless_consent!(t("cbv.submits.update.consent_to_authorize_warning"))
+        next if redirect_unless_consent!(t("cbv.submits.update.consent_to_authorize_warning"))
 
         event_logger.track(TrackEvent::ApplicantDownloadedIncomePDF, request, {
           time: Time.now.to_i,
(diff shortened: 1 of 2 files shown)


app/app/controllers/cbv/submits_controller.rb
```diff
@@ -78,6 +77,14 @@ def check_aggregator_report
     end
   end
 
+  def redirect_unless_consent!(warning_message)
+    return false if has_consent
+
+    @cbv_flow.errors.add(:consent_to_authorized_use, :blank, message: warning_message)
```
tdooner: Interesting. I see you copied this from the way it was before, but I think it's doing nothing since the errors will clear after the redirect anyway. It would do something if we did `render :show` in the `update` action, but since we do `redirect_to` then only the flash message will show, not the `@cbv_flow.errors`.

Up to you if you want to remove or just leave it vestigially remaining.

tdooner (approved):

app/app/controllers/cbv/submits_controller.rb
ccvetter: I am going to remove it.

PR #2047, opened 2026-09-08
FFS-4713 Add an environment variable to disable NSC in a given environment, forcing the Education flow to self-attestation only.

krista-skylight:
## Changes

•Added helper method `nsc_disabled?` in `ApplicationController` and `ApplicationHelper` to safely parse and check the `NSC_DISABLED` environment variable across controllers and views.
•Updated `ActivitiesHelper#education_add_path` and app/views/activities/activities/index.html.erb to route the Activity Hub's education "Add" action directly `to new_activities_flow_education_path` when `NSC_DISABLED` is true.
•Updated `Activities::EducationController` actions (verify, create, show, sync, and error) and `Activities::Education::TermCreditHoursController` to guard against data-backed NSC paths and redirect to the self-attestation flow when `NSC_DISABLED` is enabled.
•Updated app/views/activities/education/error.html.erb to hide the data-backed retry button when `nsc_disabled?` is true.
•Documented NSC_DISABLED in app/.env, app/.env.local.example, and docs/local-replication.md.
•Added unit and controller tests across `ActivitiesHelper`, `Activities::ActivitiesController`, `Activities::EducationController`, and `Activities::Education::TermCreditHoursController`.

## Acceptance testing

## AI Usage

## Infrastructure Changes

  - [ ] Plan reviewed

## Preview environment
- Link to launcher: https://p-2047.navapbc.cloud/launcher/advanced
- Deployed commit: bab10cff40ee70e3a7377950e27010907bdae535


app/app/controllers/activities/education/term_credit_hours_controller.rb
```diff
@@ -139,4 +140,8 @@ def mixed_enrollment_in_window?
     @education_activity.nsc_enrollment_terms
       .any? { |term| term.within_reporting_window?(reporting_range) && term.half_time_or_above? }
   end
+
+  def redirect_if_nsc_disabled
+    redirect_to new_activities_flow_education_path if nsc_disabled?
+  end
```
millerti: This style feels much nicer to me than scattering `if nsc_disabled?` checks all over the place. What if we moved this definition up into /activities/base_controller.rb and shared it? That way, you could replace all those checks with `before_action :redirect_if_nsc_disabled only: %i[verify show sync error]`

app/app/controllers/application_controller.rb
```diff
@@ -131,6 +131,10 @@ def activity_hub_enabled?
     Rails.env.development? || ENV["ACTIVITY_HUB_ENABLED"] == "true"
   end
 
+  def nsc_disabled?
+    ActiveModel::Type::Boolean.new.cast(ENV["NSC_DISABLED"])
+  end
```
millerti, line 136: I notice you have two of these, and they're **not** identical. To me that's a strong signal that we should try to unify this code somewhere. Maybe in a new shared /app/lib/nsc.rb?

app/.env
```diff
@@ -48,3 +48,4 @@ MISSION_CONTROL_USER=user
 MISSION_CONTROL_PASSWORD=password
 
 STRUCTURED_LOGGING_ENABLED=false
+NSC_DISABLED=false
```
millerti, line 51: This is good for a local default and testing, but we're going to need to set this in AWS Parameter Store and load the setting in each environment that way. Take a look at infra/app/app-config/env-config/environment-variables.tf to see how we do that with other variables like SITE_ALERT_ENABLED. Take a gander at the PR that added that for a guide on what's needed.

(And, in addition, we will need to go through and create this variable in Parameter Store in each environment, and set it to false.)

millerti (changes requested):

emmybot-cms:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/emmy/pr-checks/99636452-52c7-470c-9b37-a5c413846514?refs=aa9e3825794884c0e7e15fc2909e45f943a3fd9c&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/emmy/pr-checks/99636452-52c7-470c-9b37-a5c413846514/license?refs=aa9e3825794884c0e7e15fc2909e45f943a3fd9c&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

app/.env
krista-skylight: I _think_ I got it done. How do I go about getting credentials to create the variable in parameter store?

app/lib/nsc.rb
```diff
@@ -0,0 +1,9 @@
+module Nsc
+  def self.disabled?
+    ActiveModel::Type::Boolean.new.cast(ENV["NSC_DISABLED"]) || false
```
millerti, line 3: I'm wondering if this would just be better as

`ENV["NSC_DISABLED"] == "true"`

krista-skylight: I think the only difference is casting would allow it to accept other truthy values, but for clarity's sake, we can expect that the user always needs to enter 'true'. I'll go with your suggestion!

infra/app/app-config/env-config/environment-variables.tf
```diff
@@ -17,6 +17,7 @@ locals {
   ssm_environment_variables = {
     ACTIVITY_HUB_ENABLED       = "/service/${var.app_name}-${var.environment}/activity-hub-enabled"
     LA_LDH_PILOT_ENABLED       = "/service/${var.app_name}-${var.environment}/la-ldh-pilot-enabled"
+    NSC_DISABLED               = "/service/${var.app_name}-${var.environment}/nsc-disabled"
```
krista-skylight: note to self: can get rid of this

app/lib/nsc.rb
```diff
@@ -0,0 +1,9 @@
+module Nsc
+  def self.disabled?
+    ENV["NSC_DISABLED"] == "true"
```
krista-skylight: note to self:

```suggestion
    ENV.fetch("NSC_DISABLED", true)
```

app/lib/nsc.rb
krista-skylight: I think I have to use `cast` because in Ruby, strings are truthy. So `"false"` would evaluate to true.

app/lib/nsc.rb
```diff
@@ -0,0 +1,9 @@
+module Nsc
+  def self.disabled?
+    ActiveModel::Type::Boolean.new.cast(ENV.fetch("NSC_DISABLED", "true"))
```
krista-skylight: @millerti This is the key logic that changed so please double check that this looks correct to you! 

Also, I know you previously suggested not using the `cast`, but I think Ruby treats the string `"false"` as truthy, so it's necessary.

millerti: I see this is back. Is this from Tom's suggestion? If so I suppose he knows better. 

I am more concerned because I think this means "NSC is disabled unless explicitly enabled", which to my understanding is _not_ what we want. We want NSC to be enabled until someone goes in and explicitly disables it. Did that desire change, or does this line need to be fixed?

millerti: Talked with Tom and understand the situation better now. Can you please update the ticket to describe the new state of the setting (aka it defaults to true meaning NSC will be disabled by default when we deploy this)?

millerti:
These AI comments are mostly for you, but they clutter up the PR. I'd recommend reviewing them, implementing changes if they make sense, but then marking them as resolved or outdated so they don't distract us from the real work.

app/lib/nsc.rb
daphnegold: I think I accidentally got the conditions swapped in my head when talking with Krista yesterday on how to implement this. My bad entirely.

If we deploy this to Nava infrastructure, it will default flag `NSC_DISABLED` to true because the environment variable is not available, and **disable** NSC which is *not* what we want, right? So this should be `ENV.fetch("NSC_DISABLED", "false")`, right?

millerti:
Ok, I'm ready to do some manual testing. I notice that there is no code to load the parameter from the Parameter Store anymore. So is it correct to say that, when we deploy this, it will disable NSC everywhere until we manually add that param in an environment and set it to false?

If that's accurate, I think you just need to redeploy your test env (using the Github Action) because right now it's still enabled there.

millerti (approved):
Looks good!

pauljgehrig:
Approved! ✅

